PR #5, opened 2017-12-19
Implement total cube count functionality

slobodan-ilic:
- Add unit tests for 'count' API method
- Add integration tests for 'count' method (using fixtures)
- Implement count functionality


slobodan-ilic:
This has already been merged to the master branch. I'm merging this PR just for the sake of consistent history (better than a closed PR).

malecki:
Can this be changed to `total_n` or `n_rows`? I apologize for not catching this earlier. `count` is way overloaded and we should definitely try to avoid it here. I still think `margin(weighted=False)` is what one wants most of the time, as I cannot at the moment guarantee that `total_n` will report anything useful when an array or multiple response dimension is involved.

PR #6, opened 2018-01-04
Implement z scores based on Chi-squared

slobodan-ilic:
This is first in a row of the PRs, that strive to provide Z-Scores as an integral part of the actual _cube_, when it comes from ZZ9. Any suggestions welcome.

Associated work:
- [pivotal](https://www.pivotaltracker.com/n/projects/931610/stories/153613592)
- [rcrunch](https://github.com/Crunch-io/rcrunch/pull/175/files)


src/cr/cube/crunch_cube.py
```diff
@@ -837,3 +838,35 @@ def index(self, axis, weighted=True):
             margin = margin[:, np.newaxis]
 
         return np.true_divide(props, margin)
+
+    @property
+    def z_scores(self):
```
malecki: could this be called something like `std_resid` or even `standardized_residuals`?

tests/integration/test_z_scores.py
```diff
@@ -0,0 +1,88 @@
...
+
+    def test_z_scores_admit_x_gender_weighted(self):
+        '''Z-Scores for normal weighted crosstab.'''
+        cube = CrunchCube(FIXT_ADMIT_X_GENDER_WEIGHTED)
+        expected = np.array([
+            [9.42561985, -9.42561985],
+            [-9.42561985, 9.42561985],
+        ])
+        actual = cube.z_scores
+        np.testing.assert_almost_equal(actual, expected)
```
malecki: remind me, is assert_almost_equal sufficiently precise by default? I vaguely recall that it was like 5 decimals and we should be able to count on at least 8 or 12, I would think

tests/integration/test_crunch_cube.py
```diff
@@ -1407,275 +1405,6 @@ def test_subtotals_proportions_two_transforms_missing_included(self):
                                   include_transforms_for_dims=[0])
         np.testing.assert_almost_equal(actual, expected)
 
-    def test_labels_on_2d_cube_with_hs_on_1st_dim(self):
```
malecki: what's up with all this delete? did it move somewhere else? wrong branch or something?

src/cr/cube/crunch_cube.py
```diff
@@ -837,3 +838,35 @@ def index(self, axis, weighted=True):
...
+
+        if self.has_mr:
+            return self._mr_z_scores
+
+        counts = self.as_array()
+        total = self.margin()
+        p_col = self.margin(axis=0) / total
+        p_row = self.margin(axis=1) / total
+        expected_counts = chi2_contingency(counts)[3]
```
malecki: would `scipy.stats.contingency.expected_freq` be clearer than ‘the fourth thing it spits out’?

src/cr/cube/crunch_cube.py
```diff
@@ -837,3 +838,35 @@ def index(self, axis, weighted=True):
...
+        residuals = counts - expected_counts
+        V = expected_counts * ((1 - p_row[:, np.newaxis]) * (1 - p_col))
```
malecki: Is there a good reason not to do this in count space vs probability space? Only to do less floating point division and less possibility for rounding error. 
```
colsum = self.margin(axis=0)
rowsum = self.margin(axis=1)
V = np.outer(colsum, rowsum) * np.outer(total-colsum, total-rowsum) / n**3
```
(as usual, I might be transposed)

src/cr/cube/crunch_cube.py
```diff
@@ -837,3 +838,35 @@ def index(self, axis, weighted=True):
...
+    def _mr_z_scores(self):
+        counts = self.as_array()
+        total = self.margin()
+
+        if self.dimensions[0].type == 'multiple_response':
+            p_col = self.margin(axis=0)
+            p_row = self.margin(axis=1)[:, np.newaxis]
+        elif self.dimensions[1].type == 'multiple_response':
+            p_row = self.margin(axis=0)
+            p_col = np.sum(counts, 0)
```
malecki: forthcoming MR⨉MR?

src/cr/cube/crunch_cube.py
slobodan-ilic: For this, I just followed what's implemented in R. I'll try your way, and see what happens :+1:

tests/integration/test_crunch_cube.py
slobodan-ilic: It moved to a separate file (test headings and subtotals). It was just a preparation part for this PR, you can see it if you go commit by commit.

tests/integration/test_z_scores.py
slobodan-ilic: The default is `decimal=7` in the function definition... I think we can increase the precision if we want to...

src/cr/cube/crunch_cube.py
slobodan-ilic: I'm fine with `standardized_residuals`, as @rdrivers pointed out in the original PDF.

slobodan-ilic:
@malecki I've read all of the comments. I'll address the function name change, as well as the count vs. proportions space calculations. Will let you know once I'm done.

P.S. I also have to fix the coverage...

src/cr/cube/crunch_cube.py
slobodan-ilic: :+1:

slobodan-ilic:
Closed in favor of #19

PR #20, opened 2018-01-28
Pruning for single cat 154618272

slobodan-ilic:
Per [ticket](https://www.pivotaltracker.com/story/show/154618272)

Needs to be rebased #19 (only last commit is relevant)

### DIFF
--- a/src/cr/cube/crunch_cube.py
+++ b/src/cr/cube/crunch_cube.py
@@ -237,7 +237,7 @@ class CrunchCube(object):
 
         if prune and not self.has_mr:
             # Remove columns and rows where marginal value is 0 or np.nan
-            if len(self.dimensions) == 1:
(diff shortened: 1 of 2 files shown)


src/cr/cube/crunch_cube.py
```diff
@@ -237,7 +237,7 @@ def _as_array(self, include_missing=False, get_non_selected=False,
 
         if prune and not self.has_mr:
             # Remove columns and rows where marginal value is 0 or np.nan
-            if len(self.dimensions) == 1:
+            if len(self.dimensions) == 1 or len(res.shape) == 1:
```
malecki: hmm. It’s tricky where to decide to reshape `(1)` to `(1, )`. 
I wonder if there’s a way to do it earlier so as to avoid an extra condition here. This fix is super simple so 👍 but just thinking about factoring the cube in the future.

PR #23, opened 2018-01-31
Support "top" and  "bottom" H&S placement

slobodan-ilic:
Per [ticket](https://www.pivotaltracker.com/story/show/154365667)


src/cr/cube/crunch_cube.py
```diff
@@ -176,92 +198,131 @@ def _get_values(self, weighted, margin=False):
...
+        '''Prune the result based on margins content.
+
+        Pruning is the removal of rows or columns, whose corresponding
+        marginal elements are either 0 or not defined (np.nan).
+        '''
+
+        if len(self.dimensions) == 1 or len(res.shape) == 1:
+            # For 1D, margin is calculated as the row margin.
+            margin = self.margin(include_transforms_for_dims=transforms, axis=1)
+            ind_prune = np.logical_or(margin == 0, np.isnan(margin))
```
percious: why not just do "col_margin == 0 or  np.isnan(col_margin)"  not sure why logical or is needed here

src/cr/cube/crunch_cube.py
```diff
@@ -176,92 +198,131 @@ def _get_values(self, weighted, margin=False):
...
+            return res[~ind_prune]
+
+        # Prune columns by column margin values.
+        col_margin = self.margin(include_transforms_for_dims=transforms, axis=0)
+        ind_prune = np.logical_or(col_margin == 0, np.isnan(col_margin))
```
percious: same above

src/cr/cube/crunch_cube.py
```diff
@@ -176,92 +198,131 @@ def _get_values(self, weighted, margin=False):
...
+        res = res[:, ~ind_prune]
+
+        # Prune rows by row margin values.
+        row_margin = self.margin(include_transforms_for_dims=transforms, axis=1)
+        ind_prune = np.logical_or(row_margin == 0, np.isnan(row_margin))
```
percious: same above

src/cr/cube/crunch_cube.py
slobodan-ilic: Because it doesn't work with `numpy` (or I haven't figured it out). I mean, it works for single value, but not "bitwise". Will try to find an option to make it work...

PR #25, opened 2018-02-05
Fix bad anchors in H&S

slobodan-ilic:
Per [ticket](https://www.pivotaltracker.com/story/show/154871576)


src/cr/cube/dimension.py
```diff
@@ -111,19 +111,35 @@ def inserted_hs_indices(self):
...
+        ]
+        # In case of more matches, return the first one (although there
+        # shouldn't be any)
+        return contiguous_anchors[0]
+
     @property
     def hs_indices(self):
         '''Headers and Subtotals indices.'''
-        elms = self._elements
+        elements = self._elements
```
percious: 👍

PR #27, opened 2018-02-15
Add population counts support

kwypchlo:


percious:
Can you please add a test for this?

src/cr/cube/crunch_cube.py
```diff
@@ -1059,6 +1059,33 @@ def percentages(self, axis=None):
...
+               [5, 3],
+            ])
+
+            >>> cube.population_counts(9000)
+            np.array([
+                [3000, 1200],
+                [3000, 1800],
+            ])
+        '''
+        return self.proportions() * population_size
```
slobodan-ilic: Forgive my ignorance (about population counts), but is it _certain_ that we don't need to be dealing with axes here (like direction of the proportions)?
@kwypchlo :point_up:

kwypchlo: yup, that's my understanding and this is how it works in whaam

kwypchlo:
of course, I just had to set up my environment to run them first :)

src/cr/cube/crunch_cube.py
```diff
@@ -1059,6 +1059,33 @@ def percentages(self, axis=None):
...
+
+        Args
+            population_size (int): Estimated total population size
+
+        Returns
+            (nparray): Calculated array of crunch cube population counts.
+
+        Example:
+            >>> cube = CrunchCube(fixt_cat_x_cat)
+            np.array([
```
slobodan-ilic: @kwypchlo It's not clear what this output is the result of? I'm guessing it's the `cube.as_array()`. We can either add that line, or make a `__repr__` method work properly :) You decide :)

src/cr/cube/crunch_cube.py
malecki: @slobodan-ilic is right, these should be proportions *in some direction* multiplied by [filtered] population number

src/cr/cube/crunch_cube.py
malecki: I would test `cube.as_array()` explicitly

src/cr/cube/crunch_cube.py
malecki: but it's actually an estimate of the count in the cell, which doesn't change, it would be with respect to the whole table

PR #29, opened 2018-03-06
Means 155735693

slobodan-ilic:
Per [ticket](https://www.pivotaltracker.com/story/show/155735693)


src/cr/cube/crunch_cube.py
```diff
@@ -1189,3 +1107,31 @@ def index(self, axis, weighted=True, prune=False):
...
+            axis = 1 - self.mr_dim_ind
+            return np.sum(contents, axis) / np.sum(self.as_array(), axis)
+
+        axis = 0
+        if self.ca_dim_ind == 0 or self.ca_dim_ind == 2:
+            axis = 1
+        elif len(self.dimensions) > 2 and self.ca_dim_ind == 1:
+            axis = 2
+
+        return np.sum(contents, axis) / self.margin(axis)
```
malecki: I did this too the first time but it’s not quite right: we need the equivalent of https://docs.scipy.org/doc/numpy-dev/reference/generated/numpy.nanmean.html or R’s `na.rm=T` for the ones where the numeric value is NA. This could also be achieved by grabbing the margin and subtracting the sum of contents where the value is None.

src/cr/cube/utils/table_helper.py
```diff
@@ -0,0 +1,124 @@
+'''Grouping of various cube methods'''
+
+import numpy as np
+
+from ..dimension import Dimension
+
+
+class TableHelper(object):
+    '''Groups together useful cube utility methods.'''
```
percious: are tables part of a cube?  If this is a case, call it what it is, a DataTable or CubeTable and drop the Helper.  Also, perhaps create this lazily in Cube by attaching a cached property to the Cube object named table.

PS. I loath the term "Helper" for things, it's entirely non-descriptive.

src/cr/cube/crunch_cube.py
slobodan-ilic: Fixed issues with calculation in case of missing num vals.

src/cr/cube/utils/table_helper.py
slobodan-ilic: Addressed the nomenclature issues. Refactored a bit more. Also extracted `Means` and `Index` functionality into service classes.

slobodan-ilic:
Ok, all the tests are passing now, I've addressed all the comments. Can we merge this please?

tests/integration/test_crunch_cube.py
```diff
@@ -1950,3 +1957,80 @@ def test_cat_x_mr_x_mr_proportions_by_cell(test):
...
+            [1.48, 1.40740741],
+            [1.42857143, 1.53846154],
+            [1.52173913, 1.55319149],
+        ])
+        actual = cube.scale_means()
+        np.testing.assert_almost_equal(actual, expected)
+
+    def test_means_with_null_values(self):
+        cube = CrunchCube(FIXT_SCALE_WITH_NULL_VALUES)
+        expected = np.array([1.2060688, 1.0669344, 1.023199])
```
malecki: ✅ same as https://github.com/Crunch-io/whaam/blob/master/base/stats/tests/compute-score-table-spec.js#L61

PR #35, opened 2018-03-26
Hs for mr 155641308

slobodan-ilic:
Per [ticket](https://www.pivotaltracker.com/story/show/155641308)

### DIFF
--- a/src/cr/cube/crunch_cube.py
+++ b/src/cr/cube/crunch_cube.py
@@ -127,39 +127,38 @@ class CrunchCube(object):
 
     def _update_result(self, result, insertions, dimension_index,
                        valid_indices):
-        '''Actually insert subtotals into resulting ndarray.'''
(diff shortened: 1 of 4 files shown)


src/cr/cube/crunch_cube.py
```diff
@@ -405,7 +406,7 @@ def _margin(self, axis=None, weighted=True, adjusted=False,
 
         return res
 
-    def _mr_proportions(self, axis, weighted):
+    def _mr_proportions(self, axis, weighted, include_transforms_for_dims=None):
         '''Calculate MR proportions.'''
```
percious: include_transforms_for_dims is starting to feel like something we need to factor out.

slobodan-ilic: Created a ticket for this: https://www.pivotaltracker.com/story/show/156313137

PR #38, opened 2018-03-27
correct univariate_ca method

kwypchlo:
https://www.pivotaltracker.com/story/show/154613258

### DIFF
--- a/src/cr/cube/crunch_cube.py
+++ b/src/cr/cube/crunch_cube.py
@@ -475,8 +475,9 @@ class CrunchCube(object):
 
     @property
     def is_univariate_ca(self):
-        types = [d.type for d in self.dimensions]
-        return types == ['categorical_array', 'categorical']
+        types = {d.type for d in self.dimensions}
+        ca_types = {'categorical_array', 'categorical'}
+        return len(self.dimensions) == 2 and types == ca_types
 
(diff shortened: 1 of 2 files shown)


src/cr/cube/crunch_cube.py
```diff
@@ -475,8 +475,15 @@ def _mr_proportions(self, axis, weighted, include_transforms_for_dims=None):
 
     @property
     def is_univariate_ca(self):
-        types = [d.type for d in self.dimensions]
-        return types == ['categorical_array', 'categorical']
+        types = {d.type for d in self.dimensions}
```
slobodan-ilic: Docstring above this line would be nice (like inside of `'''blah blah'''`)

src/cr/cube/crunch_cube.py
```diff
@@ -475,8 +475,15 @@ def _mr_proportions(self, axis, weighted, include_transforms_for_dims=None):
...
+        ca_types = {'categorical_array', 'categorical'}
+        return len(self.dimensions) == 2 and types == ca_types
+
+    @property
+    def univariate_ca_main_axis(self):
+        # for univariate CA, the main axis is the categorical axis
```
slobodan-ilic: Also, needs to be in the form of a docstring (between `''' and '''`)

PR #39, opened 2018-04-03
Export messed up for cubes with HS and Pruning

kwypchlo:
https://www.pivotaltracker.com/story/show/155080499

### DIFF
--- a/src/cr/cube/crunch_cube.py
+++ b/src/cr/cube/crunch_cube.py
@@ -196,52 +196,77 @@ class CrunchCube(object):
         res = res + adjusted
 
         if prune:
-            return self._prune(res, include_transforms_for_dims)
+            return self._prune_body(res, include_transforms_for_dims)
 
         return res
 
-    def _mr_prune(self, res):
-        if len(res.shape) > 2:
-            # Only perform pruning for 1-D MR cubes.
-            return res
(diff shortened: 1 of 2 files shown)


src/cr/cube/crunch_cube.py
```diff
@@ -752,11 +777,21 @@ def labels(self, include_missing=False, include_transforms_for_dims=False):
         Returns
             labels (list of lists): Labels for each dimension
         '''
-        return [
+        labels = [
             dim.labels(include_missing, include_transforms_for_dims)
             for dim in self.dimensions
         ]
 
+        if prune:
```
slobodan-ilic: Could you extract all that's under this `if` in a separate method, and possibly call it `_prune_labels`?

slobodan-ilic:
I made a comment about labels pruning, and then saw that it's refactored. Nothing to add, lgtm.

kwypchlo:
@percious Coverage increased (+0.005%) to 99.842% when pulling 4792eb8 on hs-pruning-155080499 into c4e7ff9 on master. :)

kwypchlo:
@percious I have not added any new logic/functionalities so I relied on the old tests to cover the changes. If the code was covered and was passing before, and the same tests pass now too, it means everything is tested.

PR #45, opened 2018-04-16
fix index for mr x mr and add tests

kwypchlo:
https://www.pivotaltracker.com/story/show/156472341


src/cr/cube/measures/index.py
```diff
@@ -46,13 +46,30 @@ def _mr_index(self, weighted):
         selected = table[self.cube.ind_selected]
         non_selected = table[self.cube.ind_non_selected]
 
-        if self.cube.mr_dim_ind == 0 or self.cube.mr_dim_ind == (0, 1):
+        # mr by mr
+        if self.cube.mr_dim_ind == (0, 1) or self.cube.mr_dim_ind == (1, 2):
+            row_proportions = self.cube.proportions(axis=0,
```
slobodan-ilic: This is not `row_proportions`, but `col_proportions`, because you're calculating it with `axis=0`. At least it should be called that way (the variable). If it doesn't produce the expected value, than that's a bug and should be fixed...

src/cr/cube/measures/index.py
```diff
@@ -46,13 +46,30 @@ def _mr_index(self, weighted):
...
+                                                    weighted=self.weighted,
+                                                    prune=self.prune)
+            col_proportions = self.cube.proportions(axis=1,
+                                                    weighted=self.weighted,
+                                                    prune=self.prune)
+            return col_proportions / row_proportions
+
+        # mr by cat
+        if self.cube.mr_dim_ind == 0:
+            proportions = self.cube.proportions(axis=0,
```
slobodan-ilic: Two very similar pieces of code. You could extract the axis like:
```
axis = self.cube.mr_dim_ind
proportions = self.cube.proportions(axis=axis...)
margin = np.sum(selected, axis=axis) / . . .
```

PR #51, opened 2018-05-03
3d missing 156668130

slobodan-ilic:
Per [ticket](https://www.pivotaltracker.com/story/show/156668130)

PR #52, opened 2018-05-04
[#156848737]: Fix valid indices where already applied

slobodan-ilic:
Per [ticket](https://www.pivotaltracker.com/story/show/156848737)

This is a simple fix, but _will_ need to be rebased on #51 (or the other way round)


src/cr/cube/crunch_cube.py
```diff
@@ -700,7 +703,8 @@ def _mr_proportions(self, axis, weighted, prune,
                 den = np.sum(num + non_selected, axis)
             res = num / den
             res = self._transform(
-                res, include_transforms_for_dims, self.valid_indices
+                # res, include_transforms_for_dims, self.valid_indices
```
percious: turd

tests/integration/test_multiple_response.py
```diff
@@ -664,3 +665,72 @@ def test_mr_x_num_cols_margin_pruned(self):
         )
         for actual in margin:
             np.testing.assert_array_equal(actual, expected)
+
+    def test_num_x_mr_props_by_row(self):
```
percious: I'd really rather see a unit test rather than an integration test for this fix.  Perhaps you can add some unit testing in this area and remove this integration test.

PR #54, opened 2018-05-04
fix formatting hs in 3d cubes

kwypchlo:
https://www.pivotaltracker.com/story/show/157293490

### DIFF
--- a/src/cr/cube/dimension.py
+++ b/src/cr/cube/dimension.py
@@ -113,18 +113,27 @@ class Dimension(object):
     @property
     def inserted_hs_indices(self):
         if self.type == 'categorical_array':
-            # For CA subvariables, we don't do H&S insertions.
-            return []
+            return []  # For CA subvariables, we don't do H&S insertions
+
+        element_ids = [element['id'] for element in self.elements()]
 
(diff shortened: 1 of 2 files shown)


src/cr/cube/dimension.py
```diff
@@ -113,18 +113,27 @@ def _elements(self):
...
         bottoms = [st for st in self.subtotals if st.anchor == 'bottom']
         middles = [st for st in self.subtotals if st.anchor not in ['top', 'bottom']]
 
-        top_inds = [i for i, _ in enumerate(tops)]
-        middle_inds = [i + m.anchor + len(tops) for i, m in enumerate(middles)]
-        bottom_inds = [i + len(tops) + len(middles) + len(self.elements())
-                       for i, _ in enumerate(bottoms)]
-        return top_inds + middle_inds + bottom_inds
+        top_indexes = [
+            index for index, insertion in enumerate(tops)
```
slobodan-ilic: doesn't the linter complain here? You never use `insertion`. It's _pythonic_ to replace it with `_`... AFAIK. Maybe @percious would have a different take on it?

PR #62, opened 2018-05-16
fix margin calculations when transformations are applied

kwypchlo:
https://www.pivotaltracker.com/story/show/157556884

### DIFF
--- a/src/cr/cube/crunch_cube.py
+++ b/src/cr/cube/crunch_cube.py
@@ -603,20 +603,17 @@ class CrunchCube(object):
         return array
 
     def _margin_transform_dims(self, axis, include_transforms_for_dims):
-        # FIXME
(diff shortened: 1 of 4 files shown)


src/cr/cube/crunch_cube.py
```diff
@@ -603,20 +603,17 @@ def _inflate_dim(self, array, axis):
...
-                if axis is not None and isinstance(axis, int) else
-                None
-            )
-        return include_transforms_for_dims
+        """When calculating a margin that needs to sum up counts, make sure
+        we do not include subtotals if they are on the opposite axis
+        of the one we are calculating, otherwise they will be summed up
+        with the counts and the values will be wrong."""
+
+        dim_offset = 1 if len(self.dimensions) == 3 else 0
```
slobodan-ilic: You can use `self.ndim` here...

PR #65, opened 2018-05-22
[#157772177]: Initial commit for CubeSlice

slobodan-ilic:
Per [ticket](https://www.pivotaltracker.com/story/show/157772177)

- Write unit tests for CubeSlice (always do only TDD)
- Write the CubeSlice class

### DIFF
--- a/src/cr/cube/cube_slice.py
+++ b/src/cr/cube/cube_slice.py
@@ -50,6 +50,7 @@ class CubeSlice(object):
     def _increment_axis(axis):
         if axis is not None and not isinstance(axis, tuple):
             axis += 1
+        # return (1, 2)
         return axis
 
     def as_array(self, *args, **kwargs):
--- a/src/cr/cube/cube_slice.py
+++ b/src/cr/cube/cube_slice.py
@@ -1,0 +1,18 @@
+'''Home of the CubeSlice class.'''
+
+from __future__ import division
+
+from .crunch_cube import CrunchCube
+
+
+class CubeSlice(CrunchCube):
+    '''Implementation of CubeSlice class.
+
+    The CubeSlice is used to uniformly represent all cubes as 2D tables.
+    For 1D cubes, this is achieved by inflating. For 3D cubes, this is
+    achieved by slicing.
+    '''
+
+    def __init__(self, response, index=None):
+        super(CubeSlice, self).__init__(response)
+        self._index = index
--- a/src/cr/cube/cube_slice.py
+++ b/src/cr/cube/cube_slice.py
@@ -1,9 +1,5 @@
 '''Home of the CubeSlice class.'''
 
-from __future__ import division
-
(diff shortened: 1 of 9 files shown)


src/cr/cube/cube_slice.py
```diff
@@ -0,0 +1,145 @@
...
+        table_name = self._cube.labels()[0][self._index]
+        return '%s: %s' % (title, table_name)
+
+    @property
+    def rows_title(self):
+        '''Get title of the rows dimension.
+
+        For 3D it's the 1st dimension (0th dimension of the current slice).
+        '''
+        return self._cube.dimensions[1].name
```
jimfulton: But what if it's not 3d?

src/cr/cube/cube_slice.py
```diff
@@ -0,0 +1,145 @@
...
+        '''Get has_means from cube.'''
+        return self._cube.has_means
+
+    @property
+    def dimensions(self):
+        '''Get slice dimensions.
+
+        For 2D cubes just get their dimensions. For 3D cubes, don't get the
+        first dimension (only take slice dimensions).
+        '''
```
jimfulton: What about 1d cubes?

src/cr/cube/cube_slice.py
```diff
@@ -0,0 +1,145 @@
...
+    The CubeSlice is used to uniformly represent all cubes as 2D tables.
+    For 1D cubes, this is achieved by inflating. For 3D cubes, this is
+    achieved by slicing.
+    '''
+
+    def __init__(self, cube, index):
+        self._cube = cube
+        self._index = index
+
+    def _update_args(self, kwargs):
```
jimfulton: I would inline this and the next function. They're only called from one place. This would potentially make the logic in `_call_cube_method` clearer.

tests/integration/test_cube_slice.py
```diff
@@ -0,0 +1,18 @@
+# from unittest import TestCase
```
jimfulton: Um, ... :)

tests/integration/test_cube_slice.py
```diff
@@ -0,0 +1,18 @@
...
+# # from .fixtures import PETS_ARRAY_X_PETS
+# # from .fixtures import PETS_X_PETS_ARRAY
+
+
+# class TestCubeSlice(TestCase):
+#     def test_cube_slice_(self):
+#         cs = CubeSlice(ECON_GENDER_X_IDEOLOGY_WEIGHTED)
+#         expected = CAT_X_CAT['value']
+#         actual = cube._table._cube
+#         self.assertEqual(actual, expected)gg
```
jonkeane: If this test is covered elsewhere, I vote that we get rid of this file. However, I suspect that we do want to do some testing of the cube slicing machinery in general here, no?

PR #67, opened 2018-05-25
fix mr x cat margins

kwypchlo:
https://www.pivotaltracker.com/story/show/157874314


src/cr/cube/crunch_cube.py
```diff
@@ -386,25 +380,24 @@ def row_direction_axis(self):
...
     @staticmethod
-    def _margin_pruned_indices(margin, insertions):
-        ind_inserted = np.zeros(margin.shape, dtype=bool)
-        if insertions is not None and any(insertions):
-            ind_inserted[insertions] = True
+    def _margin_pruned_indices(margin, inserted_ind, axis):
         pruned_ind = np.logical_or(margin == 0, np.isnan(margin))
-        pruned_ind = np.logical_and(pruned_ind, ~ind_inserted)
+
+        if np.any(margin) and inserted_ind is not None and any(inserted_ind):
```
slobodan-ilic: It would be more elegant if you just returned `pruned_ind` on the inverted condition here, and had the rest of the code outdented. At least at first :eyes:

kwypchlo: I feel like it nicely outlines that the `pruned_ind` is basically starting with the same matrix but if there are insertions we are masking these inserted indices

PR #69, opened 2018-06-11
Refactor margin #156801059

slobodan-ilic:
Per [ticket](https://www.pivotaltracker.com/story/show/156801059)


src/cr/cube/crunch_cube.py
```diff
@@ -449,9 +481,13 @@ def _adjust_axes(self, axes):
 
         For more info on how it needs to operate, check the unit tests.
         '''
+
+        if not self._is_axes_allowed(axes):
```
kwypchlo: `_is_axes_allowed` -> `_validate_axis` IMO

src/cr/cube/crunch_cube.py
```diff
@@ -931,6 +784,37 @@ def as_array(self, include_missing=False, weighted=True, adjusted=False,
...
+        prune (because items will never be pruned). But we still need a result
+        to be able to perform subsequent operations, without explicitly
+        checking if the axes direction is across items dimension each time.
+        '''
+
+        if not self._is_axes_allowed(axis):
+            # In case we encountered axis that would go across items dimension,
+            # we need to return at least some result, to prevent explicitly
+            # checking for this condition, wherever self._margin is used
+            return self.as_array(
```
kwypchlo: returning counts when requesting margin is a big nono imo

src/cr/cube/crunch_cube.py
```diff
@@ -1207,11 +1130,11 @@ def index(self, weighted=True, prune=False):
...
         counts = self.as_array(weighted=weighted, prune=prune)
-        total = self.margin(weighted=weighted, prune=prune)
-        colsum = self.margin(axis=0, weighted=weighted, prune=prune)
-        rowsum = self.margin(axis=1, weighted=weighted, prune=prune)
+        total = self._margin(weighted=weighted, prune=prune)
+        colsum = self._margin(axis=0, weighted=weighted, prune=prune)
+        rowsum = self._margin(axis=1, weighted=weighted, prune=prune)
 
-        if self.has_mr:
+        if self.has_mr or 'categorical_array' in self.dim_types:
```
kwypchlo: can you use something like `self.has_ca` ?

src/cr/cube/crunch_cube.py
```diff
@@ -398,7 +397,8 @@ def _margin_pruned_indices(margin, inserted_ind, axis):
 
     @lazyproperty
     def col_direction_axis(self):
-        return self.ndim - 2
+        axis = self.ndim - 2
+        return axis
```
kwypchlo: just `return self.ndim - 2`

src/cr/cube/crunch_cube.py
slobodan-ilic: Well, IDK, this is more logical to me... I know that we're not adhering to any paradigm, but in a perfect world (read _functional_) you'd have a pure function. So the name of this one tries to depict that (it will return either `True` or `False`).

`_validate_axis` could be considered something as: check if they're OK and fix them if they're not (which I tried to avoid). But if you feel strongly for `validate` approach, I don't disagree strongly (just a little bit) :)

src/cr/cube/crunch_cube.py
slobodan-ilic: I know... That's why the API method will never do that. One other approach is calling this method `_margin_or_counts`. Or something like `base`, or something. I'd much prefer to keep this in a single method to having to write
```
self.margin(axis=0) if self.dim_types[0] != 'categorical_array' else self.as_array()
```

src/cr/cube/crunch_cube.py
slobodan-ilic: will do

src/cr/cube/crunch_cube.py
slobodan-ilic: sure

src/cr/cube/crunch_cube.py
kwypchlo: but why would you return the counts even? it will produce wrong numbers, there is absolutely no reason in doing that

src/cr/cube/crunch_cube.py
kwypchlo: at least it should be `_are_axes_allowed`

I feel like we overload the word axis too much btw, I cannot say just by the name and context of this method, where this method is applicable and where not (cuz we have a lot of places that have `axis` as a param)

PR #71, opened 2018-06-15
Sig test 3d tables 157314298

slobodan-ilic:


src/cr/cube/crunch_cube.py
```diff
@@ -1181,3 +1188,9 @@ def pvals(self, weighted=True, prune=False, hs_dims=None):
     def scale_means(self):
         '''Get cube means.'''
         return ScaleMeans(self).data
+
+    def _get_slices(self):
+        if self.ndim < 3:
+            return [CubeSlice(self, 0)]
+
+        return [CubeSlice(self, i) for i, _ in enumerate(self.as_array())]
```
kwypchlo: replace `enumerate(self.as_array())` with `range(len(cube.labels()[0]))`
`labels` doesn't involve any calculations compared to `as_array`

slobodan-ilic: Sure, but I'd like to keep the `enumerate` :smile:

PR #75, opened 2018-06-21
Enable pruning in crosstabs #158458376

slobodan-ilic:


src/cr/cube/cube_slice.py
```diff
@@ -150,3 +150,33 @@ def ca_main_axis(self):
...
+            '''Get pruned labels for single dimension, besed on prune inds.'''
+            labels = [
+                label for label, prune in zip(labels, prune_indices)
+                if not prune
+            ]
+            return labels
+
+        prune_indices = self.prune_indices(transforms=hs_dims)
+        labels = [
+            prune_dimension_labels(dim_labels, dim_prune_inds)
```
jimfulton: Why not just inline this? The function logic is trivial, a single expression.

The over-optimizer in me bristles at the function call overhead and the expense of creating an inline function that's not a closure. :)

But the mental overhead of having to consider the function and this expression separately is the bigger cost, IMO.

src/cr/cube/subtotal.py
```diff
@@ -21,7 +21,10 @@ def is_valid(self):
             required_keys = {'anchor', 'args', 'function', 'name'}
             has_keys = set(self._data.keys()) == required_keys
             if has_keys:
-                return self._data['function'] == 'subtotal'
+                is_subtotal = self._data['function'] == 'subtotal'
```
jimfulton: If you move the `subtotal` function check into the `if` above, you can avoid doing work of the 2 lines below when it's false.

src/cr/cube/subtotal.py
```diff
@@ -43,10 +46,15 @@ def anchor(self):
     def _all_dim_ids(self):
         return [el.get('id') for el in self._dim.elements(include_missing=True)]
 
+    def _validate_hs_ids(self, hs_ids):
+        element_ids = [el['id'] for el in self._dim.elements()]
+        return any(arg in element_ids for arg in hs_ids)
```
jimfulton: Why not just: `[el for el in self._dim.elements() if el['id'] in hs_ids]`?

Why not just inline this in `is_valid`? It would make `is_valid` easier to read.

(If your answer would be, it's used in `args` below, see my comment there. :)

src/cr/cube/subtotal.py
```diff
@@ -43,10 +46,15 @@ def anchor(self):
...
+
     @lazyproperty
     def args(self):
         '''Get H&S args.'''
-        if self.is_valid:
+        hs_ids = self._data.get('args', None)
+        if hs_ids and self.is_valid and self._validate_hs_ids(hs_ids):
```
jimfulton: `and self._validate_hs_ids(hs_ids)` is redundant, as that check is done in `is_valid`.

src/cr/cube/subtotal.py
```diff
@@ -43,10 +46,15 @@ def anchor(self):
     def _all_dim_ids(self):
         return [el.get('id') for el in self._dim.elements(include_missing=True)]
 
+    def _validate_hs_ids(self, hs_ids):
+        element_ids = [el['id'] for el in self._dim.elements()]
```
jimfulton: You're creating a list when you could have used a generator, but ...

src/cr/cube/cube_slice.py
slobodan-ilic: Done

src/cr/cube/subtotal.py
slobodan-ilic: :+1:

PR #81, opened 2018-07-09
Scale means #154065274

slobodan-ilic:


src/cr/cube/measures/scale_means.py
```diff
@@ -19,15 +19,17 @@ def data(self):
         contents = self._inner_prod(table, self.values)
 
         if self._cube.has_mr and not self._cube.is_double_mr:
-            axis = 1 - self._cube.mr_dim_ind
+            # axis = 1 - self._cube.mr_dim_ind
+            # axis = self._cube.dim_types[-2:].index('categorical')
+            axis = self._cube.dim_types.index('categorical')
             return np.sum(contents, axis) / np.sum(table, axis)
```
jonkeane: I commented to this end on the exporter PR for this change, I think we should be able to specify which axis the means are to be calculated for in crunch-cube and have exporter explicitly ask for the axis it wants/needs rather than having the special cases here. 

In the circumstance where an axis is requested that is ineligible for means (a multiple response, or a categorical without numeric values), you can either return all `NaNs` or an empty array.

tests/integration/test_scale_means.py
```diff
@@ -0,0 +1,81 @@
...
+
+def test_cat_x_ca_cat_x_items():
+    cube = CrunchCube(CAT_X_CA_CAT_X_ITEMS)
+    expected = np.array([
+        [1.34545455, 2.46938776, 2.7037037 , 2.65454545],  # noqa
+        [1.41935484, 3.25663717, 3.48      , 3.58536585],  # noqa
+        [1.49429038, 3.44905009, 3.59344262, 3.53630363],
+        [1.43365696, 3.02816901, 3.37987013, 3.32107023],
+        [1.22670025, 2.49473684, 2.79848866, 2.78987342],
+        [2.53061224, 3.68421053, 3.9862069 , 4.03472222],  # noqa
```
jonkeane: These are the means for each slice of the higher-dimension array, yeah? I'm a little bit surprised that they are provided in a single np.array like this, or is this an artifact of testing?

tests/integration/test_scale_means.py
```diff
@@ -0,0 +1,81 @@
...
+    ])
+    actual = cube.scale_means()
+    np.testing.assert_almost_equal(actual, expected)
+
+
+def test_cat_x_cat():
+    cube = CrunchCube(CAT_X_CAT)
+    expected = np.array([2.61411765, 2.34285714, 2.33793103, 3.38461538])
```
jonkeane: These values look good and conform with R

tests/integration/test_scale_means.py
```diff
@@ -0,0 +1,81 @@
...
+    actual = cube.scale_means()
+    np.testing.assert_almost_equal(actual, expected)
+
+
+def test_cat_x_mr():
+    cube = CrunchCube(CAT_X_MR)
+    expected = np.array([2.45070423, 2.54471545, 2.54263006, np.nan])
```
jonkeane: 👍

tests/integration/test_scale_means.py
```diff
@@ -0,0 +1,81 @@
...
+    actual = cube.scale_means()
+    np.testing.assert_almost_equal(actual, expected)
+
+
+def test_mr_x_cat():
+    cube = CrunchCube(MR_X_CAT)
+    expected = np.array([2.45070423, 2.54471545, 2.54263006, np.nan])
```
jonkeane: 👍

tests/integration/test_scale_means.py
```diff
@@ -0,0 +1,81 @@
...
+    actual = cube.scale_means()
+    np.testing.assert_almost_equal(actual, expected)
+
+
+def test_univariate_cat():
+    cube = CrunchCube(UNIVARIATE_CAT)
+    expected = np.array([2.6865854])
```
jonkeane: 👍

PR #96, opened 2018-08-28
Ponder population size #159762881

slobodan-ilic:


tests/unit/test_crunch_cube.py
```diff
@@ -1,4 +1,6 @@
 '''Unit tests for the CrunchCube class.'''
+
+import pytest
```
percious: where is this import used?

tests/unit/test_crunch_cube.py
```diff
@@ -930,4 +932,39 @@ def test_ca_dim_ind_is_none(self):
         cc = CrunchCube({})
         actual = cc.ca_dim_ind
         expected = None
-        assert actual == expected
\ No newline at end of file
+        assert actual == expected
+
+    def test_population_fraction(self):
```
percious: ❤️

PR #100, opened 2018-09-24
Scale mean marginal 160719473

slobodan-ilic:


tests/unit/test_cube_slice.py
```diff
@@ -356,7 +356,7 @@ def test_ca_dim_ind(self):
 
         cube.ca_dim_ind = None
         cs = CubeSlice(cube, 0)
-        assert cs.ca_dim_ind == None
+        assert cs.ca_dim_ind is None
```
percious: ❤️

tests/integration/test_scale_means.py
```diff
@@ -237,7 +242,18 @@ def test_cat_x_cat_pruning_and_hs():
...
+
+def test_cat_x_cat_scale_means_margin():
+    cs = CrunchCube(CAT_X_CAT_SM_MARGIN['slides'][0]['cube']).slices[0]
+    expected = 2.6846246973365617
+    actual = cs.scale_means_margin(1)
+    assert actual == expected
+
+    expected = 2.536319612590799
+    actual = cs.scale_means_margin(0)
+    assert actual == expected
```
jonkeane: Are these the only new expectations? I scrolled through and it looked like most of the ones above were just `#noqa` additions, but I want to make sure that I'm not missing something!

tests/unit/test_scale_means.py
```diff
@@ -0,0 +1,50 @@
+from __future__ import division
```
percious: is there a test for single-dimensional analysis of scale means?  I know this is probably not a thing, but the users will do it and I think it would be best not to error in this case.

tests/integration/fixtures/cubes/scale_means/cat-x-cat-sm-margin.json
```diff
@@ -0,0 +1,376 @@
+{
```
jonkeane: This looks like a multitable export, but I'm not sure it should be. It's not a _big deal_ but a CrunchCube response alone should be sufficient here and be a little bit less distracting as to what's going on.

tests/integration/test_scale_means.py
jonkeane: These values, however are good!

```
> cube <- loadCube("~/Desktop/cat-x-cat-sm-margin.json") # edited to extract just the cube
> 
> # get the category values
> categories(variables(cube)[[1]])
     id        name value missing
1     1    Bucatini     1   FALSE
2     2    Chitarra     2   FALSE
3     0     Boccoli     0   FALSE
4     4 Orecchiette     4   FALSE
5     5 Quadrefiore     5   FALSE
6     6      Fileja     6   FALSE
7 32766     Skipped 32766    TRUE
8 32767   Not asked 32767    TRUE
9    -1     No Data    NA    TRUE
> buc <- 1
> chi <- 2
> boc <- 0
> ore <- 4
> qua <- 5
> fil <- 6
> 
> # calcualte the univariate mean for rows
> sum(margin.table(cube, 1) * c(buc, chi, boc, ore, qua, fil)) / margin.table(cube)
[1] 2.684625
> 
> categories(variables(cube)[[2]])
     id       name value missing
1     0 Vegetables     0   FALSE
2 32766 Don't know 32766    TRUE
3     2      Fruit     2   FALSE
4     5      Grain     5   FALSE
5    -1    No Data    NA    TRUE
6     4       Meat     4   FALSE
7 32767  Not asked 32767    TRUE
> veg <- 0
> fruit <- 2
> grain <- 5
> meat <- 4
> 
> # calcualte the univariate mean for cols
> sum(margin.table(cube, 2) * c(veg, fruit, grain, meat)) / margin.table(cube)
[1] 2.53632
```

slobodan-ilic: Yup, these are the only expectations. We can add more when we do exporter tests. But I wanted to keep it as simple as possible for the first iteration.

tests/integration/test_scale_means.py
```diff
@@ -22,7 +22,7 @@
 
 def test_ca_cat_x_items():
     cube = CrunchCube(CA_CAT_X_ITEMS)
-    expected = [[np.array([1.50454821, 3.11233766, 3.35788192, 3.33271833]), None]]
+    expected = [[np.array([1.50454821, 3.11233766, 3.35788192, 3.33271833]), None]]  # noqa
```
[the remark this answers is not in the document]
slobodan-ilic: sure, will do

tests/integration/fixtures/cubes/scale_means/cat-x-cat-sm-margin.json
slobodan-ilic: It's a deck export. I can make it a pure cube JSON.

tests/integration/test_scale_means.py
percious: i'd rather have more tests here and less in exporter.  Try to keep the tests as close to the determination of values as possible.

slobodan-ilic: Done, I've updated the existing tests with calls to new functionality.

tests/unit/test_scale_means.py
slobodan-ilic: There is now. It's not supposed to happen, and we raise explicitly.

tests/integration/fixtures/cubes/scale_means/cat-x-cat-sm-margin.json
slobodan-ilic: Changed to explicit cube.

PR #101, opened 2018-09-26
Deprecate CubeSlice shape property

slobodan-ilic:
- Deprecation warning when `shape` property is used
- Implement `get_shape` method, that uses `prune` argument
- Unit tests

### DIFF
--- a/src/cr/cube/cube_slice.py
+++ b/src/cr/cube/cube_slice.py
@@ -2,10 +2,10 @@
 
 from functools import partial
 import numpy as np
+import warnings
 
 from cr.cube.measures.scale_means import ScaleMeans
-
-from .utils import lazyproperty
+from .utils import lazyproperty, compress_pruned
 
 
 # pylint: disable=too-few-public-methods
@@ -262,7 +262,17 @@ class CubeSlice(object):
 
     @lazyproperty
     def shape(self):
-        return self.as_array().shape
+        warnings.warn('Deprecated. Use `get_shape` instead.', DeprecationWarning)
+        return self.get_shape()
 
     def scale_means_margin(self, axis):
         return ScaleMeans(self).margin(axis)
+
+    def get_shape(self, prune=False):
+        if not prune:
+            return self.as_array().shape
+
+        shape = compress_pruned(self.as_array(prune=True)).shape
+        # Eliminate dimensions that get reduced to 1
+        # (e.g. single element categoricals)
+        return tuple(n for n in shape if n > 1)
--- a/src/cr/cube/cube_slice.py
+++ b/src/cr/cube/cube_slice.py
@@ -1,8 +1,8 @@
 '''Home of the CubeSlice class.'''
 
 from functools import partial
-import numpy as np
 import warnings
+import numpy as np
 
 from cr.cube.measures.scale_means import ScaleMeans
 from .utils import lazyproperty, compress_pruned
@@ -34,6 +34,7 @@ class CubeSlice(object):
 
     @lazyproperty
     def col_dim_ind(self):
+        """Return 1 if not categorical array as 0th, 0 otherwise."""
         return 1 if not self.ca_as_0th else 0
 
     def __getattr__(self, attr):
@@ -160,6 +161,19 @@ class CubeSlice(object):
 
     @lazyproperty
     def ca_dim_ind(self):
+        """Return items dimension index if there is one.
+
+        If the slice is a part of a cube that has Categorical Array variable,
+        return the index of the items dimension (if it belongs to the slice).
+        Examples:
+        - For a CA(items) x CAT => returns 0
+        - For CAT x CA(items) => returns 1
+        - For CAT x CA(items) x CAT => returns 0 (because the items is the 0th
+          dimension of each slice)
+        - For CA(items) x CAT x CAT => returns None (because the 0th items
+          dimension doesn't belong to any one slice, and is itself used for
+          slicing the cube).
+        """
         index = self._cube.ca_dim_ind
         if index is None:
             return None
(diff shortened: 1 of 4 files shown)


src/cr/cube/cube_slice.py
```diff
@@ -258,4 +259,14 @@ def ndim(self):
 
     @lazyproperty
     def shape(self):
-        return self.as_array().shape
+        warnings.warn('Deprecated. Use `get_shape` instead.', DeprecationWarning)
+        return self.get_shape()
+
+    def get_shape(self, prune=False):
```
scanny: Please add docstring (PEP 257) describing interface and what a shape is.

slobodan-ilic: Done. I've run the `pylint` on the entire file, should be better now...

src/cr/cube/cube_slice.py
```diff
@@ -250,12 +265,43 @@ def is_double_mr(self):
...
+        This property is anloguous to numpy's 'shape' property of ndarray. It
+        returns a tuple, returning the shape of the slice's 'as_array' method.
+        This property is deprecated, use 'get_shape' instead. Pruning is not
+        supported (supported in 'get_shape').
+        """
+        deprecation_msg = 'Deprecated. Use `get_shape` instead.'
+        warnings.warn(deprecation_msg, DeprecationWarning)
+        return self.get_shape()
+
+    def get_shape(self, prune=False):
```
percious: memoize this.

src/cr/cube/cube_slice.py
```diff
@@ -250,12 +265,43 @@ def is_double_mr(self):
...
+        """
         return min(self._cube.ndim, 2)
 
     @lazyproperty
     def shape(self):
-        return self.as_array().shape
+        """Return the shape of the slice.
+
+        This property is anloguous to numpy's 'shape' property of ndarray. It
+        returns a tuple, returning the shape of the slice's 'as_array' method.
```
scanny: It would always be preferable to avoid re-using the name of the method in its docstring. If a person doesn't know what "shape" means in this context, this docstring is going to make them work for it. So in this case, the second sentence might be better as: *"It returns a tuple of ints, each representing the length of a cube dimension, in the order those dimensions appear in the cube."*

I would add a blank line before the deprecated sentences. That topic is separate enough to warrant a distinct paragraph.

tests/unit/test_cube_slice.py
```diff
@@ -379,9 +380,52 @@ def test_ca_dim_ind(self):
...
+        """Test if slice method invokes cube method."""
         mock_sm_init.return_value = None
 
         cs = CubeSlice({}, 0)
         fake_axis = Mock()
         cs.scale_means_margin(fake_axis)
         assert mock_sm_margin.called_once_with(fake_axis)
+
+    def test_shape(self):
+        """Test shape based on 'as_array' and pruning."""
```
scanny: This looks like a good candidate for a parameterized test, maybe something like this:

```python
def it_knows_its_shape(self, shape_fixture):  # ---or named `test_shape()` if you prefer---
    cube_slice, prune, expected_shape = shape_fixture
    shape = cube_slice.get_shape(prune=prune)
    assert shape == expected_value

@pytest.fixture(params=[
    (False, None, (3, 2)),
    (True, [[True, False], [True, False], [True, False]], (3,)),
    etc.
])
def shape_fixture(self, request, cube_):
    prune, mask_map, expected_value = request.param
    cube_.ndim = 2
    cube_.as_array.return_value = np.ma.masked_array(
        np.zeros((3, 2)), mask=np.array(mask_map) if mask_map else None
    )
    cube_slice = CubeSlice(cube_, 0)
    return cube_slice_, prune, expected_value
```

This isn't complete (or even correct) in every detail, but maybe gives you an idea. I recommend making the warning test a separate test method since its form and asserts are very different.

Using the fixture allows the test to be very clear and adding new cases is trivial and doesn't duplicate code.

src/cr/cube/cube_slice.py
```diff
@@ -252,15 +266,65 @@ def is_double_mr(self):
...
     def shape(self):
-        return self.as_array().shape
+        """Tuple of array dimensions' lengths.
+
+        It returns a tuple of ints, each representing the length of a cube
+        dimension, in the order those dimensions appear in the cube.
+
+        This property is deprecated, use 'get_shape' instead. Pruning is not
+        supported (supported in 'get_shape').
+        """
```
scanny: Sweet :heart: :)

src/cr/cube/cube_slice.py
```diff
@@ -252,15 +266,65 @@ def is_double_mr(self):
         return self.dim_types == ['multiple_response'] * 2
 
     def scale_means(self, hs_dims=None, prune=False):
+        """Calculate scale means of the CubeSlice."""
```
scanny: It's a fine point, but I always use the word "Return" as the first word of the docstring for a method that returns a value. So something like "Return scale-means for this cube slice." This is the key aspect of the contract of such a method (the return value and type) and what the caller cares about most. Saying "Calculate" could just indicate a side-effect and leaves the reader to dig into the code (which often they won't have when just looking at the docs).

As an aside, in the case of a property, I leave the 'Return' word out, like "int count of foos in bar" instead of 'Return int count of foos in bar', since the semantics of a property are a value rather than a procedure.

You'll also need to say what scale-means are and what form you deliver them in. If that's not compact enough for the summary line, then in a body paragraph of the docstring. Even then, it's worth every effort not to repeat the method name in the summary because it provides no additional information. It could be something like """Return list containing scaled-mean for each slice column.""" or something and already starts to give more insight into the method contract.

PR #103, opened 2018-10-02
Pre commit hooks 160906189

slobodan-ilic:


setup.py
```diff
@@ -19,6 +19,7 @@ def get_long_desc():
     'pytest',
     'pytest-cov',
     'mock',
+    'pre-commit',
```
bertjwregeer: I don't think that this is something we should add to setup.py for this project. pre-commit is something the developer has to have installed on their machine globally, it is NOT a dependency of this project, not even a dev dependency IMHO.

.gitignore
```diff
@@ -108,4 +108,4 @@ ENV/
 *.~
 
 # idea
-.idea
\ No newline at end of file
+.idea
```
scanny: This change is developer environment-specific so belongs in your `~/.config/git/ignore` rather than the project `.gitignore`. (Local `.gitignore` location varies a bit based on your configuration, yours might be `~/.gitignore` or something.)

bertjwregeer: I am with @scanny on this one. Developers should have local gitignores that block those files from being committed.

bertjwregeer: That being said, it's not a change, the file just has a newline added. The gitignore already had that in it.

bertjwregeer:
Excellent! This'll be great!

PR #107, opened 2018-10-12
Fix sentry 161120801

scanny:
See last commit for actual fix.

All the rest is getting the code base and documentation normalized and PEP8-ed.

Refactored to remove a couple modules and fixed pesky numpy warnings. Replaced vulnerable lazyproperty implementation.


src/cr/cube/crunch_cube.py
```diff
@@ -80,27 +76,26 @@ def __init__(self, response):
 
     def as_array(self, include_missing=False, weighted=True, adjusted=False,
                  include_transforms_for_dims=None, prune=False, margin=False):
-        """Get crunch cube as ndarray.
+        """Return `ndarray` representing cube values.
```
slobodan-ilic: ❤️

src/cr/cube/measures/scale_means.py
```diff
@@ -68,13 +68,24 @@ def margin(self, axis):
         return np.sum(values * margin) / total
 
     def valid_indices(self, axis):
+        # --there is an interaction with CrunchCube._fix_shape() which
```
slobodan-ilic: We might want to 🎫 this abstraction. I'm thinking of either `collapsed_dimensions` or `hidden_dimensions` or both (in respective contexts). But yeah, this abstraction definitely needs to happen.

src/cr/cube/crunch_cube.py
```diff
@@ -78,1137 +74,1260 @@ def __init__(self, response):
...
-        if den.shape[0] == 1 and len(den.shape) > 1 and self.ndim < 3:
-            den = den.reshape(den.shape[1:])
-        return den
+    @classmethod
+    def _calculate_constraints_sum(cls, prop_table, prop_margin, axis):
+        """Calculate sum of constraints (part of the standard error equation).
 
-    def proportions(self, axis=None, weighted=True,
-                    include_transforms_for_dims=None, include_missing=False,
-
```
slobodan-ilic: This one should go above the line just before it.

src/cr/cube/crunch_cube.py
```diff
@@ -9,17 +9,17 @@
...
 from scipy.stats import norm
 from scipy.stats.contingency import expected_freq
 
-from .mixins.data_table import DataTable
-from .cube_slice import CubeSlice
-from .measures.index import Index
-from .measures.scale_means import ScaleMeans
-from .utils import lazyproperty
-from . import ITEM_DIMENSION_TYPES
+from cr.cube import ITEM_DIMENSION_TYPES
```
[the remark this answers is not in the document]
scanny: Ok, I checked with Bert, he says it's about personal preference. I find the argumentation to avoid relative imports compelling, mostly because they're ambiguous (without mental calculation) and more difficult to read (as a human).

src/cr/cube/measures/scale_means.py
scanny: The vague idea I have (vague still because I haven't gotten my head completely around the required complexity yet) is that we need some pre-processed working objects that encapsulate a lot of the differences that are now spread all through the code.

One example of this might be `_RawCube` vs. `_SquashedCube`. A `_RawCube` object would be formed directly from the underlying result dict and would know nothing about pruning, MR dimension squashing, or HS-insertion. It would be the source that `_SquashedCube` pulls value from. (Both of these would be internal implementation details, hence the leading underscore.) `_SquashedCube` would be constructed by a factory by passing parameters like `show_empty` and `show_hs` (or whatever that option is), and would then encapsulate providing a consistent "squashed" cube that was appropriately pruned, mr-squashed, etc. (including that its dimensions were squashed in a consistent way) such that clients of `_SquashedCube` (like CrunchCube and CubeSlice) didn't have to be carrying all those "flags" around with them and distributing the prune-vs-show-empty and mr-squashing logic through their code (which shows up as all these if-statements).

Anyway, a topic to discuss further.

PR #112, opened 2018-10-21
Refactor dimension

scanny:
This is a fairly thorough refactor of cr.cube.dimension.

Let's discuss when we speak next.

### DIFF
--- a/src/cr/cube/crunch_cube.py
+++ b/src/cr/cube/crunch_cube.py
@@ -149,7 +149,7 @@ class CrunchCube(object):
     @lazyproperty
     def ca_dim_ind(self):
         for (i, dim) in enumerate(self.dimensions):
-            if dim.type == 'categorical_array':
+            if dim.dimension_type == 'categorical_array':
                 return i
         else:
             return None
@@ -192,7 +192,7 @@ class CrunchCube(object):
 
     @lazyproperty
     def dim_types(self):
-        return [dim.type for dim in self.dimensions]
+        return [d.dimension_type for d in self.dimensions]
 
     @lazyproperty
     def dimensions(self):
@@ -300,7 +300,7 @@ class CrunchCube(object):
     @lazyproperty
     def is_univariate_ca(self):
         """Check if cube is a just the CA ("ca x cat" or "cat x ca" dims)"""
-        types = {d.type for d in self.dimensions}
+        types = {d.dimension_type for d in self.dimensions}
         ca_types = {'categorical_array', 'categorical'}
         return self.ndim == 2 and types == ca_types
 
@@ -461,7 +461,7 @@ class CrunchCube(object):
     def mr_dim_ind(self):
         indices = [
             i for i, dim in enumerate(self.dimensions)
-            if dim.type == 'multiple_response'
+            if dim.dimension_type == 'multiple_response'
         ]
         if indices:
             return indices[0] if len(indices) == 1 else tuple(indices)
@@ -479,7 +479,7 @@ class CrunchCube(object):
         mr_dimensions_indices = [
             i for (i, dim) in enumerate(self.all_dimensions)
             if (i + 1 < len(self.all_dimensions) and
-                dim.type == 'multiple_response')
+                dim.dimension_type == 'multiple_response')
         ]
 
         # For each MR and CA dimension, the 'selections' dimension
@@ -795,7 +795,7 @@ class CrunchCube(object):
     @lazyproperty
     def univariate_ca_main_axis(self):
         """For univariate CA, the main axis is the categorical axis"""
-        dim_types = [dim.type for dim in self.dimensions]
+        dim_types = [d.dimension_type for d in self.dimensions]
         return dim_types.index('categorical')
 
     def valid_indices_with_selections(self, include_missing=False):
@@ -880,7 +880,7 @@ class CrunchCube(object):
         # axis (that were provided by the user). But we don't need to update
         # the axis that are "behind" the current MR.
         for i, dim in enumerate(self.dimensions):
-            if dim.type == 'multiple_response':
+            if dim.dimension_type == 'multiple_response':
                 # This formula updates only the axis that come "after" the
                 # current MR (items) dimension.
                 new_axis[axis >= i] += 1
@@ -1322,10 +1322,11 @@ class CrunchCube(object):
             # Check if transformations can/need to be performed
             transform = (dim.has_transforms and
                          i - dim_offset in include_transforms_for_dims)
-            if dim.type == 'multiple_response':
+            if dim.dimension_type == 'multiple_response':
                 dim_offset += 1
             if (not transform or
-                    dim.type in ITEM_DIMENSION_TYPES or dim.is_selections):
+                    dim.dimension_type in ITEM_DIMENSION_TYPES or
+                    dim.is_selections):
                 continue
             # Perform transformations
             insertions = self._insertions(res, dim, i)
--- a/src/cr/cube/crunch_cube.py
+++ b/src/cr/cube/crunch_cube.py
@@ -801,7 +801,7 @@ class CrunchCube(object):
     def valid_indices_with_selections(self, include_missing=False):
         """Get all valid indices (including MR selections)."""
         return [
-            dim.valid_indices(include_missing)
+            dim.element_indices(include_missing)
             for dim in self.all_dimensions
         ]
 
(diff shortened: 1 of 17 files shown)


src/cr/cube/crunch_cube.py
```diff
@@ -149,7 +149,7 @@ def as_array(self, include_missing=False, weighted=True, adjusted=False,
     @lazyproperty
     def ca_dim_ind(self):
         for (i, dim) in enumerate(self.dimensions):
-            if dim.type == 'categorical_array':
+            if dim.dimension_type == 'categorical_array':
```
scanny: I renamed `Dimension.type` to `.dimension_type`. There are some comments in the commit going into the rationale. Short version is having a property or method named `.type` or `.id` is to be avoided. Among other benefits this lets the reader know what kind of type it is when context can't make that clear.

src/cr/cube/crunch_cube.py
```diff
@@ -795,13 +795,13 @@ def scale_means(self, hs_dims=None, prune=False):
...
         """For univariate CA, the main axis is the categorical axis"""
-        dim_types = [dim.type for dim in self.dimensions]
+        dim_types = [d.dimension_type for d in self.dimensions]
         return dim_types.index('categorical')
 
     def valid_indices_with_selections(self, include_missing=False):
         """Get all valid indices (including MR selections)."""
         return [
-            dim.valid_indices(include_missing)
+            dim.element_indices(include_missing)
```
scanny: I renamed `Dimension.valid_indices()` to `.element_indices()` since it can be used to get both all and only valid indices. This eliminates a jarring contradiction in the naming.

src/cr/cube/crunch_cube.py
```diff
@@ -1065,32 +1065,29 @@ def _inserted_dim_inds(self, transform_dims, axis):
...
+                    [slice(None) for _ in range(dimension_index)] +
+                    [np.array(addend_idxs)]
+                )
+                yield (
+                    insertion_idx,
+                    np.sum(result[addend_fancy_idx], axis=dimension_index)
+                )
 
-        return insertions
+        return [insertion for insertion in iter_insertions()]
```
scanny: I extracted the body of the code into the local method `iter_insertions()` to make this simpler to understand.

src/cr/cube/dimension.py
```diff
@@ -19,124 +21,125 @@ class Dimension(object):
     :attr:`.CrunchCube.dimensions`.
     """
 
-    def __init__(self, dim, selections=None):
-        self._dim = dim
-        self._type = self._get_type(dim, selections)
+    def __init__(self, dimension_dict, next_dimension_dict=None):
+        self._dimension_dict = dimension_dict
+        self._next_dimension_dict = next_dimension_dict
```
scanny: I renamed `dim` to the more descriptive `_dimension_dict`, especially since `dim` is so often used as a shorthand for a `Dimension` object. I renamed `selections` to the more precise `next_dimension_dict` for clarity.

src/cr/cube/dimension.py
```diff
@@ -19,124 +21,125 @@ class Dimension(object):
...
     def description(self):
-        """Description of a cube's dimension."""
-        refs = self._dim['references']
-        return refs.get('description')
+        """str description of this dimension."""
+        description = self._dimension_dict['references'].get('description')
+        return description if description else ''
+
+    @lazyproperty
+    def dimension_type(self):
```
scanny: This is the refactored form of `.type`, `._type`, `_get_type()`, and `._is_multiple_response()`. Calling `._get_type()` in the constructor represented "real work", which is why you had to mock it so often in unit tests even when it didn't have anything to do with that unit. The classmethods were unnecessary once `next_dimension_dict` was stored as an instance var and `.dimension_type` was made lazy. All this logic really belongs in a `Dimensions` collection object anyway, which we'll probably get to before too long.

src/cr/cube/dimension.py
```diff
@@ -19,124 +21,125 @@ class Dimension(object):
...
-        for i, el in enumerate(self._elements):
-            el['index'] = i
-            r[el['id']] = el
-        return r
+    @memoize
+    def elements(self, include_missing=False):
+        """_Elements object providing access to elements of this dimension."""
+        return (
+            self._all_elements if include_missing else self._valid_elements
+        )
```
scanny: One of the big items in this PR is extracting `_Elements` and `_Element` (and their object hierarchy siblings) as full-fledged objects. This removes all the widely distributed dict manipulation that is hard to follow and very error prone. Using dicts instead of real objects for jobs like these is an example of the "Primitive Obsession" anti-pattern, and this refactoring provides a good example of how to eliminate that and the benefits in clarity, conciseness, and testability that can bring.

src/cr/cube/dimension.py
```diff
@@ -156,287 +159,481 @@ def is_mr_selections(self, others):
...
+        return _AllElements(self._dimension_dict['type'])
+
+    def _iter_interleaved_items(self, elements):
+        """Generate element or subtotal items in interleaved order.
+
+        This ordering corresponds to how value "rows" (or columns) are to
+        appear after subtotals have been inserted at their anchor locations.
+        Where more than one subtotal is anchored to the same location, they
+        appear in their document order in the cube response. pairs for this
+        dimension.
```
scanny: This last sentence fragment is an editing error. I've fixed it in my copy and will push with any other changes before we proceed.

src/cr/cube/dimension.py
slobodan-ilic: ❤️

src/cr/cube/dimension.py
```diff
@@ -401,6 +305,208 @@ def _update_with_subtotals(self, labels_with_cat_ids):
...
 
+    @lazyproperty
+    def _valid_elements(self):
+        """_Elements object providing access to non-missing elements.
+
+        Any categories or subvariables representing missing data are excluded
+        from the collection; this sequence represents a subset of that
+        provided by `._all_elements`.
+        """
+        return self._all_elements.valid_elements
```
slobodan-ilic: From this expression it seems that having "elements" repeated two times is redundant. Maybe something like
```
self._elements.all
```
vs
```
self._elements.valid
```
would be more concise and/or readable.
But this is just a comment, not sure it's good/easy to change it like that.

src/cr/cube/dimension.py
```diff
@@ -19,124 +21,125 @@ class Dimension(object):
...
+        # ---object, which is where it will move to once that's implemented
+
+        def next_dim_is_mr_cat():
+            """True if subsequent dimension is an MR_CAT dimension."""
+            if not self._next_dimension_dict:
+                return False
+
+            categories = self._next_dimension_dict['type'].get('categories')
+            if not categories:
+                return False
```
slobodan-ilic: This line is untested. Is there an easy way of testing it?

src/cr/cube/dimension.py
```diff
@@ -156,287 +159,481 @@ def is_mr_selections(self, others):
...
+        if include_cat_ids:
+            element_ids = tuple(
+                None if item.is_insertion else item.element_id
+                for item in interleaved_items
+                if include_subtotals or not item.is_insertion
             )
-            for label in labels_with_cat_ids
-            if self._include_in_labels(label, valid_indices)
-        ]
+            return list(zip(labels, element_ids))
```
slobodan-ilic: This (under if) is untested

src/cr/cube/dimension.py
```diff
@@ -156,287 +159,481 @@ def is_mr_selections(self, others):
...
-            return value
+        """tuple storing actual sequence of element objects.
 
-        # For categorical array variables
-        name = value.get('references', {}).get('name')
-        if name:
-            return name
+        Must be implemented by each subclass.
+        """
+        raise NotImplementedError('must be implemented by each subclass')
```
slobodan-ilic: Also this, although it's understandable why :)

src/cr/cube/dimension.py
scanny: This whole method is reimplemented in the next PR (to come) that extracts dimension type discovery into the Dimensions collection. I propose we skip elaborating the test to exercise this line since it will only be cut out a commit or two later to be replaced by a more permanent test elsewhere.

scanny:
Ok, I think this is ready to go. I merged master to resolves a parallel development conflict.

We'll need make some changes in our exporter code to make this work over there. Once this is released I'll prepare a commit containing those changes along with the new version bump.

src/cr/cube/dimension.py
scanny: I think this is the right naming. This location is the only place we expect this expression to arise and I'm strongly inclined to keep things explicit. If we neglect to distinguish all from valid here it makes readability poor because you have to remember (or more likely go look) to determine what that simplified name means.

End users will have no direct access to the _AllElements object so they will not need to use the `_AllElements.valid_elements` property.

PR #113, opened 2018-10-29
Differentiated dimension-types

scanny:
This PR builds on the work done in the `refactor-dimension` branch.

At a high level, it extracts dimension type from a str object like 'multiple_response' to an enumeration object like `DIMENSION_TYPE.MULTIPLE_RESPONSE` (commonly `DT.MR` in the code). The enumeration has distinct types for `CA_CAT`, `MR_CAT`, (regular) `CAT`, and `LOGICAL`, all of which used to report `'categorical'` as their dimension type. This full differentiation allows several helper properties that were needed to distinguish those to go away, simplifying the interface and reducing the support burden as well as making the code more explicit and therefore easier to reason about.

In addition, all the dimension-type discovery logic is now localized in a single place. As a result, several bugs were revealed and fixed. Many other improvements were made along the way to docstrings, naming, and tests, and this refactoring paves the way for the next round. The API for `CrunchCube` was also narrowed, which improves the API documentation and makes the library easier to learn for newcomers.


src/cr/cube/dimension.py
```diff
@@ -6,10 +6,244 @@
...
+
+    @lazyproperty
+    def _dimensions(self):
+        """tuple of dimension objects in this collection.
+
+        This composed tuple is the source for the dimension objects in this
+        collection.
+        """
+        return tuple(
+            d for d in self._all_dimensions
```
slobodan-ilic: `d` seems a bit too short for me personally. I know that linter used to complain, but I didn't check it now. It's super minor, you don't have to fix it, but just wanted to point out that it _might_ need to be fixed if we firm our position on code formatting.

scanny: Well, my rationale here is that the scope is so very, very small (one line) and the contents of `._all_dimensions` is clear.

We could go with `dimension for dimension in self._all_dimensions`, but for whatever reason I find this expression clearer, like it would be in math "Let d represent a dimension". There's some tendency in functional programs to use very short names like these in a short scope (like 3 or 4 lines, no more), and I've been experimenting a little to see when I find they read better. I understand of course that readability is somewhat subjective at this level of detail.

I would suggest we let this one stand and see how we like it when returning back to it after a few weeks or whatever. It's difficult to distinguish 'different' from 'worse' at first, so if we age it a while and see what we think then we can make a change if we want.

PR #115, opened 2018-11-01
Full test coverage

scanny:
Bring test coverage to 100%.

Let's get this CR'ed and merged to master. No need for a new release but we should base all future branches on this.


tests/unit/test_util.py
```diff
@@ -0,0 +1,93 @@
+# encoding: utf-8
+
+"""Unit test suite for cr.cube.cube_slice module."""
```
slobodan-ilic: `cr.cube.cube_slice` -> `cr.cube.util`

PR #116, opened 2018-11-06
rfctr: move test fixtures to tests/fixtures

scanny:
* Move cube-response JSON from tests/integration/fixtures/cube to
  /tests/fixtures.

* Add fixture lazy-loading.


scanny:
@slobodan-ilic Can you CR this and merge at your earliest opportunity? It's pretty big (although simple) so it will be easier for both of us if we can base on this on master rather than have to do a merge/rebase onto it later.

tests/integration/test_multiple_response.py
```diff
@@ -355,11 +346,7 @@ def test_array_x_mr_by_row():
 
 
 def test_array_x_mr_by_cell():
-    cube = CrunchCube(ARRAY_X_MR)
```
slobodan-ilic: ❤️

PR #117, opened 2018-11-06
Extract measures

scanny:
This refactoring extracts a `_Measures` object and a hierarchy of measure objects that subclass `_BaseMeasure`. Specifically, these are `_MeanMeasure`, `_UnweightedCountMeasure`, and `_WeightedCountMeasure`.

These objects extract the measure-related logic out of CrunchCube as pure-functions (lazyproperties). The new objects have 100% unit test coverage and simplify CrunchCube.


src/cr/cube/crunch_cube.py
```diff
@@ -1177,6 +1134,32 @@ def _margin_pruned_indices(margin, inserted_ind, axis):
...
+        If the cube response includes a means measure, the return value is
+        means. Otherwise it is counts, with the choice between weighted or
+        unweighted determined by *weighted*.
+
+        Note that weighted counts are provided on an "as-available" basis.
+        When *weighted* is True and the cube response is not weighted,
+        unweighted counts are returned.
+        """
+        return (
+            self._measures.means if self._measures.means is not None else
```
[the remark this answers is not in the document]
slobodan-ilic: Maybe the explicit form would be more readable here:
```
measures = self._measures
if measures.means is not None:
    return measures.means
if weighted:
    return measures.weighted_counts
return measures.unweighted_counts
```
I believe it would also not hide potential misses by coverage (not sure about this though). I tried it locally, and it increases the unit test coverage from 65.94% to 66.10%

src/cr/cube/crunch_cube.py
```diff
@@ -940,6 +940,19 @@ def _calculate_std_res(self, counts, total, colsum, rowsum, slice_):
...
+
+        The return value is a _WeightedCountMeasure object if *weighted* is
+        True and the cube response is weighted. Otherwise it is an
+        _UnweightedCountMeasure object. Any means measure that may be present
+        is not considered. Contrast with `._measure()` below.
+        """
+        return (
+            self._measures.weighted_counts if weighted else
+            self._measures.unweighted_counts
+        )
```
slobodan-ilic: Writing it like this:
```
if weighted:
    return self._measures.weighted_counts
return self._measures.unweighted_counts
```
Increases the unit test coverage from 65.94 to 66.01 percent.

src/cr/cube/crunch_cube.py
```diff
@@ -869,26 +868,20 @@ def _apply_missings_and_insertions(self, res, include_transforms_for_dims,
...
                 include headings & subtotals (H&S) transformations.
-            margin (bool): Designates whether array is intented for margin
-                calculation. This essentially tells the CrunchCube to use
-                counts (and not means, or other measures) as the result.
         Returns
             res (ndarray): Tabular representation of crunch cube
         """
         return self._apply_missings_and_insertions(
-            self._raw_cube_array(weighted, margin),
+            self._measure(weighted).raw_cube_array,
```
slobodan-ilic: ❤️

tests/fixtures/scale_means/__init__.py
```diff
@@ -1,22 +0,0 @@
-import os
```
slobodan-ilic: I'm not sure that this particular commit corresponds to the message.

PR #118, opened 2018-11-07
Extract denominator

slobodan-ilic:
This PR extracts the duplicated code from `proportions` and `margin`, and puts it into a separate internal method. Furthermore, it introduces a clear distinction between the concepts of the denominator, versus that one of the margin. There's potentially more refactoring that can be done after this work.

~~*Note*: Needs rebase on #117 (only last 6 commits relevant).~~ Rebased


slobodan-ilic:
Closed in favor of #152

PR #121, opened 2018-11-12
Make margin explicit in slices 161883263

slobodan-ilic:


slobodan-ilic:
Updated, please re-check. It might not be perfect yet, but it's better thanks to your objections.

slobodan-ilic:
Hey @scanny , it's been a while :) Could you please approve this PR? I've answered all of your remarks. I need it to move forward on exporter work. I think this fixes the state of the slice in general, in a very good way. I plan to follow through with additional work...

PR #122, opened 2018-11-12
fix: add xfail for new broadcast Sentry error

scanny:
@slobodan-ilic this one is over to you now. The error message is similar to an earlier one, but I don't think they're closely related. This one happens on MR x MR x CAT zscores.

I've based this on the latest release (v1.8.2) so we can push it more quickly once we have a fix, but I can rebase it on master after you've worked it out if that makes most sense at the time.

----

* Add CubeSlice integration test that triggers "cannot broadcast..."
  error caught by Sentry.

* Silence W504 'line break after binary operator' flake8 error arising
  in latest version as those occur commonly in our code and we currently
  have no quick option to fix that.

* Remove slice reprs from CrunchCube repr as they are overwhelming and
  not useful when there is more than a couple small slices. Each slice
  can still be repr'ed individually on the command-line as desired.

* Rename MR_X_MR_X_CAT fixture to CAT_X_MR_X_MR_EMPTY_COLS to reflect
  its actual contents. Rename integration test that uses it similarly.

* Add new (actual) MR_X_MR_X_CAT fixture pulled from this error scenario
  and use it in new integration test that triggers this error.


src/cr/cube/cube_slice.py
```diff
@@ -309,6 +329,29 @@ def _apply_pruning_mask(self, res, prune):
...
+            res = (counts - expected) / np.sqrt(variance)
+        else:
+            expected_counts = expected_freq(counts)
+            residuals = counts - expected_counts
+            variance = (
+                np.outer(rowsum, colsum) *
+                np.outer(total - rowsum, total - colsum) / total ** 3
+            )
+            res = residuals / np.sqrt(variance)
+        return res
```
scanny: I think this would be better as:

```python
    def _calculate_std_res(self, counts, total, colsum, rowsum):
        """Return ndarray containing standard residuals.

        The shape of the return value is the same as that of *counts*.
        """
        has_mr_or_ca = set(self.dim_types) & DT.ARRAY_TYPES
        if has_mr_or_ca:
            if self.mr_dim_ind == 0:
                total = total[:, np.newaxis]
                rowsum = rowsum[:, np.newaxis]
            expected_counts = rowsum * colsum / total
            variance = (
                rowsum * colsum * (total - rowsum) * (total - colsum) /
                total ** 3
            )
            return (counts - expected_counts) / np.sqrt(variance)

        expected_counts = expected_freq(counts)
        residuals = counts - expected_counts
        variance = (
            np.outer(rowsum, colsum) *
            np.outer(total - rowsum, total - colsum) / total ** 3
        )
        return residuals / np.sqrt(variance)
```

and even better as:

```python
    def _calculate_std_res(self, counts, total, colsum, rowsum):
        """Return ndarray containing standard residuals.

        The shape of the return value is the same as that of *counts*.
        """
        if set(self.dim_types) & DT.ARRAY_TYPES:  # ---has-mr-or-ca---
            return self._array_type_std_res(counts, total, colsum, rowsum)
        return self._cat_type_std_res(counts, total, colsum, rowsum)
        # ---or maybe this is called _scalar_type_std_res()---

    def _array_type_std_res(self, counts, total, colsum, rowsum):
        """Return ndarray containing standard residuals for array values.

        The shape of the return value is the same as that of *counts*.
        Array variables require special processing because ...
        """
        if self.mr_dim_ind == 0:
            # --helpful comment explaining to me why this preliminary
            # --(inflation?) operation is required and what this special case
            # --is.
            total = total[:, np.newaxis]
            rowsum = rowsum[:, np.newaxis]
        expected_counts = rowsum * colsum / total
        variance = (
            rowsum * colsum * (total - rowsum) * (total - colsum) /
            total ** 3
        )
        return (counts - expected_counts) / np.sqrt(variance)

    def _scalar_type_std_res(self, counts, total, colsum, rowsum):
        """Return ndarray ..."""
        expected_counts = expected_freq(counts)
        residuals = counts - expected_counts
        variance = (
            np.outer(rowsum, colsum) *
            np.outer(total - rowsum, total - colsum) / total ** 3
        )
        return residuals / np.sqrt(variance)
```

Which is all staged up for later extraction into separate subclasses, like maybe ScalarSlice and ArraySlice. In any case, each method is now more compact and simpler to reason about, especially the `_scalar_type_std_res()` which is now straight-through (no if statements), and very importantly, they can now be thoroughly unit tested.

The first and the third are easily unit tested. The second one would require a parameterized fixture to handle the multiple cases, but is still well within range of unit testing.

I'm not sure if the right name is `_scalar_type_std_res` or perhaps `_cat_type_std_res`, but in either case it's an opportunity to articulate in the code a fundamental distinction in processing cases that I bet shows up elsewhere as well (and can then be signaled as similar by using the same prefix name array or scalar). These distinctions have a way of growing into subclasses that gather those differences in processing (polymorphishm).

Note that these methods need docstrings that clearly articulate their contract, the same as any other method. You don't need those now because you wrote them, but I need it to understand the behavior and you need them later to remember. Also they have a way of letting you know when you haven't got the contract right because it's too hard to explain in a concise docstring.

src/cr/cube/cube_slice.py
```diff
@@ -300,6 +301,29 @@ def table_name(self):
...
+        only.
+
+        :param weighted: Use weighted counts for zscores
+        :param prune: Prune based on unweighted counts
+        :param hs_dims: Include headers and subtotals (as NaN values)
+        :returns: 2 or 3 Dimensional ndarray, representing the p-values for each
+                  cell of the table-like representation of the crunch cube.
+        """
+        stats = self.zscore(weighted=weighted, prune=prune, hs_dims=hs_dims)
+        res = 2 * (1 - norm.cdf(np.abs(stats)))
```
scanny: This extensive use of the variable name 'res' seems smelly to me. What does it mean? I assume 'result', which is not descriptive and should at least be spelled out if we must use it. It's easy to confuse with 'residual' for one thing. I recommend using a descriptive term meaningful in the current context, like perhaps 'pvals' in this case. Then when you see the `.pvals()` method returning `pvals` it makes more sense and you can trace their development up through the rest of the method.

src/cr/cube/cube_slice.py
```diff
@@ -315,10 +339,31 @@ def zscore(self, weighted=True, prune=False, hs_dims=None):
...
+        if hs_dims:
+            res = self._intersperse_hs_in_std_res(hs_dims, res)
+
+        if prune:
+            arr = self.as_array(
+                prune=prune,
+                include_transforms_for_dims=hs_dims,
+            )
+            if isinstance(arr, np.ma.core.MaskedArray):
+                res = np.ma.masked_array(res, arr.mask)
```
scanny: I think this whole segment that occurs in multiple places is inlining a method like `._pruning_mask()` which can be memoized for performance. After extracting that (and using 'zscores' instead of 'res'), perhaps something like: `._apply_pruning_mask()` is possible such that these 7 lines becomes simply:

```python
if prune:
    zscores = self._apply_pruning_mask(zscores, hsdims)
```

src/cr/cube/cube_slice.py
```diff
@@ -315,10 +339,31 @@ def zscore(self, weighted=True, prune=False, hs_dims=None):
...
+        return res
+
+    def _intersperse_hs_in_std_res(self, hs_dims, res):
+        for dim, inds in enumerate(self.inserted_hs_indices()):
+            for i in inds:
+                if dim not in hs_dims:
+                    continue
+                res = np.insert(res, i, np.nan, axis=(dim - self.ndim))
+        return res
+
```
scanny: This new method looks out of position.

tests/integration/test_cube_slice.py
```diff
@@ -19,6 +25,37 @@ def it_provides_a_console_friendly_repr_for_a_slice(self):
...
 
+    def it_knows_zscore_for_MR_X_MR_X_CAT(self):
+        cube = CrunchCube(CR.MR_X_MR_X_CAT)
+        slice_ = cube.slices[0]
+
+        zscore = slice_.zscore()
+
+        np.testing.assert_array_almost_equal(
+            zscore,
+            # ---fix this expectation once we have plausible result---
```
scanny: This comment can come out now.

src/cr/cube/crunch_cube.py
```diff
@@ -732,17 +731,21 @@ def univariate_ca_main_axis(self):
...
         for slice_ in self.slices:
-            counts = slice_.as_array(weighted=weighted)
-            total = slice_.margin(weighted=weighted)
-            colsum = slice_.margin(axis=0, weighted=weighted)
-            rowsum = slice_.margin(axis=1, weighted=weighted)
-            std_res = self._calculate_std_res(
-                counts, total, colsum, rowsum, slice_,
-            )
-            res.append(std_res)
+            res.append(slice_.zscore(weighted, prune, hs_dims))
```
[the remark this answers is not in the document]
slobodan-ilic: Yeah, you probably went commit by commit, this is tackled in the last commit

src/cr/cube/cube_slice.py
slobodan-ilic: I agree. H&S pose many problems when passed around as parameters, but I hope they'll go away eventually.

Anyhow, I don't think this PR is the best place to address this particular issue, since it originally only _moves_ the functionality to a more appropriate place to live in. I'd ticket this and move along with the work done here.

src/cr/cube/cube_slice.py
slobodan-ilic: Done and unit tested (still only partially)

src/cr/cube/cube_slice.py
slobodan-ilic: Done

src/cr/cube/cube_slice.py
slobodan-ilic: Done

slobodan-ilic:
Addressed all of the comments. Hopefully it's in better shape now.

src/cr/cube/cube_slice.py
```diff
@@ -176,7 +176,7 @@ def index_table(self, axis=None, baseline=None, prune=False):
 
         indexes = proportions / baseline * 100
 
-        return self._apply_pruning_mask(indexes, prune)
+        return self._apply_pruning_mask(indexes) if prune else indexes
```
scanny: Very nice :) ❤️

src/cr/cube/cube_slice.py
```diff
@@ -339,63 +334,60 @@ def zscore(self, weighted=True, prune=False, hs_dims=None):
...
+            return self._apply_pruning_mask(zscore, hs_dims)
 
-    def _intersperse_hs_in_std_res(self, hs_dims, res):
-        for dim, inds in enumerate(self.inserted_hs_indices()):
-            for i in inds:
-                if dim not in hs_dims:
-                    continue
-                res = np.insert(res, i, np.nan, axis=(dim - self.ndim))
-        return res
+        return zscore
```
scanny: Very nice! I can totally see what this method does at a glance, which makes for quick navigation (and also ready unit testing). :heart:

PR #127, opened 2019-01-14
Pairwise comparisons

slobodan-ilic:


src/cr/cube/distributions/wishart.py
```diff
@@ -0,0 +1,135 @@
...
+
+try:
+    xrange
+except NameError:
+    xrange = range
+
+
+def wishartCDF(X, n_min, n_max):
+    w_p = np.zeros(X.shape)
+    for i, x in np.ndenumerate(X):
```
malecki: we could only evaluate the lower triangle and copy. this is still, as far as I can tell, sufficiently fast even doing more than double the work. consider benchmarking.

src/cr/cube/distributions/wishart.py
```diff
@@ -0,0 +1,135 @@
...
+        VectorXmp alpha_vec(n_min);
+        T alpha_last = 0.5*(n_max + n_min + 1);
+        for (int i = 0; i < n_min; i++) {
+            alpha_vec(i) = alpha + i + 1;
+            A(i, n_min) = pow(2, -alpha_last)*boost::math::gamma_p(alpha_vec(i), 0.5*xx)/boost::math::tgamma(alpha_last);
+            A(n_min, i) = - A(i, n_min);
+        }
+    }
+    """
+    size = n_min + (n_min % 2)
```
malecki: there are 4 commonly encountered special cases that I also don’t do here.

src/cr/cube/distributions/wishart.py
```diff
@@ -0,0 +1,135 @@
...
+    alpha_vec = alpha_ind + alpha + 1
+    gammainc_a = gammainc(alpha_vec, 0.5 * x)
+    if n_min != size:
+        alpha_ind = alpha_ind.astype(np.int)
+        other_ind = np.full(n_min, size - 1, dtype=np.int)
+        alpha_last = (n_max + n_min + 1) / 2
+        A[other_ind, alpha_ind] = (
+            np.float_power(2, -alpha_last) * gammainc_a / gamma(alpha_last)
+        )
+        # TODO check that this is not transposed
```
malecki: fixed (the values are put on the bottom, and their negative on the right side)

src/cr/cube/distributions/wishart.py
```diff
@@ -0,0 +1,135 @@
...
+    }
+    for (int i = 0; i < (2*n_min - 2); i++) {
+        q(i) = pow(0.5, 2*alpha + i + 2) * boost::math::tgamma(2*alpha+i+2)*boost::math::gamma_p(2*alpha + i + 2, xx);
+    }
+    """
+    p = gammainc_a
+    g = gamma(alpha_vec)
+    q_ind = np.arange(2 * n_min - 2)
+    q = (
+        np.float_power(0.5, 2 * alpha + q_ind + 2)
```
malecki: could use just `power` here but I figured explicit>implicit for cases where the base is int.

malecki:
```suggestion
        q_vec = 2 * alpha + q_ind + 2
        np.float_power(0.5, q_vec)
```

src/cr/cube/distributions/wishart.py
```diff
@@ -0,0 +1,135 @@
...
+        * gamma(2 * alpha + q_ind + 2)
```
malecki:
```suggestion
        * gamma(q_vec)
```

src/cr/cube/distributions/wishart.py
```diff
@@ -0,0 +1,135 @@
...
+        * gammainc(2 * alpha + q_ind + 2, x)
```
malecki:
```suggestion
        * gammainc(q_vec, x)
```

src/cr/cube/crunch_cube.py
```diff
@@ -713,6 +714,36 @@ def zscore(self, weighted=True, prune=False, hs_dims=None):
...
+        For 3D cubes, the measures of separate slices are stacked together
+        and returned as the result.
+
+        :param weighted: Use weighted counts for zscores
+        :param prune: Prune based on unweighted counts
+        :param hs_dims: Include headers and subtotals (as NaN values)
+        :returns zscore: ndarray representing zscore measurements
+        """
+        res = [s.pairwise_chisq(axis=0) for s in self.slices]
+        return res[0]
```
slobodan-ilic: I'm not sure that this is right. I'd expect us to stack the results of slices into a list or something... that is, I'd expect `return res` to be the correct value. And then it's the job of the exporter to discern between tabs, if there are multiples thereof...

src/cr/cube/crunch_cube.py
```diff
@@ -713,6 +714,36 @@ def zscore(self, weighted=True, prune=False, hs_dims=None):
...
+        """Return square ndarray of pairwise Chi-square along axis.
+
+        Square, symmetric matrix along *axis* of pairwise p-values for the
+        null hypothesis that col[i] = col[j] for each pair of columns.
+
+        :param axis: axis along which to perform comparison. Only columns (0)
+        implement currently.
+        :returns pvals: square, symmetric matrix of column-comparison p-values.
+        """
+        statistics = [s.pairwise_chisq(axis=0) for s in self.slices]
```
slobodan-ilic: Consequently, we wouldn't need this... We'd use instead: `self.pairwise_chisq`. Please lmk if I'm missing something here...

src/cr/cube/crunch_cube.py
```diff
@@ -713,6 +714,36 @@ def zscore(self, weighted=True, prune=False, hs_dims=None):
...
+        shapes = [s.get_shape() for s in self.slices]
+        return PairwisePvalues(statistics[0], shapes[0], axis=axis).values
```
slobodan-ilic: This class could take `self` as a parameter, and do all the work when needed, and not before...

src/cr/cube/cube_slice.py
```diff
@@ -453,6 +482,27 @@ def _scalar_type_std_res(self, counts, total, colsum, rowsum):
...
+            elements = counts.shape[1]
+            chisq = np.zeros([elements, elements])
+            for i in xrange(1, elements):
+                for j in xrange(0, elements - 1):
+                    chisq[i, j] = chisq[j, i] = np.sum(
+                        np.square(proportions[:, i] - proportions[:, j]) / wts
+                    ) / (1 / margin[i] + 1 / margin[j])
+
+            return chisq
+
```
slobodan-ilic: There should be an explicit return path, or a raise, which should in turn be covered by a unit test :)

src/cr/cube/distributions/wishart.py
```diff
@@ -0,0 +1,135 @@
...
+    for i, x in np.ndenumerate(X):
+        w_p[i] = _wishart_pfaffian(x, n_min, n_max)
+    return _normalizing_const(n_min, n_max) * w_p
+
+
+def _wishart_pfaffian(x, n_min, n_max):
+    """Value providing the CDF of the largest eigenvalue of a Wishart(n, p)
+    evaluated at x.
+
+    if (n_min != n_mat) {
```
slobodan-ilic: Is this supposed to live in here?

src/cr/cube/distributions/wishart.py
```diff
@@ -0,0 +1,135 @@
...
+        other_ind = np.full(n_min, size - 1, dtype=np.int)
+        alpha_last = (n_max + n_min + 1) / 2
+        A[other_ind, alpha_ind] = (
+            np.float_power(2, -alpha_last) * gammainc_a / gamma(alpha_last)
+        )
+        # TODO check that this is not transposed
+        A[alpha_ind, other_ind] = -A[other_ind, alpha_ind]
+    """
+
+    for (int i = 0; i < n_min; i++) {
```
slobodan-ilic: Same here, do we need this commented out code?

src/cr/cube/distributions/wishart.py
```diff
@@ -0,0 +1,135 @@
...
+        b = 0.5 * p[i] * p[i]
+        for j in xrange(i, n_min - 1):
+            b -= q[i + j] / (g[i] * g[j + 1])
+            A[j + 1, i] = p[i] * p[j + 1] - 2 * b
+            A[i, j + 1] = -A[j + 1, i]
+    return np.sqrt(det(A))
+
+
+def _normalizing_const(n_min, n_max):
+    """int n_mat = n_min + (n_min % 2);
```
slobodan-ilic: Should this go out?

src/cr/cube/crunch_cube.py
malecki: yeah this was just the expedient 'give me some numbers' assuming only one slice.

src/cr/cube/crunch_cube.py
malecki: agree

src/cr/cube/cube_slice.py
malecki: fine with the same `NotImplementedError`, or we can do the index-twiddling that the R impl does and test with the transpose.

src/cr/cube/distributions/wishart.py
malecki: 🚮

src/cr/cube/distributions/wishart.py
malecki: 🚮

src/cr/cube/distributions/wishart.py
malecki: 🚮

src/cr/cube/cube_slice.py
malecki: Is that the source of my minuscule coverage drop? oops

src/cr/cube/cube_slice.py
```diff
@@ -453,6 +482,29 @@ def _scalar_type_std_res(self, counts, total, colsum, rowsum):
...
         return residuals / np.sqrt(variance)
 
+    def _categorical_pairwise_chisq(self, counts, axis=0):
+        """Return ndarray containing pairwise comparisons along axis
+
+        Returns a square, symmetric matrix of test statistics for the null
+        hypothesis that each vector along *axis* is equal to each other.
+        """
+        if axis != 0:
+            raise NotImplementedError("Pairwise comparison only implemented for colums")
```
malecki: there I went with this for now

src/cr/cube/distributions/wishart.py
```diff
@@ -0,0 +1,89 @@
...
+    q = np.float_power(0.5, q_vec) * gamma(q_vec) * gammainc(q_vec, x)
+
+    # TODO consider index tricks instead of iteration here
+    for i in xrange(n_min):
+        b = 0.5 * p[i] * p[i]
+        for j in xrange(i, n_min - 1):
+            b -= q[i + j] / (g[i] * g[j + 1])
+            A[j + 1, i] = p[i] * p[j + 1] - 2 * b
+            A[i, j + 1] = -A[j + 1, i]
+    return np.sqrt(det(A))
```
scanny: To me, this thing screams out for implementation as a value object, which would encapsulate all these closely-related functions into a coherent class that was naturally understood as a single piece. It also would allow you to make most of these functions lazyproperties and would give you a handy place to extract a few more methods and/or properties. This function in particular looks too big to me. I can't really understand it because the notation is so terse and is not introduced in comments. But just given its size I'm inclined to think this function is doing well more than one thing. Perhaps it's implementation could become just `return np.sqrt(det(self._A_vector))` or something and then the implementation of `._A_vector` become a few lines that suitably arranged the interaction of further extracted small units that were each independently testable. If those units are given good names, then suddenly the complex calculation becomes much more accessible to the uninitiated.

src/cr/cube/distributions/wishart.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+def _normalizing_const(n_min, n_max):
```
scanny: Every module, class, method, property and function requires a docstring consistent with PEP 257. This is part of your bare minimum obligation to a maintainer (which of course is most likely yourself).

src/cr/cube/measures/pairwise_pvalues.py
```diff
@@ -0,0 +1,74 @@
...
+    def n_max(self):
+        """zero based int representing the bigger of the two cube's dimension."""
+        return max(self.slice.get_shape()) - 1
+
+    @lazyproperty
+    def n_min(self):
+        """zero based int representing the smaller of the two cube's dimension."""
+        return min(self.slice.get_shape()) - 1
+
+    def pairwise_chisq(self, axis=0, weighted=True):
```
scanny: Overall I like this class, although I wonder why `axis` and `weighted` can't be made construction-time arguments and make the entire interface lazyproperties. This is the characteristic of a value object; you create it and then you can read its values (no methods).

This decision is made based on expected usage, but a value object is inherently simpler to reason about and easier to understand. I'm strongly inclined to use them whenever possible and I generally find it possible. In the end the question is what the calling protocol will need to look like in the calling code, which is this class's reason for being.

src/cr/cube/measures/pairwise_pvalues.py
```diff
@@ -0,0 +1,74 @@
...
+except NameError:  # pragma: no cover
+    xrange = range
+
+
+class PairwisePvalues(object):
+    """Value object providing matrix of pairwise-comparison P-values"""
+
+    def __init__(self, slice_, axis=0):
+        self.slice = slice_
+        self.axis = axis
```
scanny: ALL instance variables should be private, *especially* in a value object. If you leave them public, that encourages a caller/user to access them and also allows them to mutate them, which breaks the value-integrity of the object. I have not found a good reason to ever have a public instance attribute.

While you're at it, only *actually intended interface properties* should be public (lack a leading underscore). If any of the lazyproperties below are just for internal use, like serve interim calculation purposes, make them private too. Private until proven public is a good policy in my experience.

src/cr/cube/measures/pairwise_pvalues.py
```diff
@@ -0,0 +1,93 @@
...
+        return min(self._slice.get_shape()) - 1
+
+    @lazyproperty
+    def _numel(self):
+        """int number of elements of the dimension opposite to axis."""
+        return self._slice.get_shape()[1 - self._axis]
+
+    @lazyproperty
+    def _wts(self):
+        # TODO: @mike - come up with a better docstring and property name
```
slobodan-ilic: @malecki This property needs a better name and docstring (i.e. this is just a placeholder that I came up with based on the original name of the variable from within the function).

src/cr/cube/measures/pairwise_pvalues.py
```diff
@@ -0,0 +1,93 @@
...
+        expected counts. It's only applicable to a 2D contingency tables.
+        """
+        return self._categorical_pairwise_chisq
+
+    @lazyproperty
+    def _proportions(self):
+        return self._slice.proportions(axis=self._axis)
+
+    @lazyproperty
+    def _n_max(self):
```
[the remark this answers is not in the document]
malecki: Gonna resist this one: I already changed the paper’s $ n_{mat} $ to the more obvious `size` but the n_min and n_max are used in nearly every part and breaking that connection to Chiani (2014) makes it harder to read not easier.

PR #133, opened 2019-01-30
Handle H&S when calculating pair-wise comparisons

slobodan-ilic:


src/cr/cube/cube_slice.py
```diff
@@ -328,7 +328,7 @@ def pairwise_pvals(self, axis=0):
         """
         if axis != 0:
             raise NotImplementedError("Pairwise comparison only implemented for colums")
-        return PairwisePvalues(self, axis=axis).values
+        return PairwisePvalues(self, axis=axis).values_with_hs
```
malecki: so I’ve never liked `hs` — why not more generic `insertions`? For instance, instead of subtotals we already have talked about a 'difference' insertion (cat A minus B), which is neither an `h` nor an `s`.

PR #140, opened 2019-03-11
Implement pairwise for MR x CAT

slobodan-ilic:
* Start with integration tests
* Modify pairwise functionality for MR support

As for the introduction of the parameter `include_mr_cat`, I'm beginning to think that i's actually the right thing to do. There are at least 2 situations that we might want the `MR_CAT` dimension included in the result (margin and percentages in this case, but I can anticipate such situations for counts, etc., since the whaam can do it...).

This is not to say that we don't need better separation of concerns in the cube library itself, especially between insertions, pruning, and calculations of actual numbers.


src/cr/cube/measures/pairwise_pvalues.py
```diff
@@ -18,61 +18,67 @@
...
-            for j in xrange(0, self._numel - 1):
-                chisq[i, j] = chisq[j, i] = np.sum(
-                    np.square(self._proportions[:, i] - self._proportions[:, j])
-                    / self._observed
-                ) / (1 / self._margin[i] + 1 / self._margin[j])
+    def _intersperse_insertions_rows_and_columns(self, pairwise_pvals):
+        """Return pvals matrix with inserted NaN rows and columns, as numpy.ndarray.
 
-        return chisq
+        Insertions (Headers and Subtotals) create offset in calculated pvals, and
```
scanny: Just a little technical writing tip, always strongly prefer an expression based on the singular form. A statement involving a plural is harder to reason about and is often ambiguous. So in this case, it would start "An insertion (a header or subtotal) creates an offset in the calculated pvals ...", or perhaps "Each insertion (header or subtotal) creates an offset ...".

Only when referring to a characteristic inherent in the _collection as a whole_ would you want to choose plural expression, and even then, I usually find it avoids ambiguity to say something like "The insertions collection ..." as opposed to "Insertions" (so still a singular expression, again avoiding the plural).

src/cr/cube/measures/pairwise_pvalues.py
```diff
@@ -18,61 +18,67 @@
...
+        these need to be taken into account, when converting them to columnar letters
```
scanny: no comma required between account and when. Also I think you mean "column letter" rather than "columnar letters".

src/cr/cube/measures/pairwise_pvalues.py
```diff
@@ -18,61 +18,67 @@
...
-        Zscore is a measure of statistical significance of observed vs.
-        expected counts. It's only applicable to a 2D contingency tables.
-        """
-        return self._categorical_pairwise_chisq
-
-    @lazyproperty
-    def _proportions(self):
-        """Slice proportions for *axis* as numpy.ndarray."""
-        return self._slice.proportions(axis=self._axis)
+    def _numel(self):
```
scanny: I would strongly prefer `._element_count` here. What is a numel? Sounds like it could be something I don't know about (but isn't). This name will be used outside this property; best to make it communicate clearly without having to look down here to understand what it means.

src/cr/cube/measures/pairwise_pvalues.py
```diff
@@ -85,9 +91,20 @@ def _n_min(self):
         return min(self._slice.get_shape()) - 1
 
     @lazyproperty
-    def _numel(self):
-        """Number of elements of the dimension opposite to axis, as int."""
-        return self._slice.get_shape()[1 - self._axis]
+    def _off_margin(self):
```
scanny: docstrings on all properties and methods please. I have no idea what this does or why. Don't make the (common) mistake of thinking that what is perfectly clear to you in this moment will be clear to others (or yourself next month :).

I'm not placing this same comment on every method without a docstring, but please look them all over and add one wherever missing, in this PR at least, but any others you come across along the way as well.

tests/fixtures/nordic-x-shapes.json
```diff
@@ -0,0 +1,586 @@
+{
```
scanny: Better to name fixtures by the structure they have (like MR x CAT or whatever) and any other relevant characteristics. "nordic" means nothing to me and does not help me understand what is distinctive about this test case. Also, when named by structure, it helps locate existing fixtures that can be reused for other tests.

tests/integration/test_pairwise.py
```diff
@@ -18,18 +18,132 @@
 class TestStandardizedResiduals(TestCase):
     """Test cr.cube implementation of column family pairwise comparisons"""
 
+    def test_mr_x_cat_residuals(self):
+        chisq_rows = [
```
scanny: large expectations like this are better stored as JSON and loaded rather than inlining like this. It makes the test much harder to read when it's more than a screenful.

src/cr/cube/measures/pairwise_pvalues.py
```diff
@@ -18,61 +18,67 @@
...
 
-    @lazyproperty
-    def values(self):
-        """Square matrix of pairwise Chi-square along axis, as numpy.ndarray."""
+    @classmethod
+    def pvals(cls, slice_, axis=0, weighted=True):
+        return cls._factory(slice_, axis, weighted)._pvals
+
+    @staticmethod
+    def _calculate_chi_squared(numel, proportions, margin, observed):
```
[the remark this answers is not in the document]
slobodan-ilic: I was thinking that using the verb in imperative form would be better, since this is the method and not the property. But yeah, I can fix the naming, I agree that `calculate` is not very informative...

src/cr/cube/measures/pairwise_pvalues.py
slobodan-ilic: Makes perfect sense. Thanks for the tip.

src/cr/cube/measures/pairwise_pvalues.py
slobodan-ilic: Will do. This is a remnant of the initial implementation...

src/cr/cube/measures/pairwise_pvalues.py
slobodan-ilic: It's perfectly clear to me :trollface:

PR #144, opened 2019-03-25
compare one or all; try to match what exporter wants to consume

malecki:


src/cr/cube/crunch_cube.py
```diff
@@ -737,6 +737,25 @@ def pairwise_pvals(self, axis=0):
         """
         return [slice_.pairwise_pvals(axis=axis) for slice_ in self.slices]
 
+    def compare_to_column(self, slice=0, column=0):
```
slobodan-ilic: I'm not sure we'll ever need this on the entire cube. It might be better to define this just on the slice level. Except maybe as an aggregation here, but definitely without the `slice` argument that you've provided here (unless I'm missing something obvious).

src/cr/cube/cube_slice.py
```diff
@@ -479,6 +480,59 @@ def zscore(self, weighted=True, prune=False, hs_dims=None):
...
+        """Test statistic and p-value for t-tests of proportions compared
+        to a column
+
+        Returns TtestResult, a pair of arrays of the same size as the current
+        slice: *statistic*, the t-statistic that proportions are different from
+        column *column*; and pvalue, the two-tailed p-value associated with that
+        test statistic (df = table N-2).
+        """
+
+        counts = self.as_array(weighted=weighted)
```
slobodan-ilic: Doesn't seem to be used

src/cr/cube/cube_slice.py
```diff
@@ -479,6 +480,59 @@ def zscore(self, weighted=True, prune=False, hs_dims=None):
...
+        props = self.proportions(axis=0)
+        diff = props[:, [column]] - props
+        margin = self.margin(axis=0, weighted=weighted)
+        var_props = props * (1.0 - props) / margin
+        se_diff = np.sqrt(var_props + var_props[:, [column]])
+        statistic = diff / se_diff
```
slobodan-ilic: This line concludes what's presented in Doug's PDF. After this line, the broadcasting and everything, feels a little crammed up. I'd like to see the explicit broadcasting gone, but I'm only reading this for the first time, so I can't really see _why_ it's necessary.

src/cr/cube/cube_slice.py
```diff
@@ -479,6 +480,59 @@ def zscore(self, weighted=True, prune=False, hs_dims=None):
...
+        df = np.broadcast_to(margin, self.get_shape()) + margin[column] - 2
+        pvals = 2 * (1 - t.cdf(abs(statistic), df=df))
+        TtestResult = namedtuple("TTestResult", ("statistic", "pvalue"))
```
slobodan-ilic: We probably want an explicit class for this, if we want to group these values together. Better testing, documentation, etc...

src/cr/cube/cube_slice.py
malecki: 🚮

src/cr/cube/cube_slice.py
malecki: and fixes two bugs in his code :P (the direction of the difference, and a very sneaky broadcasting error in R in `/ margin`)

src/cr/cube/cube_slice.py
malecki: halp :) I was cribbing from scipy, and we should coördinate with karol for what we want to send to whaam which needs both these pieces to do coloring.

src/cr/cube/cube_slice.py
```diff
@@ -479,6 +485,59 @@ def zscore(self, weighted=True, prune=False, hs_dims=None):
...
+        var_props = props * (1.0 - props) / margin
+        se_diff = np.sqrt(var_props + var_props[:, [column]])
+        statistic = diff / se_diff
+        unweighted_n = self.margin(axis=0, weighted=False)
+        df = unweighted_n + unweighted_n[column] - 2
+        pvals = 2 * (1 - t.cdf(abs(statistic), df=df))
+        TtestResult = namedtuple("TTestResult", ("statistic", "pvalue"))
+        return TtestResult(statistic, pvals)
+
+    def pairwise_indices(self, alpha=0.05, both_pairs=False):
```
slobodan-ilic: Unit tests please :)

PR #147, opened 2019-04-02
Decouple pruning and missingness

slobodan-ilic:
Break ugly method that does 2 things that shouldn't be together into two independent methods.

### DIFF
--- a/src/cr/cube/crunch_cube.py
+++ b/src/cr/cube/crunch_cube.py
@@ -325,8 +325,8 @@ class CrunchCube(object):
         # dividing. Those across dims which are summed across MUST NOT be
         # included, because they would change the result.
         hs_dims = self._hs_dims_for_den(include_transforms_for_dims, axis)
-        den = self._apply_missings_and_insertions(
-            table, hs_dims, include_missing=include_missing
+        den = self._apply_subtotals(
+            self._apply_missings(table, include_missing=include_missing), hs_dims
         )
 
         # Apply correct mask (based on the as_array shape)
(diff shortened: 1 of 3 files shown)


slobodan-ilic:
So, @scanny I've modified how we interpret valid indices. However, I've had to add another parameter to the `__init__` of some classes in `dimension.py`, to be able to fetch the index of an element in _only valid_ elements (and not all elements). Please see if you can suggest something better there.

However, I believe it's a step forward in the cube code, where some decoupling is done. In the future, I plan to do more of this, but from `cube_slice.py` mostly.

scanny:
I'm not sure I can propose a better way to do that; it's been a while since I've had my head into this code so I can't remember what gets done before other things. But it looks fine to me.

I think the big payoff comes when we get a `FrozenCube` (or whatever we name it), because that will allow the statefulness to come out and greatly reduce the amount of parameter passing that makes things so hard to reason about. The state is really insidious. Like you have to reason about "okay, is this table I have before or after applying missing? Does it have headings and subtotals yet? Oh shoot, I need the version before that. etc."

Once you have `CubeSlice` pretty well cleaned up I'd be inclined to create a `FrozenSlice` that just wraps the functionality of `CubeSlice`. That would be pretty fast because it's just mapping `CubeSlice` calls using instance variables. Initially that would allow us to change the interface of exporter without really changing any functionality. But then, you could incrementally reimplement the lazyproperties until the FrozenSlice stood on its own, and then we could use that base for new development after that.

I guess the short version of my take is you aren't really going to be able to get clean until you get rid of the state (like an ndarray in various stages of mutation). We can talk more about this when we get a chance to connect real-time.

PR #148, opened 2019-04-05
Add pruning to pairwise indices

slobodan-ilic:


slobodan-ilic:
Rendered irrelevant by #152

PR #152, opened 2019-04-18
Initial stab at transforms refactor

slobodan-ilic:
This is just the spike, so we can iterate back and forth, while doing this POC.

[This](https://www.pivotaltracker.com/story/show/165641665) is the relevant ticket.

[This](https://github.com/Crunch-io/zoom/pull/953) is the relevant exporter PR.


src/cr/cube/slices.py
```diff
@@ -0,0 +1,1140 @@
...
+        return dimensions
+
+    @lazyproperty
+    def _insertions(self):
+        return (
+            Insertions(self._dimensions, self._slice) if self._use_insertions else None
+        )
+
+    @lazyproperty
+    def _slice(self):
```
slobodan-ilic: @scanny This will essentially be the factory method, it can probably go to what will become the `_BaseSlice`, and generate the appropriate slice from there. It still uses the cube internals, but those can be extracted (namely the part that deals with missingness and measures), after we throw away everything else (margins, percentages, and most importantly pruning).

src/cr/cube/slices.py
```diff
@@ -0,0 +1,1140 @@
...
+        type_ = self._cube.dim_types[-2:]
+        if self._cube.ndim > 2 or self._ca_as_0th:
+            base_counts = base_counts[self._slice_idx]
+            counts = counts[self._slice_idx]
+            if self._cube.dim_types[0] == DT.MR:
+                base_counts = base_counts[0]
+                counts = counts[0]
+            elif self._ca_as_0th:
+                return _1DCatSlice(self._dimensions, counts, base_counts)
+        elif self._cube.ndim < 2:
```
slobodan-ilic: We have essentially 2 types of 1D stands, a Cat one and an MR one. CA as 0th is essentially a 1D Cat.

src/cr/cube/slices.py
```diff
@@ -0,0 +1,1140 @@
...
+
+    @lazyproperty
+    def zscore(self):
+        # return self._calculate_std_res(
+        return self._scalar_type_std_res(
+            self.counts, self.table_margin, self.column_margin, self.row_margin
+        )
+
+    @staticmethod
+    def _scalar_type_std_res(counts, total, colsum, rowsum):
```
slobodan-ilic: This will go to CAT x CAT slice, and get computed at construction time of either columns or rows (but will still be a lazy property). By assigning it to vectors, we'll be able to naturally intersperse NaNs where needed, when assembling rows (or cols).

src/cr/cube/slices.py
```diff
@@ -0,0 +1,1140 @@
...
+        columns = []
+        for i in range(len(self._slice.columns)):
+            columns.append(self._assembled_columns[i])
+            for insertion_column in self._assembled_insertion_columns:
+                if i == insertion_column.anchor:
+                    columns.append(insertion_column)
+        return tuple(columns)
+
+
+class _AssembledVector(object):
```
slobodan-ilic: Maybe the name should say `VectorWithOppositeDimensionInsertions`

src/cr/cube/slices.py
```diff
@@ -0,0 +1,1140 @@
...
+
+
+class Assembler(object):
+    def __init__(self, slice_, transforms):
+        self._slice = slice_
+        self._transforms = transforms
+
+    @lazyproperty
+    def slice(self):
+        """Apply all transforms sequentially."""
```
slobodan-ilic: This will need to apply the transforms sequentially. It might even turn out to be agnostic of the application order, but for now I'd keep it rigid.

src/cr/cube/slices.py
```diff
@@ -0,0 +1,1140 @@
...
+        return tuple(row for row in self._rows if row.anchor not in ("top", "bottom"))
+
+    @lazyproperty
+    def columns(self):
+        return tuple(
+            column for column in self._columns if column.anchor not in ("top", "bottom")
+        )
+
+    @lazyproperty
+    def intersections(self):
```
slobodan-ilic: This is not used anywhere, but I didn't delete it yet (probably won't need it).

src/cr/cube/slices.py
```diff
@@ -0,0 +1,1140 @@
...
+            return tuple()
+
+        return tuple(
+            InsertionRow(self._slice, subtotal)
+            for subtotal in self._row_dimension._subtotals
+        )
+
+    @lazyproperty
+    def _columns(self):
+        if len(self._dimensions) < 2 or self._column_dimension.dimension_type in (
```
slobodan-ilic: This can probably be taken care of with inheritance, but it's relatively simple (still) so I just kept it...

src/cr/cube/slices.py
```diff
@@ -0,0 +1,1140 @@
...
+    def row_anchors(self):
+        return tuple(row.anchor for row in self.rows)
+
+    @lazyproperty
+    def column_anchors(self):
+        return tuple(column.anchor for column in self.columns)
+
+    @lazyproperty
+    def _rows(self):
+        if self._row_dimension.dimension_type in (DT.MR, DT.CA):
```
slobodan-ilic: This can probably be thrown out as well with proper inheritance.

src/cr/cube/slices.py
```diff
@@ -0,0 +1,1140 @@
...
+        return np.sum(self._counts, axis=(1, 2))
+
+
+class _1DMrSlice(_MrXCatSlice):
+    @lazyproperty
+    def columns(self):
+        return tuple(
+            [
+                _MultipleResponseVector(
+                    self._counts.T, self._base_counts.T, "Summary", self.table_margin
```
slobodan-ilic: Label says "Summary", cause why not :) I'm open for better ideas. It's the name of the opposite dimension labels, which don't exist for 1D vector.

src/cr/cube/slices.py
```diff
@@ -0,0 +1,1140 @@
...
+        """Use bother selected and not-selected counts."""
+        return tuple(
+            _MultipleResponseVector(
+                counts, base_counts, element.label, self.table_margin
+            )
+            for counts, base_counts, element in self._column_generator
+        )
+
+    @lazyproperty
+    def table_margin(self):
```
slobodan-ilic: We need this at slice level, to be able to assign it to vectors, so each vector (either row or col) can know _all_ that it needs to know in order for _any_ of its measures. I'm planning for this to also happen with zscores and pvals.

src/cr/cube/slices.py
```diff
@@ -0,0 +1,1140 @@
...
+    @lazyproperty
+    def _row_elements(self):
+        return self._row_dimension.valid_elements
+
+    @lazyproperty
+    def _column_elements(self):
+        return self._column_dimension.valid_elements
+
+    @lazyproperty
+    def _row_generator(self):
```
slobodan-ilic: You will probably find a more elegant way to do this, something with `iter` or yield. I didn't find it necessary for the POC work done here.

src/cr/cube/slices.py
```diff
@@ -0,0 +1,1140 @@
...
+
+    @lazyproperty
+    def _column_generator(self):
+        return zip(self._counts.T, self._base_counts.T, self._column_elements)
+
+    @lazyproperty
+    def rows(self):
+        return tuple(
+            _CategoricalVector(counts, base_counts, element.label, self.table_margin)
+            for counts, base_counts, element in self._row_generator
```
slobodan-ilic: Just a quick description here: We're iterating across various rows and labels here. We need for each row the following:
- weighted count
- unweighted count
- label
- table margin
The first 3 are different for each row, while the table margin is the same. This will change for various MR cases, where the table margin changes.

We could also say that we _need_ the row margin, or the col margin for each vector here, but the fact is that the vector is smart enough (i.e. has access to enough information) that it can calculate those by itself, therefore those are implemented on the vector class.

src/cr/cube/slices.py
```diff
@@ -0,0 +1,1140 @@
...
+            _CategoricalVector(counts, base_counts, element.label, self.table_margin)
+            for counts, base_counts, element in self._column_generator
+        )
+
+    @lazyproperty
+    def table_margin(self):
+        return np.sum(self._counts)
+
+
+class _1DCatSlice(_CatXCatSlice):
```
slobodan-ilic: I don't know about you guys, but I find this rather elegant. The need for it arose when I was dealing with exporter tests. I was able to adopt the vector-based logic of the 2D slice to a 1D slice in a very short amount of time.

src/cr/cube/slices.py
```diff
@@ -0,0 +1,1140 @@
...
+        return tuple(
+            [
+                _MultipleResponseVector(
+                    self._counts.T, self._base_counts.T, "Summary", self.table_margin
+                )
+            ]
+        )
+
+    @lazyproperty
+    def table_margin(self):
```
slobodan-ilic: Table margin is different for 1D and 2D (it's debatable whether one exists for a 1D case). However this was the quickest way to make things still work for all the wired exporter tests.

src/cr/cube/slices.py
```diff
@@ -10,6 +10,200 @@
...
+        self._ca_as_0th = ca_as_0th
+
+    # interface ----------------------------------------------------------------------
+
+    @lazyproperty
+    def base_counts(self):
+        return self._calculator.base_counts
+
+    @lazyproperty
+    def column_base(self):
```
scanny: One thing that jumped out at me after I ordered these alphabetically is the pattern between column, row, and table. Column and row both have base, labels, margin, percentages, and proportions. Table has most of those (no labels). For now it's just an interesting symmetry to me, but makes me suspect an extractable object somewhere. I'm not sure what would be extracted though since these are all one-liners :). Anyway, too early for me to say there's anything to do about it, but an interesting pattern :) I really like that symmetry in the interface, that will make it much simpler to learn, because once you have the row, column, and table concepts, and the base, labels, margin, percentages, and proportions concepts, you basically understand the whole thing. Eight concepts to an interface? That's as simple as they come :)

slobodan-ilic: Yes! I've noticed that as well! I'm not sure how to deal with it. A little background: It didn't originate like that at all (a vector was responsible for all, nothing was on the slice). But with the introduction of the `table_margin`, and the decision that it be passed to the vector at construction time, we ended up in these shenanigans. So there might be a couple of things to tackle this:

1. At vector construction time (either row or column) - just pass it the whole slice, and the correct indexing (for the counts and bases) structure. It would go something like this:
`_CategoricalVector(self._row_dimension, index, self) for index in self._row_indexing`, or the like (not sure if we need row dimension in there, but the vector will need to know (maybe) if it's a row or a col, again, maybe in order to calculate something.
2. We might limit the `table_percentages` or `zscores` to only rows or only columns, so we just propagate the symmetrical methods to them. In this class we'd differentiate between a row and a col, which we now don't.

But I do like the concept of a particular vector not having to care about whether it's a row or a col. I think it removes a lot of responsibility from it, and consequently from duplicating the logic for that responsibility in two places (rows and columns).

scanny: Yes, the bit about a vector not having to care whether it's a row or column sounds right to me. We could work at the next-higher abstraction level where it perhaps has an "opposing" dimension (row if its a col vector, col-vector if its rows). I especially think this makes sense because a "table" should be pivotable without any change necessary to the logic. An interface property like `.row_percentages` would still be on the slice (because that's how the user thinks about it), but it's implementation might be something like `return self._row_vectors.percentages` or something.

I think the take-away for me here is that rows and columns is a "user-interface" concept (the user here being the developer consuming the FrozenSlice interface properties) and should _not_ have technical/object coupling unless we discover it absolutely has to.

So for example, maybe there's a Vectors object and it is a collection of vectors and kind of also represents a dimension. Maybe it has a type, like MR or CAT, but definitely it at least contains vectors of a type, like MR-subvar vectors or CAT vectors or whatever. But mostly it's just the collection. Then the slice has a `._row_vectors` item that's a Vectors object and it has a `._column_vectors` item that is a Vectors object. Maybe some vectors objects also get their "opposing" vectors object (which makes for an interesting chicken-and-egg construction problem, but deferring that discussion for now). Then MR x CAT uses the same vectors as CAT x MR, just in a different position.

Anyway, just food for thought at this point. I'd say if we're within range of making it work, that's the first objective, because refactoring based on a fully working set of integration tests is a much more orderly environment and allows incremental refinement.

scanny: Another possibility for situations where a Vectors object needs its opposing vectors for a calculation is to make that a method that takes the opposing vectors object. So one example, maybe not a good one, might be `slice.table_percentage` which could be implemented something like `self._row_vectors.table_percentage(self._column_vectors)`.

Because `Slice.table_percentage` is a lazyproperty, it would never calculate more than once. And if table-percentages are never needed locally, like as a basis for some other Vectors calculation, then the method would be called at most once for a Slice object.

Anyway, something to consider. Methods are not necessarily evil, they just have a cost we like to avoid whenever we can :)

src/cr/cube/slices.py
scanny: Yeah, I'm not sure, but I really like that it works so far and that totally makes it what we should do now in my mind. The first objective has to be to get to a working version that passes all the tests. This is _not_ so we can declare victory and move on, but because refactoring from there is such a more predictable and low-risk proposition. Any changes made with a full working set of integration tests is easy to see any breakage and it makes incremental refinement productive and fun. And, you can deliver at any moment if that turns out to be necessary, the critical thing being a stable external interface.

On the nature of a 1D cube result, I'm not convinced that thinking of it as a variant of a 2D matrix is the right long-term design. The thing is that a 1D item (we'll say "strand" for now) _has no "body" columns_. In particular, it doesn't "just have 1 column". It _also_ doesn't have any _row_values_ of the type that appear in a 2D slice. It does have a dimension, type, and counts, but I'm inclined to think the "column" of values it has is really a _margin_. So like a 2D slice _without the body_. It can still have _summary-columns_, but by nature cannot have any body columns (or body rows of course, since those are two sides of the same coin).

A confirming view is that a 2D/slice represents a _crosstab_ and a 1D/strand _does not_.

So I'm inclined to think this means a _strand_ should be a class hierarchy distinct from _slice_ (although perhaps some shared base class is possible).

All of this can wait until the refactoring step, but just some food for thought.

src/cr/cube/slices.py
scanny: If this works, let's run with it and sort it out in refactoring. The general idea I have here I mentioned in another comment, which is to produce it at the slice level when we need it, perhaps by calling a method on row-vectors that passes col-vectors to it. I don't have my head in the code enough yet to say specifically, but the general principle here is that a calculation responsibility should be on the object that has the data to perform it. That avoids passing it around to other objects, which has a strong tendency to grow without bound.

Anyway, another item we can address in refactoring.

tests/integration/test_dimension.py
```diff
@@ -47,7 +58,206 @@ def type_fixture(self, request):
...
+    to FrozenSlice. Once that work is completed, this will become DescribeDimension and
+    the NewDimension object will be merged into the Dimension object.
+    """
+
+    def it_constructs_from_a_legacy_Dimension_object(self, legacy_dimension):
+        dimension = NewDimension(legacy_dimension, None)
+        assert isinstance(dimension, NewDimension)
+
+    def it_provides_access_to_all_elements_in_its_collection(self, legacy_dimension):
+        legacy_dimension = CrunchCube(CR.ECON_BLAME_WITH_HS).dimensions[0]
```
slobodan-ilic: No need to override fixture here

tests/integration/test_dimension.py
```diff
@@ -47,7 +58,206 @@ def type_fixture(self, request):
...
+
+    def but_it_uses_element_name_when_not_transformed(self, legacy_dimension):
+        dimension_transforms_dict = {}
+        dimension = NewDimension(legacy_dimension, dimension_transforms_dict)
+
+        name = dimension.name
+
+        assert name == "ShutdownBlame"
+
+    def it_knows_whether_it_should_be_pruned(self, legacy_dimension):
```
slobodan-ilic: Should we add `but_it_defaults_to_false_if_not_specified`?

scanny: We could, but I only added these tests to drive the development of NewDimension and defend it until we got it integrated integrated into a single new Dimension object. At that point, I think most of these tests would be replaced by unit tests, which would be much more thorough, like cover all possible execution paths and input combinations. So I didn't try to be exhaustive here.

tests/integration/test_dimension.py
scanny: Oh, right, good catch, must have been left from before I added the fixture, thanks :)

src/cr/cube/slices.py
```diff
@@ -266,7 +262,7 @@ def _slice(self):
 
     @lazyproperty
     def _transforms(self):
-        return Transforms(self._ordering, self._pruning, self._insertions)
+        return Transforms(self._slice, self._dimensions, self._pruning)
```
scanny: I'd say let the Transforms object work out the value of `._pruning` for itself, now that it has the `._dimensions` (which is the authority). So that would mean delete `._pruning` property here and remove pruning param from Transforms constructor. Then think of who accesses pruning on a Transforms object and see if they can't get it for themselves as they likely have the dimension there, especially when pruning is dimension-by-dimension rather than all-or-nothing.

Also, shouldn't `._ordering` property come out too since this was the only caller?

src/cr/cube/slices.py
```diff
@@ -282,6 +283,11 @@ def table_margin(self):
         return np.sum(self._base_counts)
 
 
+# Used to represent the non-existent dimension in case of 1D vectors (that need to be
+# accessed as slices, to support cr.exporter).
+_PlaceholderElement = namedtuple("_PlaceholderElement", "label, is_hidden")
```
scanny: Should this be `"label", "is_hidden"`? looks like you missed the interior set of double-quotes.

src/cr/cube/slices.py
```diff
@@ -1090,6 +1094,11 @@ def __init__(self, slice_, subtotal):
         self._slice = slice_
         self._subtotal = subtotal
 
+    @lazyproperty
+    def hidden(self):
```
scanny: Should be `.is_hidden` I think, to be consistent with existing properties using that name, like in `_NewElement` for example.

slobodan-ilic: I feel you... But I like the sound of it in a list comprehension, or a generator expression:
`[row for row in self._slice.rows if not row.hidden]`
in my head - it sounds more like English - "do this, if not hidden..." then "do this if not is hidden" (which sounds more like a Spanish guy trying to transliterate Spanish into English).

We can decide on what we want to use and stick with it through the `cr.cube`. This is just a _slight_ preference, not a strong one.

src/cr/cube/slices.py
slobodan-ilic: It's funny for me as well, but it works like this :|

slobodan-ilic: Example from [this](https://docs.python.org/2/library/collections.html#collections.namedtuple) page:

```
EmployeeRecord = namedtuple('EmployeeRecord', 'name, age, title, department, paygrade')
```

PR #160, opened 2019-05-17
Continue development of Frozen Cube

slobodan-ilic:
For starters - increase coverage...


scanny:
@slobodan this one needs to be released before fill will work on exporter PPTX. Let me know if you need me to add anything.

PR #164, opened 2019-05-22
Initial tests for scorecards

slobodan-ilic:
Add artificially created cube response JSON, and check it for column
percentages. The cube in the response should have the following
dimensions:

    MR_SUBVAR x MR_CAT x ITEMS

The first two are parts of a single multiple response variable, while
the third one represents the "fuzed" dimension (i.e. a couple of MR
variables _just_ stacked together).

The ITEMS dimension doesn't have a special representation in the
FrozenSlice world, and is just treated as a CAT, although it's never
used that way.


slobodan-ilic:
Closing as irrelevant

PR #165, opened 2019-05-22
rfctr: normalize and document

scanny:
Normalizing method order and documenting along the way of learning this new module.


slobodan-ilic:
Everything looks great. I'll merge and we can continue from master. We don't need a release just yet, probably.

PR #167, opened 2019-05-25
Extract frozen dimension 166252645

scanny:
Create a cr.cube.legacy_dimension module to support legacy CrunchCube and CubeSlice.

Clean out cr.cube.dimension to support only FrozenCube and FrozenSlice.


scanny:
@slobodan there's a little more to do on this, I'll pick it back up on Tuesday (Monday is a US holiday). It's stable and passes all the tests on every commit, as usual, so we could release it in a pinch, there's just more related clean up to follow when I get back to it. If needed we can release this and I'll start another branch, whatever you need just go ahead and let me know.

src/cr/cube/dimension.py
```diff
@@ -251,222 +251,30 @@ class Dimension(object):
...
+        return _AllElements(
+            self._dimension_dict["type"], self._dimension_transforms_dict
         )
-        self._dimension_transforms_dict = dimension_transforms_dict
 
-    @lazyproperty
-    def all_elements(self):
-        """_NewAllElements object providing cats or subvars of this dimension.
+    def apply_transforms(self, dimension_transforms):
+        """Return a new `Dimension` object with `dimension_transforms` applied.
```
scanny: So there was an interesting problem in this process, which was where to apply transforms. Transforms are known at Cube construction time, passed in as a parameter. The question is where to apply them to the dimensions for individual slices.

My first instinct was to let `AllDimensions` do the job; transforms are on dimensions and AllDimensions creates all the dimensions, so it makes sense it should parcel out the transforms on first construction. The problem that arose is that `AllDimensions` then needed to duplicate all the rules that determine which dimension later becomes the rows-dimension in a slice and which becomes the columns-dimension, if any.

That was more than I could stomach after I thought about it awhile, just too brittle. So the solution I came up with is that all dimensions are initially constructed without transforms (defaults to None/{}). This takes `AllDimensions` out of the transforms assignment role. Then, `FrozenSlice` gets the right dimensions it wants and calls this method on each to produce the same dimension, but with transforms applied. The returned dimension object is different, so still immutable, but the same in every way except this time it has the desired transforms.

Anyway, thought that was a little bit interesting so thought I would mention.

scanny:
@slobodan-ilic I made a lot changes over the weekend, as you can see. I'd like to get this on master at our earliest convenience because it could be tough to merge into a parallel branch.

Getting prune and hide unified as a single transformation step would be the culmination of this work, so if you can see how to do that it would be great, otherwise let's get it released and we can work out how to build that on top of these changes.

slobodan-ilic:
I'll create a unified hide in a new branch

PR #168, opened 2019-05-28
Unify hidden and pruned 166287756

slobodan-ilic:


src/cr/cube/matrix.py
```diff
@@ -80,24 +79,18 @@ class PrunedMatrix(_BaseTransformedMatrix):
...
+            PrunedVector(column, self._base_slice.rows, self._applied)
             for column in self._base_slice.columns
-            if not column.pruned
+            if not column.hidden
         )
 
     @lazyproperty
     def rows(self):
-        if not self._applied:
-            return self._base_slice.rows
```
scanny: I'm wondering why we can't also remove this `"prune==True"` (`self._applied`) test from `PrunedMatrix.table_base` and `PrunedMatrix.table_margin`? I'm thinking all knowledge of prune-or-not should live on the dimension and that we can remove `Transforms.prune`. Shouldn't those two produce the same result regardless of whether pruning is turned on? Or are we possibly still making some distinction between a pruned vector and a hidden one?

src/cr/cube/slices.py
```diff
@@ -167,7 +161,7 @@ def name(self):
     @lazyproperty
     def ndim(self):
         """int count of dimensions for this slice, unconditionally 2."""
-        return 2
+        return self._matrix.ndim
```
scanny: Why is this the right thing to do? What value other than 2 can it have and why? It breaks a fundamental invariant about a Slice object if this can vary, at least on the public interface.

src/cr/cube/slices.py
```diff
@@ -528,7 +495,9 @@ def name(self):
     @lazyproperty
     def ndim(self):
         """int count of dimensions for this strand, unconditionally 1."""
-        return 1
+
+        # The value of the `ndim` on the `self._stipe` is also unconditionally 1
+        return self._stripe.ndim
```
scanny: If this is always 1, it's better to say so here so a reader doesn't need to refer to a collaborator to get that news. If the stripe doesn't need `.ndim` for its own purposes, we should remove it there, but it's no violation of principle for us to make this a constant value and it improves readability.

src/cr/cube/slices.py
```diff
@@ -676,7 +645,9 @@ def means(self):
     @lazyproperty
     def ndim(self):
         """int count of dimensions, unconditionally 0 for a Nub."""
-        return 0
+
+        # The ndim on the underlying _scalar is also unconditionally 0
+        return self._scalar.ndim
```
scanny: same here. If we're not expecting variation, we should keep it a literal value here.

src/cr/cube/vector.py
```diff
@@ -239,10 +231,23 @@ def anchor(self):
         return self._base_vector.anchor
 
 
-class _BasePrunedOrHiddenVector(_BaseTransformationVector):
-    def __init__(self, base_vector, opposite_vectors):
+class PrunedVector(_BaseTransformationVector):
+    def __init__(self, base_vector, opposite_vectors, prune=True):
```
scanny: Default parameters are best avoided except in public (user) interface classes where they can improve ease-of-use (at the expense of code complexity). In particular, they should not be used just to avoid updating a set of unit tests.  This parameter is always provided on all existing calls. Let's leave it as `, prune` and ditch the `=True`.

src/cr/cube/vector.py
```diff
@@ -239,10 +231,23 @@ def anchor(self):
...
+    def _valid_elements_idxs(self):
+        """An 1D ndarray of int idxs of non-hidden values, suitable for indexing."""
+        return np.array(
+            [
+                index
+                for index, opposite_vector in enumerate(self._opposite_vectors)
+                if not opposite_vector.hidden
+            ],
+            dtype=int,
+        )
```
scanny: Helper/implementation methods should go at the bottom of the class, in alphabetical order among the other helper methods. This is part of the important agile principle of _Seiri_ https://en.wikipedia.org/wiki/5S_(methodology)#Sort_.28Seiri.29

I fixed this in a following commit.

src/cr/cube/matrix.py
```diff
@@ -139,6 +139,33 @@ def table_margin_unpruned(self):
         return self._base_matrix.table_margin
 
 
+class StripeWithHidden(object):
+    """Stripe with hidden rows removed.
+
+    A row can be hidden explicitly by the user, or it can be automatically hidden when
```
slobodan-ilic: I'd add some description here, that the `row` is actually _just_ a single element inside the row, in this particular case.

PR #170, opened 2019-06-10
Numeric means 166482293

scanny:
Mostly renaming to remove "Frozen" designation from key objects. Then:

* rename row_margin to rows_margin, same with column
* elaborate the ca-as-0th condition to not fail on 0D first cube that appears in numeric-means tabbook
* Make CubePartition the base class for partitions and give it `.dimension_types`, `.ndim` etc. based on a `._dimensions` property implemented by each subclass.


scanny:
Hi Slobodan, this still needs one thing to support the numeric-means tabbook:

* `CubePartition.has_scale_mean`, I think `NotImplemented(must be implemented by each subclass)` on `CubePartition` itself, and then implemented separately on `_Slice`, `_Strand`, and `_Nub`, such that it can be called on any partition without knowing its dimensionality and give the right answer.

Once this is done let me know and I'll fix up the workarounds I have in place; once all that is passing nicely you can cut a new release and we can move this through on the exporter side.

PR #171, opened 2019-06-19
Add cr.cube.cube.CubeSet

scanny:
Add `cr.cube.cube.CubeSet` which takes a full tabbook response set (multiple cube-responses) and provides table-wise partition sets and other appropriate properties characterizing the overall response-set.


scanny:
Hi Slobodan, can you CR this one and push a new release if it's okay? If you'll let me know the new release number I'll update the pin on the exporter side and then that branch can go on dev.

PR #177, opened 2019-09-17
Increase test coverage

slobodan-ilic:


tests/unit/test_index_table.py
```diff
@@ -27,20 +28,17 @@ def test_index_performs_correct_division(index_fixture):
...
         ),
     ]
 )
 def index_fixture(request):
     proportions, axis, base, expected = request.param
-    cc = Mock()
+    cc = Mock(dim_types=(DT.CAT, DT.CAT))
     cc.ndim = 2
     cc.mr_dim_ind = None
     cc.proportions.return_value = np.array(proportions)
```
scanny: A couple best-practices I've adopted for unit test structuring you may want to consider:

```python
    def test_index_performs_correct_division(index_fixture, cube_):
        proportions, axis, base, expected = index_fixture
        # ---avoid all use of non-autospec-ed mocks. Use utility like instance_mock() that
        #    automatically produces autospec-ed mock and takes care of tear-down---
        cube_.dim_types=(DT.CAT, DT.CAT))
        cube_.ndim = 2
        cube_.mr_dim_ind = None
        cube_.proportions.return_value = np.array(proportions)
        cube_.as_array.return_value = np.array(proportions)
        # ---construction of the object under test is last line of fixture area---
        # ---object under test is not created in fixture unless absolutely unavoidable;
        #    remote construction obscures important aspect of fixture---
        # ---test object is spelled out so next guy doesn't have to puzzle out abbreviation
        cube_slice = CubeSlice(cube, 0)

        # ---tests always have fixturing, call, and asserts; fixturing is contiguous (no empty lines),
        #    then blank line separates call, then blank line separates asserts. When you're scanning
        #    tests to find the one that exercises the part you're interested in, you're looking for
        #    the call (.index_table() in this case). Making it stand out on a single line by itself
        #    makes that common operation fast.---
        actual = cube_slice.index_table(axis, base)

        np.testing.assert_almost_equal(actual, expected)


    @pytest.fixture
    def cube_(self, request):
        return instance_mock(CrunchCube)


    @pytest.fixture(
        params=[
            (
                [[0.714285714286, 0.285714285714], [0.625, 0.375]],
                1,
                [0.6, 0.4],
                [[119.047619047619, 71.4285714285714], [104.16666666666667, 93.75]],
            ),
            (
                [[0.5, 0.4], [0.5, 0.6]],
                0,
                [0.6, 0.4],
                [[83.3333333333333, 66.6666666666667], [125, 150]],
            ),
        ]
    )
    def index_fixture(request):
        proportions, axis, base, expected = request.param
        # ---fixture only returns case data; combining that data into test-fixturing happens
        #    up in the test itself to give better context when reading test---
        return proportions, axis, np.array(base), expected
```

The big thing is to avoid "bare" mocks (any time you see `Mock()` or `MagicMock()`). These are dangerous because they don't protect against interface changes so a test using those can pass when the code fails.

PR #180, opened 2019-10-01
Coverage attempt

ernestoarbitrio:
- New tests for CubeSet
- Fixed some typos


src/cr/cube/dimension.py
```diff
@@ -748,7 +748,7 @@ def __init__(self, element_transforms_dict, prune):
         self._prune = prune
 
     @lazyproperty
-    def fill(self):
+    def fill(self):  # pragma: no cover
```
scanny: I don't see the justification for a no-cover pragma here. This method does real work and deserves a unit test.

tests/integration/test_dimension.py
```diff
@@ -330,3 +331,12 @@ def test_numeric_values(self):
         numeric_values = dimension.numeric_values
 
         assert numeric_values == (1, 2, np.nan, np.nan)
+
+    def it_requires_each_subclass_to_implement__dimensions(self):
```
scanny: this should be handled with a no-cover pragma rather than a test.

tests/unit/test_cube.py
```diff
@@ -0,0 +1,21 @@
...
+"""Unit test suite for `cr.cube.cube` module."""
+
+from __future__ import absolute_import, division, print_function, unicode_literals
+
+import pytest
+
+from cr.cube.cube import Cube
+
+
+class DescribeBaseCrunchCube(object):
```
scanny: The name should mirror the class under test. So this one would be named `DescribeCube`. In general, a test class tests the methods of a single production class.

tests/unit/test_cube.py
```diff
@@ -0,0 +1,21 @@
...
+    def it_raises_a_type_error_if_not_serializable_object_provided(self):
```
scanny: In general, exception-path tests should be paired with the "happy-path" test. So here would go:

```python
def it_provides_access_to_the_cube_response_dict_to_help(self):
    # ---fixturing---

    cube_dict = cube._cube_dict

    # ---assertions about cube_dict---

def and_it_accepts_a_JSON_format_cube_response(self):
    # ---fixturing ---

    cube_dict = cube._cube_dict

    # ---assertions about cube_dict---

def but_it_raises_on_other_cube_response_types(self):
    # ---fixturing---

    with pytest.raises(TypeError) as e:
        cube._cube_dict

    # ---assertions about e---

File: tests/unit/test_cubeset.py [hunk 5]
```diff
@@ -0,0 +1,176 @@
...
+
+    def it_knows_whether_it_has_weighted_counts(
+        self, has_weighted_counts_fixture, _cubes_prop_, cube_
+    ):
+        first_cube_has_w_counts, expected_value = has_weighted_counts_fixture
+        cube_.is_weighted = first_cube_has_w_counts
+        _cubes_prop_.return_value = (cube_,)
+        cube_set = CubeSet(None, None, None, None)
+
+        is_weighted = cube_set.has_weighted_counts
```
scanny: In general, name the result variable exactly as the method or property under test (without any leading underscore). This reduces the concepts to be understood and improves speed-readability. It also makes writing faster because you don't have to reason about what a good name would be. So in this case:

```python
has_weighted_counts = cube_set.has_weighted_counts

assert has_weighted_counts == expected_value
```

tests/unit/test_cubeset.py
```diff
@@ -0,0 +1,176 @@
...
+
+        assert is_weighted == expected_value
+
+    def it_knows_its_name(self, has_name_fixture, _cubes_prop_, cube_):
```
scanny: I don't think many of these really benefit from a fixture because there is no real logic in the method. Returning "foobar" or `None` is not really different in this case because it just passes back what the name of the first cube is. So in these cases we can dispense with the fixture and just pick a representative case like "Beverage" and just test the one case.

Fixtures are great when you need to test multiple cases, but they do obscure the test somewhat because they are an "effect at a distance" and to really understand the test you need to see the fixture. So we should only use them when we need them, which is generally to exercise some branch (or implicit branch) in the code.

tests/unit/test_cubeset.py
```diff
@@ -0,0 +1,176 @@
...
+        cube_set = CubeSet(None, None, None, None)
+
+        cubeset_population_fraction = cube_set.population_fraction
+
+        np.testing.assert_almost_equal(cubeset_population_fraction, expected_value)
+
+    # fixture components ---------------------------------------------
+
+    @pytest.fixture(params=[(1.0, 1.0), (0.54, 0.54), (np.nan, np.nan)])
+    def population_fraction_fixture(self, request):
```
scanny: Arrange fixtures in alphabetical order. They are searched for frequently and ordering them makes that much faster along with other benefits.

tests/unit/test_cubeset.py
```diff
@@ -0,0 +1,176 @@
...
+        population_fraction, expected_value = population_fraction_fixture
+        cube_.population_fraction = population_fraction
+        _cubes_prop_.return_value = (cube_,)
+        cube_set = CubeSet(None, None, None, None)
+
+        cubeset_population_fraction = cube_set.population_fraction
+
+        np.testing.assert_almost_equal(cubeset_population_fraction, expected_value)
+
+    # fixture components ---------------------------------------------
```
scanny: Separate fixtures (used to provide varying parameters to a test method) from fixture components (small one-liner mocks used by one or more test methods. Fixtures first, fixture components beneath those, each with the appropriate header line like `# fixtures -----` etc.

tests/unit/test_cubeset.py
```diff
@@ -0,0 +1,176 @@
...
+        cubes_ = (cube_,) * ncubes
+        cubes_[0].dimension_types = (DT.CA_SUBVAR,) * ncubes
+        _cubes_prop_.return_value = cubes_
+        cube_set = CubeSet(cubes_, None, None, None)
+
+        isca_as_0th = cube_set.is_ca_as_0th
+
+        assert isca_as_0th == expected_value
+
+    def it_has_the_right_partition_set(
```
scanny: Naming should be from the perspective of the caller. In this case, this would be better "it_provides_access_to_the_partition_sets()". The return value is a sequence, so partitions is plural.

tests/unit/test_cubeset.py
```diff
@@ -0,0 +1,176 @@
...
+        self, partition_set_fixture, _cubes_prop_, cube_
+    ):
+        partition_set, expected_value = partition_set_fixture
+        cube_.partitions = partition_set
+        _cubes_prop_.return_value = (cube_,)
+        cube_set = CubeSet(None, None, None, None)
+
+        partitions_set = cube_set.partition_sets
```
scanny: `partition_sets`, not `partitions_set`. I think the item arriving from the fixture is `cube_partitions`, not `partition_set.`

src/cr/cube/dimension.py
```diff
@@ -937,7 +941,7 @@ def is_insertion(self):
 
         Unconditionally True for _Subtotal objects.
         """
-        return True
+        return True  # pragma: no cover
```
slobodan-ilic: This needs to be tested IMO. The rationale - if you put a `raise` here - it will destroy exporter tests.

A bit more context: you need to check if some rows or columns from the exporter will need to be *bold*. So you need to be able to tell if they're insertions or not. An integration test will preserve this context, but you don't have to do that, a simple unit test will prevent regression.

ernestoarbitrio: I changed an integration test here: 2ecc9795f88a2ef833b9948da5eadd720d2d2ef7

tests/integration/test_headers_and_subtotals.py
```diff
@@ -137,7 +137,9 @@ def test_1D_subtotals_rows_dimension_fills(self):
         assert strand.rows_dimension_fills == (None,) * 7
 
     def test_1D_subtotals_inserted_row_idxs(self):
-        strand = Cube(CR.ECON_BLAME_WITH_HS_MISSING).partitions[0]
+        cube = Cube(CR.ECON_BLAME_WITH_HS_MISSING)
```
slobodan-ilic: This is not necessary, but OK...

src/cr/cube/dimension.py
```diff
@@ -931,14 +935,6 @@ def addend_idxs(self):
             for addend_id in self.addend_ids
         )
 
-    @property
-    def is_insertion(self):
```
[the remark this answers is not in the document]
ernestoarbitrio: According to Slobodan we decided to remove this property because it's not use either in zoom/exporter or in cube

PR #183, opened 2019-10-22
Matrix refactoring

scanny:
Some refactorings to trim-down and clarify transform-application matrix objects.

### DIFF
--- a/src/cr/cube/matrix.py
+++ b/src/cr/cube/matrix.py
@@ -109,9 +109,7 @@ class _MatrixWithHidden(object):
     @lazyproperty
     def _base_matrix(self):
         """Apply all transforms sequentially."""
-        matrix = _OrderedMatrix(self._unordered_matrix)
-        matrix = _MatrixWithInsertions(matrix)
-        return matrix
+        return _MatrixWithInsertions(self._unordered_matrix)
 
     @lazyproperty
     def _rows_ind(self):
@@ -125,8 +123,8 @@ class _MatrixWithHidden(object):
 class _MatrixWithInsertions(object):
     """Represents slice with both normal and inserted bits."""
 
-    def __init__(self, base_matrix):
-        self._base_matrix = base_matrix
+    def __init__(self, unordered_matrix):
+        self._unordered_matrix = unordered_matrix
 
     @lazyproperty
     def columns(self):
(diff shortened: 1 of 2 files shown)


scanny:
Hi @slobodan-ilic I think this one is ready to go. It gets the same seven test failures on exporter that master does, so I expect those are unresolved integrations for the new pval work Ernesto has done.

Anyway, let me know what you think. I was able to trim out a lot of code and I think clarify things along the way, but I'll let you be the judge of that :)

src/cr/cube/matrix.py
```diff
@@ -28,187 +31,110 @@ def matrix(cls, cube, dimensions, slice_idx):
...
-            _VectorAfterHiding(column, self._base_matrix.rows)
-            for column in self._base_matrix.columns
-            if not column.hidden
+            _AssembledVector(row, opposing_insertions, 0 if idx < 0 else idx)
+            for _, idx, row in sorted(
+                itertools.chain(
+                    (row.ordering for row in self._inserted_rows),
+                    (row.ordering for row in self._base_rows),
+                )
+            )
```
scanny: @slobodan-ilic @ernestoarbitrio It smells to me like a bug that an assembled row that it an insertion row relies on its idx value to be 0. Why would it be that the the "intersection" pval would always be the first value of `.pvals` on the intersecting column? I suppose if that `.pvals` was always a sequence of exactly one value that would make sense of it, but it sounds a little odd to me and at the very least is obscure without a comment.

slobodan-ilic: This is probably indicative of needing two extractions (either as classes or as mechanisms within one class). The assembled row that is _not_ an insertion row itself _needs_ this index. The assembled row that _is_ an insertion row, doesn't need it (but currently has it defaulting to 0).

scanny: Well, the interesting thing is that at least one test fails on expectation doesn't match) if you replace `0 if idx < 0 else idx` with `idx`. `idx` always has a legitimate value, it's just end-based (like -2) and indexes into the insertions-collection if the vector is an insertion.

I took a quick look, and it appears this `idx` value is directly involved in the `pvals` calculation for an insertion, which just seems odd to me. I can't think of what calculation would _need_ a col_idx but work correctly with the same value for all insertions.

scanny: @ernestoarbitrio can you weigh in on this one? ^^^

PR #184, opened 2019-10-29
Compute t-tests for scale means [#169418662]

ernestoarbitrio:
Variance properties for scale means in _Strand and _Slice:

- `var_scale_mean` -> `_Strand`
- `var_scale_menas_row` -> `_Slice`
- `var_scale_menas_column` -> `_Slice`

New ttests for scale_means in pairwise significance:

- `t_stats_scale_means`
- `p_vals_scale_means`

### DIFF
--- a/src/cr/cube/measures/new_pairwise_significance.py
+++ b/src/cr/cube/measures/new_pairwise_significance.py
@@ -79,14 +79,19 @@ class _ColumnPairwiseSignificance:
 
     @lazyproperty
     def t_stats_scale_means(self):
-        # This property calculates the Two-Sample t-test using the formula:
-        # t = X1 - X2 / Sx1x2 * sqrt(1/n1 + 1/n2)
-        # where X1 and X2 are the scale mean value for the 2 sample we're
-        # comparing, n1 and n2 are the number of people from the 1st ans 2nd
-        # sample who provided a response to the survey, Sx1x2 is the standard
-        # deviation. In this case the standard deviation is:
-        # Sx1x2 = sqrt(((n1-1)*s2x1 + (n2-2)*s2x2)/(n1+n2+2)), where s2x1 and
-        # s2x2 are the is the standard deviation for sample 1 and 2.
+
+        """
+        This property calculates the Two-tailed t-test using the formula:
+        t = X1 - X2 / Sx1x2 * sqrt(1/n1 + 1/n2)
+        where X1 and X2 are the scale mean value for the 2 sample we're
+        comparing, n1 and n2 are the number of people from the 1st ans 2nd
+        sample who provided a response to the survey, Sx1x2 is the standard
+        deviation. In this case the standard deviation is:
+        Sx1x2 = sqrt(((n1-1)*s2x1 + (n2-2)*s2x2)/(n1+n2+2)), where s2x1 and
+        s2x2 are the is the standard deviation for sample 1 and 2.
+
+        :return: numpy 1D array of tstats
+        """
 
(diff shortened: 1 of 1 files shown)


src/cr/cube/cubepart.py
```diff
@@ -321,6 +337,26 @@ def scale_means_row(self):
...
+    def var_scale_means_row(self):
+        if np.all(np.isnan(self._rows_dimension_numeric)):
+            return None
+
+        not_a_nan_index = ~np.isnan(self._rows_dimension_numeric)
+        row_dim_numeric = self._rows_dimension_numeric[not_a_nan_index]
+        numerator = (
+            self.counts[not_a_nan_index, :]
+            * pow(
+                np.broadcast_to(
```
malecki: yes! this is one of my favorite lesser-known numpy functions.

tests/integration/test_pairwise_significance.py
```diff
@@ -8,12 +8,151 @@
...
 from cr.cube.cube import Cube
 
-from ..fixtures import CR
+from ..fixtures import CR, SM
 
 
 class TestStandardizedResiduals(TestCase):
     """Test cr.cube implementation of column family pairwise comparisons"""
 
+    def test_pairwise_measures_scale_means_nps_type(self):
```
ernestoarbitrio: @malecki this test is the most significant. In here we compare col 0 with others and col 1 with others using the fixture (facebook app vs ages)

src/cr/cube/cubepart.py
```diff
@@ -669,6 +703,18 @@ def unweighted_bases(self):
         """
         return tuple(np.broadcast_to(self.table_base, self._shape))
 
+    @lazyproperty
+    def var_scale_mean(self):
+        if np.all(np.isnan(self._numeric_values)):
+            return None
+
+        numeric_values_masked, is_a_number_mask = self._numeric_values_masked
+        counts = self._counts_as_array[is_a_number_mask]
```
slobodan-ilic: The counts are indexed here, while the numeric values are indexed in a different place. But still using the same mask. I'd prefer if the indexing was done in the same place, so it's more explicit and concise. I.e. maybe create a single property for `_numeric_mask`, and then use that from within this method, to get both the numeric values and counts:
```
num_vals = self._numeric_values[self._numeric_mask]
counts = self._counts_as_array[self._numeric_mask]
```
The way it is now, you kinda have to swap back and forth between two different functions, to figure out what's happening.

ernestoarbitrio: u right ... fixed in here 32cab83a2e48c62179fa4b764eebf72629a22750

src/cr/cube/measures/new_pairwise_significance.py
```diff
@@ -83,6 +123,17 @@ def pairwise_indices(self):
...
 
+    @lazyproperty
+    def pairwise_indices_scale_mean(self):
+        # Returns a list of n tuples where n is the number of the columns
+        # of the slice. Each tuple contains the indices of the significant
+        # column compared with the col ix where ix is the index of the tuple
+        # in the list. e.g [(), (), (1,), (0, 1, 2)]
+        significance = self.p_vals_scale_means < self._alpha
+        if self._only_larger:
+            significance = np.logical_and(self.t_stats_scale_means < 0, significance)
```
malecki:
```suggestion
            significance = np.logical_and(self.t_stats_scale_means > 0, significance)
```
I would have thought it was `t = observed - expected / se` in both cases, but this flip should do the trick I was trying to communicate.

PR #186, opened 2019-11-28
tstats correct for CATxMRxITSELF Cubes

ernestoarbitrio:
- New tstats measure in MRxMR Matrix
- new pairwise sig test for CATxMRxITSELF 5d cubes
- new factory condition for base matrix creation


src/cr/cube/matrix.py
```diff
@@ -667,6 +712,10 @@ def columns(self):
             for counts, base_counts, element, table_margin in self._column_generator
         )
 
+    @lazyproperty
+    def _mr_shadow_proportions(self):
```
scanny: add a docstring explaining what a shadow proportion is.

src/cr/cube/matrix.py
```diff
@@ -725,6 +774,10 @@ def _column_generator(self):
             self.table_margin.T,
         )
 
+    @lazyproperty
+    def _pairwise_overlap_total(self):
+        return np.sum(np.sum(self._counts, axis=1), axis=2)
```
scanny: add a docstring.

src/cr/cube/matrix.py
```diff
@@ -745,6 +798,53 @@ def _zscores(self):
             np.sum(self._counts, axis=3)[:, 0, :],
         )
 
+    @lazyproperty
+    def tstats_overlap(self):
+        """
```
scanny: Place this in alphabetical order within public instance methods.

tests/integration/test_cubepart.py
```diff
@@ -12,6 +12,18 @@
...
+        slice_ = cube.partitions[0]
+        assert slice_.table_name == cube.name
+        assert slice_.cube_is_mr_by_itself is True
+
+    def it_knows_cube_is_not_mr_by_itself(self):
+        cube = Cube(CR.AGE_FAVMR)
+        slice_ = cube.partitions[0]
+        assert slice_.table_name is None
+        assert slice_.cube_is_mr_by_itself is False
+
```
scanny: These two are the same test with a different parameter. Parameterize a single test with `@pytest.mark.parametrize`, grep for examples in exporter if you need them.

tests/integration/test_multiple_response.py
```diff
@@ -5,6 +5,13 @@
 from ..fixtures import CR
 
 
+# def test_dummy():
+#     slice_ = Cube(CR.GGG).partitions[0]
+#     import ipdb
+#
+#     ipdb.set_trace()
+
+
```
scanny: Remove commented lines before committing.

tests/unit/test_cube.py
```diff
@@ -47,6 +49,34 @@ def but_it_raises_on_other_cube_response_types(
 
         assert str(exception) == expected_value
 
+    def it_knows_if_it_is_mr_by_itself_to_help(
```
scanny: The `_to_help` suffix is used to indicate a private (helper) method/property. Since this is a public method it is not needed here.

tests/unit/test_cube.py
```diff
@@ -75,6 +105,18 @@ def cube_response_type_fixture(self, request):
...
+            ((DT.CAT, DT.MR, DT.MR), ("alias1", "alias2", "alias2"), True),
+            ((DT.CAT, DT.MR), ("alias1", "alias2"), False),
+            ((DT.CAT, DT.MR, DT.MR), ("alias1", "alias2", "alias3"), False),
+            ((DT.CAT, DT.TEXT, DT.TEXT), ("alias1", "alias2", "alias2"), False),
+        ]
+    )
+    def cube_dimensions_fixture(self, request):
+        dimension_types, aliases, expected_value = request.param
+        return dimension_types, aliases, expected_value
+
```
scanny: Place fixtures in alphabetical order. This aids locating them when search is unavailable.

tests/unit/test_cube.py
```diff
@@ -85,6 +127,10 @@ def cube_(self, request):
     def dimension_types_prop_(self, request):
         return property_mock(request, Cube, "dimension_types")
 
+    @pytest.fixture
+    def cube_dimensions_prop_(self, request):
+        return property_mock(request, Cube, "dimensions")
+
```
scanny: same. respect alphabetical order.

src/cr/cube/matrix.py
ernestoarbitrio: @malecki can u provide a meaningful description for what a shadow prop is?

src/cr/cube/matrix.py
ernestoarbitrio: @malecki also here :D

src/cr/cube/cubepart.py
```diff
@@ -182,6 +190,35 @@ def inserted_column_idxs(self):
     def inserted_row_idxs(self):
         return tuple(i for i, row in enumerate(self._matrix.rows) if row.is_insertion)
 
+    @lazyproperty
+    def insertions(self):
+        """Returns masked array with residuals for insertions
```
malecki: oh, this makes more sense now — even compress_rowcols would leave this whole thing intact. this mask is fine, but we may want to provide a way to get just rows or just columns. 🤔 

What do we serialize in second-order for inf, since json is deficient?

src/cr/cube/matrix.py
```diff
@@ -218,7 +226,24 @@ def factory(cls, cube, dimensions, slice_idx):
         base_counts = cube.base_counts
         counts_with_missings = cube.counts_with_missings
         dimension_types = cube.dimension_types[-2:]
+        if cube.dimension_types == (DT.CAT, DT.MR, DT.MR) and cube.is_mr_by_itself:
 
+            overlap_tstats = calculate_overlap_tstats(
+                _MrXMrMatrix, dimensions, counts, base_counts, counts_with_missings
+            )
+
+            # These are apparent dimensions (user dimensions). Do we need to get all the dims?
```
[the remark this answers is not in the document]
malecki:
```suggestion
            # These are apparent dimensions which hide 'selections' dims behind 'MR'
```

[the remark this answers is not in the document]
malecki: (consider it resolved)

src/cr/cube/matrix.py
malecki:
```suggestion
    def _mr_shadow_proportions(self):
        """
        Cube containing item-wise selections, overlap, and nonoverlap 
        with all other items in a multiple response dimension, for each 
        element of any prepended dimensions: 
        A 1d interface to a 4d hypercube of underlying counts.
        """
```

src/cr/cube/matrix.py
```diff
@@ -725,6 +821,10 @@ def _column_generator(self):
             self.table_margin.T,
         )
 
+    @lazyproperty
+    def _pairwise_overlap_total(self):
+        return np.sum(np.sum(self._counts, axis=1), axis=2)
```
malecki:
```suggestion
        return np.sum(np.sum(self._counts, axis=1), axis=2)
        """
        Given a 4d hypercube of multiple response items, return the
        symmetric square matrix of valid observations between all pairs.
        n1 = 2; n2 = 2; n12 = 1; overlap total = 3
        """
```

src/cr/cube/matrix.py
```diff
@@ -725,6 +821,10 @@ def _column_generator(self):
             self.table_margin.T,
         )
 
+    @lazyproperty
+    def _pairwise_overlap_total(self):
```
malecki:
```suggestion
    def _pairwise_overlap_total(self):
        return np.sum(np.sum(self._counts, axis=1), axis=2)
        """
        Given a 4d hypercube of multiple response items, return the
        symmetric square matrix of valid observations between all pairs.
        n1 = 2; n2 = 2; n12 = 1; overlap total = 3
        """
```

scanny: Docstrings should be of the following PEP 257 form:

```python
def method(self):
    """Return 2D ndarray symmetric-square matrix of valid observations.

    Given a 4D hypercube of multiple-response (MR) items, this method calculates the
    *symmetric square matrix* which is a fancy word for {this simple idea}.
    """
```

The key features are a single line to start, short enough not to wrap, that gives some gist of what to expect but should always start by mentioning the return-type, since that is the effective "type" of the function. Then, when more is called for, a separate passage, separated by a blank line that fills in any important remaining details. The key thing is to describe the contract of the method, i.e. what it gives back and what you need to give it, in that order.

src/cr/cube/cubepart.py
ernestoarbitrio: We use "Infinity" string for `np.inf` ... in the next release we can provide a way to provide only rows, only cols or both; also we can change the mask using the compress thing of numpy

PR #187, opened 2019-12-09
bug fix for means vector margins

ernestoarbitrio:


src/cr/cube/matrix.py
```diff
@@ -1573,6 +1573,7 @@ class _MeansVector(_BaseVector):
 
     def __init__(self, element, base_counts, means):
         super(_MeansVector, self).__init__(element, base_counts)
+        self._base_counts = base_counts
```
scanny: @ernestoarbitrio this line is unnecessary because the base class (`_BaseVector`) already defines `self._base_counts`.

PR #195, opened 2020-04-29
bug fix for wrong proportions with explicit order

ernestoarbitrio:
This issue was found for the following dashboard: https://alpha.crunch.io/dashboard/ff3c, specifically for the tile that is a table showing just a single cell. The dashboard correctly shows the cell value as 48.3, but the export incorrectly shows the cell value as 70.4.

This PR addresses https://www.pivotaltracker.com/story/show/172542747

The main problem was that when we have a specified order in the transforms we need to return 
`self._base_vector.margin[self._opposing_order]`. So we miss the opposing order so for example:

```
values = [12, 34, 45]
margin = [45, 56, 78]

order = [0,2,1]
values = [12, 45, 34]
margin = [45, 56, 78] <- same margin without order
```

So `values/margin` was wrong


src/cr/cube/matrix.py
```diff
@@ -1397,6 +1397,12 @@ def ordering(self):
     def values(self):
         return self._base_vector.values[self._opposing_order]
 
+    @lazyproperty
+    def margin(self):
+        if not isinstance(self._base_vector.margin, np.ndarray):
+            return self._base_vector.margin
+        return self._base_vector.margin[self._opposing_order]
```
scanny, line 1404: Please respect the existing ordering and position this property alphabetically.

scanny: Add comments to describe why this conditional behavior is necessary.

ernestoarbitrio: done

PR #196, opened 2020-05-04
Filter only multitable placeholder 172658719

scanny:
Add `_Strand.cube_index` and `_Strand.title` for to enable handling of filter-only multitable-template placeholders.


src/cr/cube/cube.py
```diff
@@ -188,19 +188,17 @@ class Cube(object):
...
-        transforms=None,
-        first_cube_of_tab=False,
-        population=None,
-        mask_size=0,
+        self, response, cube_idx=None, transforms=None, population=None, mask_size=0
     ):
         self._cube_response_arg = response
         self._transforms_dict = {} if transforms is None else transforms
-        self._first_cube_of_tab = first_cube_of_tab
+        self._cube_idx_arg = cube_idx
```
ernestoarbitrio, line 202: why `_cube_idx_arg` and not `cube_idx`?

src/cr/cube/cube.py
```diff
@@ -233,6 +231,11 @@ def counts(self):
     def counts_with_missings(self):
         return self._measure(self.is_weighted).raw_cube_array
 
+    @lazyproperty
+    def cube_index(self):
```
ernestoarbitrio, line 236: out of curiosity: does this property returns a cube idx or an offset?

tests/unit/test_cube.py
```diff
@@ -392,6 +400,10 @@ def but_it_raises_on_other_cube_response_types(self, cube_response, expected_val
 
     # fixture components ---------------------------------------------
 
+    @pytest.fixture
+    def _cube_dict_prop_(self, request):
```
ernestoarbitrio, line 431: is this fixture component mandatory considering that it's used only in `it_knows_its_title` test? IMHO I would mock that property in the test directly

src/cr/cube/cube.py
scanny: `cube_idx_arg` (the `cube_idx` argument) can be `None` or `int`. Cube index is always an int. So the `_arg` suffix cues the reader to look for some reason why the argument doesn't directly become the served value.

src/cr/cube/cube.py
scanny: `idx` and "offset" are generally synonymous. In this case, it is always a positive int value, the first cube returning `0`, the next `1`, etc. It is never `None` however.

tests/unit/test_cube.py
scanny: Not at the moment, but that's because the unit test coverage is still very spotty on this class. I made the judgement to include it since that property is used in like four other methods so will be needed then.

src/cr/cube/cube.py
ernestoarbitrio: 👍

tests/unit/test_cube.py
ernestoarbitrio: 👍

PR #197, opened 2020-05-11
Fix anchor idx bug 172616313

scanny:
Fix problem where pre-ordering anchor-idx was used for locating inserted subtotal vectors, leading to incorrect placement in many cases when an explicit vector ordering transform was applied to a dimension having inserted vectors.

### DIFF
--- a/src/cr/cube/matrix.py
+++ b/src/cr/cube/matrix.py
@@ -1196,6 +1196,11 @@ class _BaseMatrixInsertedVector(object):
         return self._residuals / np.sqrt(variance)
 
     @lazyproperty
+    def _addend_vectors(self):
+        """Sequence of base-vectors contributing to this inserted subtotal."""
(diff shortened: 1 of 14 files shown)


src/cr/cube/matrix.py
```diff
@@ -97,14 +93,44 @@ def _assembled_rows(self):
 
         Each row vector also reflects any new elements introduced by inserted columns.
         """
-        opposing_insertions = self._inserted_columns
+        return self._assembled_vectors(
+            self._base_rows, self._inserted_rows, self._inserted_columns
+        )
+
+    def _assembled_vectors(
```
ernestoarbitrio, line 100: This method seems to be generic for row or columns, but why the iteration in the return statement is based on `cols`?

scanny: Yes, good catch, that should be vector, I've made the fix, thanks! :)

PR #204, opened 2020-06-09
Two level pairwise t

scanny:
Changes required for two-level pairwise-t display, primarily adding `_Slice.pairwise_indices_alt` and `_Slice.scale_mean_pairwise_indices_alt` along with processing the transforms-dict construction parameter for multiple threshold values.


src/cr/cube/measures/pairwise_significance.py
```diff
@@ -36,87 +56,70 @@ def values(self):
...
-        col_idx,
-        axis=0,
-        weighted=True,
-        alpha=0.05,
-        only_larger=True,
-        hs_dims=None,
-    ):
+class _ColumnPairwiseSignificance(object):
+    """Value object providing matrix of T-score based pairwise-comparison
+    P-values"""
```
ernestoarbitrio, line 79: the first line of the docstring here should be on one line i think

src/cr/cube/measures/pairwise_significance.py
```diff
@@ -126,28 +129,83 @@ def summary_pairwise_indices(self):
...
+        props = self._slice.column_proportions
+        diff = props - props[:, [self._col_idx]]
+        var_props = props * (1.0 - props) / self._slice.column_base
+        se_diff = np.sqrt(var_props + var_props[:, [self._col_idx]])
+        return diff / se_diff
+
+    @lazyproperty
+    def t_stats_correct(self):
+        """It returns the t_statistic for MR variables considering the overlaps
+        """
```
ernestoarbitrio, line 153: this carriage return seems odd

src/cr/cube/measures/pairwise_significance.py
```diff
@@ -126,28 +129,83 @@ def summary_pairwise_indices(self):
...
+
+        tstats_scale_means = (
+            self._slice.scale_means_row[self._col_idx] - self._slice.scale_means_row
+        ) / (standard_deviation * np.sqrt((1 / counts[self._col_idx]) + (1 / counts)))
+
+        return tstats_scale_means
 
     @lazyproperty
     def _df(self):
+        # if the cube to which the slice belongs is a CATxMRxITSELF
```
ernestoarbitrio, line 196: If it's possible i would add a docstring for degree of freedom just summarising the existing comment here

src/cr/cube/measures/pairwise_significance.py
```diff
@@ -12,22 +12,42 @@
...
+
+        Indicates "whole columns" that are significantly different, based on the mean of
+        the scale (category numeric-values) for each column. The length of the array is
+        that of the columns-dimension.
+        """
+        return cls(slice_, alpha, only_larger)._scale_mean_pairwise_indices
+
+    @lazyproperty
+    def summary_pairwise_indices(self):
+        """ndarray containing tuples of pairwise indices for the column summary."""
```
ernestoarbitrio, line 43: considering that we're using the -> convention for the return values in the docstring I would use also here:
```
-> ndarray, tuple of pairwise ... 
```

src/cr/cube/measures/pairwise_significance.py
```diff
@@ -36,87 +56,70 @@ def values(self):
...
             )
-            for col_idx in range(self._slice.get_shape(hs_dims=self._hs_dims)[1])
+            for col_idx in xrange(self._slice.shape[1])
         ]
 
     @lazyproperty
-    def pairwise_indices(self):
-        """ndarray containing tuples of pairwise indices."""
+    def _pairwise_indices(self):
+        """2D ndarray containing tuples of pairwise indices."""
```
ernestoarbitrio, line 68: also here `->`

src/cr/cube/measures/pairwise_significance.py
```diff
@@ -36,87 +56,70 @@ def values(self):
...
-        """ndarray containing tuples of pairwise indices for the column summary."""
-        summary_pairwise_indices = np.empty(
-            self.values[0].t_stats.shape[1], dtype=object
-        )
-        summary_pairwise_indices[:] = [
-            sig.summary_pairwise_indices for sig in self.values
-        ]
-        return summary_pairwise_indices
+    def _scale_mean_pairwise_indices(self):
+        """ndarray containing tuples of pairwise indices."""
```
ernestoarbitrio, line 73: and here

tests/expectations/cat-hs-x-mr-pw-idxs.py
```diff
@@ -0,0 +1,10 @@
+[
+    [(1, 2, 3, 4), (2, 3), (), (), (2,)],
+    [(), (), (), (), (3,)],
+    [(1, 2, 3, 4), (2, 3), (), (), (3,)],
+    [(), (), (), (), ()],
+    [(), (0,), (0,), (0,), (0,)],
+    [(), (), (), (0, 1, 4), ()],
+    [(), (), (), (), ()],
+    [(), (0,), (0, 1), (0, 1, 4), (0,)],
+]
```
ernestoarbitrio, line 10: out of curiosity, saving numpy array on disk using py file instead of binary npy files is just for having readable expectations? cause i always used .npy file to preserve numpy arrays on disk and also to reduce size ... but i don't know if is this the case to use the binary files rather than normal py files

ernestoarbitrio: oh, maybe they're not all numpy arrays but standard python structures

src/cr/cube/cubepart.py
```diff
@@ -109,13 +110,97 @@ def variable_name(self):
...
+        """Pair (tuple) of confidence-interval thresholds to be used for t-tests.
+
+        The second value is optional and is None when no secondary alpha value was
+        defined for the cube-set.
+        """
+        value = self._transforms_dict.get("pairwise_indices", {}).get("alpha")
+
+        # --- handle omitted, None, [], (), {}, "", 0, and 0.0 cases ---
+        if not value:
+            return (0.05, None)
```
ernestoarbitrio, line 138: Can we use `value = self._transforms_dict.get("pairwise_indices", {}).get("alpha", (0.05, None))` and avoid this if statement?

PR #206, opened 2020-06-23
fix: collapsed scale-mean-pairwise-indices

scanny:
The array of scale-mean pairwise-indices collapses when all the tuples it contains are empty, producing a wrong shape. Change interface to return a tuple of tuples and avoid numpy idiosyncrasies on edge cases.


scanny:
@ernestoarbitrio can you push this as release 1.11.27 if you approve on the CR?

PR #216, opened 2020-08-14
Document matrix.py and stripe.py

scanny:
Add docstrings to each property in matrix.py and stripe.py. Find bugs during that process and fix them. Improve naming where necessary. Perform refactoring in places pleading for it once it becomes clear what the property's contract is.


scanny:
@ernestoarbitrio @slobodan-ilic: @malecki wanted you both to have a look. This is a lot of lines (+2,278 −1,930) but it's well groomed so I think it will be faster and easier to review it commit-by-commit. Welcome any comments you can give. I learned a lot about the guts by working out what the docstrings should be. The `rfctr: improve matrix.py docstrings` commit is the big one.

src/cr/cube/dimension.py
```diff
@@ -828,8 +824,16 @@ def __len__(self):
...
+        # same element-id.
+
+        def occurrences(s, lst):
+            return (i for i, e in enumerate(lst) if e == s)
+
         anchors = self._anchors
-        occurrences = lambda s, lst: (i for i, e in enumerate(lst) if e == s)
         top_anchors = list(occurrences("top", anchors))
         int_anchors = [a + len(top_anchors) for a in anchors if isinstance(a, int)]
         bottom_anchors = [
```
ernestoarbitrio, line 839: FYI: this prop is removed in https://github.com/Crunch-io/crunch-cube/pull/215

src/cr/cube/matrix.py
```diff
@@ -671,32 +958,22 @@ def _row_generator(self):
         )
 
     @lazyproperty
-    def _table_std_dev(self):
```
ernestoarbitrio, line 674: why this property has been removed?

src/cr/cube/matrix.py
```diff
@@ -850,53 +1188,105 @@ def _zscores(self):
...
             )
         )
 
 
-class _MrXMrMatrix(_MatrixWithMR):
+class _MrXMrMatrix(_CatXCatMatrix):
     """Represents MR x MR slices.
 
-    Needs to properly index both rows and columns (selected/not-selected).
+    It's `._counts` is a 4D ndarray with axes (rows, selected-and-not, cols,
```
ernestoarbitrio, line 1220: I think it's only a typo ... `its` pronoun instead of `it's`

src/cr/cube/matrix.py
```diff
@@ -850,53 +1188,105 @@ def _zscores(self):
...
+
+         [[[1822.67560537 2883.99243344]
+           [1616.70492531 3089.96311351]
+           [1735.59793395 2971.07010487]]
+
+          [[2365.46106889 4970.87137654]
+           [2011.71672718 5324.61571825]
+           [3398.16687766 3938.16556777]]]]
+
+    It's rows and columns are both `_OpposingMrVector` objects.
```
ernestoarbitrio, line 1258: also here

src/cr/cube/matrix.py
```diff
@@ -2616,13 +2620,14 @@ def __init__(
             table_std_err,
             column_index,
         )
-        self._all_unweighted_counts = unweighted_counts
-        self._all_counts = counts
+        self._both_counts = counts
+        self._both_unweighted_counts = unweighted_counts
```
ernestoarbitrio, line 2620: I don't know if `both` could be a meaningful name ... if i got the point here, all counts are referring to the complete cases for the MR. So what about `complete_counts`?

src/cr/cube/matrix.py
```diff
@@ -766,8 +1009,33 @@ def rows(self):
...
+          [1255.87038944 1244.28291533]]
+
+         [[   9.83166357   25.9551254 ]
+          [   8.23140253   27.55538645]
+          [  22.214956     13.57183298]]]
+
+    Each value is np.float64 if the cube-result is weighted (as in this example), or
+    np.int64 if unweighted.
+
+    It's rows are `_OpposingMrVector` objects. Its columns are `_MrOpposingCatVector`
```
ernestoarbitrio, line 1036: Typo `It's`

src/cr/cube/dimension.py
```diff
@@ -828,8 +828,16 @@ def __len__(self):
     @lazyproperty
     def anchor_idxs(self):
         """List of int indicating the actual position of the subtotals."""
+        # TODO: this appears wrong, using an anchor (element_id) as an index (offset of
```
slobodan-ilic, line 827: I'm not sure that I understand this comment. Does it refer to how we specify the anchors, or to what this method returns? Because the method seems to return the indexes, as its name suggests. If this is in relation to how we specify things, then I don't think it's a good place for this comment (i.e. it's not responsibility of this method's scope to eventually fix it in the future).

src/cr/cube/cube.py
```diff
@@ -289,15 +289,18 @@ def inflate(self):
         )
 
     @lazyproperty
-    def is_mr_by_itself(self):
-        """True if the cube contains MRxItself as last 2 dimensions."""
+    def is_mr_aug(self):
```
slobodan-ilic, line 288: I like the shortening of the prop name. However, coming from crunchdb after a while, I find myself _guessing_ what the `MR_AUG` cube is. I _think_ it's the one we use to calculate the selected vs non-selected significance, that we did sometime after Montreal. Anyhow, I'd feel more reassured, as a reader, if I found at least one more sentence in the last paragraph, that would give more context, as to the usage of such cubes. Something like. An `MR_AUG` cube is used to calculate statistical significance of complete-cases, and is created by artificially augmenting the original cube's dimensions, such as [MR], or [CAT x MR], by an additional MR dimension. The extra MR dimension must be the _same_ as the one originally figuring in the cube's dimensions. After the augmentation, it becomes [MR x MR], or [CAT x MR x MR].

After typing this out, I'm aware that it's a bit too elaborate, so I leave it to your best judgement :)

src/cr/cube/cubepart.py
```diff
@@ -66,7 +66,7 @@ def cube_index(self):
         return self._cube.cube_index
 
     @lazyproperty
-    def cube_is_mr_by_itself(self):
+    def cube_is_mr_aug(self):
```
slobodan-ilic, line 69: Seeing the first word as `cube`, and the `lazyproperty` on top, automatically directs me to thinking that it's some kind of an object. I'd probably prefer the `is_cube_mr_aug`, if it's a relatively easy refactor.

That said, if a cube has `is_mr_aug`, why can't a cube-part have the same name for the property? Class membership should be enough to determine if it's in cube or cube-part. But then again, does a "is mr aug" make sense in the cube-part context?

Probably just the first of the two paragraphs makes sense (to actually act on), but I'm being verbose on purpose, as the nature of the work is such.

src/cr/cube/matrix.py
```diff
@@ -317,9 +317,9 @@ def factory(cls, cube, dimensions, slice_idx):
                 counts_with_missings = counts_with_missings[0]
         if dimension_types == (DT.MR, DT.MR):
             return _MrXMrMatrix(dimensions, counts, base_counts, counts_with_missings)
-        elif dimension_types[0] == DT.MR:
+        if dimension_types[0] == DT.MR:
```
slobodan-ilic: 💯

src/cr/cube/cube.py
```diff
@@ -366,6 +362,17 @@ def title(self):
...
         return self._cube_dict["result"].get("title", "Untitled")
 
+    @lazyproperty
+    def unweighted_counts(self):
+        """ndarray of unweighted counts, valid elements only.
+
+        Unweighted counts are drawn from the `result.counts` field of the cube result.
+        These counts are always present, even when the measure is `mean` and there are
+        no count measures. These counts are always unweighted, regardless of whether the
+        cube is "weighted".
```
slobodan-ilic, line 372: 💯

src/cr/cube/cubepart.py
```diff
@@ -756,6 +752,11 @@ def table_std_dev(self):
     def table_std_err(self):
         return np.array([row.table_std_err for row in self._matrix.rows])
 
+    @lazyproperty
+    def unweighted_counts(self):
+        """2D tuple of unweighted counts."""
```
slobodan-ilic, line 759: I guess the subsequent commits will introduce similar change for other properties. However, I'm not sure that "2D tuple" is clear enough in terms of what we actually put in. If it's a 2D `ndarray`, it's clear by definition, because the concept of a dimension is very well defined there, and very common. Not as much in the tuple world, afaict. I'll get back to this after I review the whole thing. I guess the main motivation behind this change is to have an unmutable type returned, it that right?

src/cr/cube/matrix.py
```diff
@@ -1511,6 +1505,19 @@ def table_std_err(self):
     def table_proportions(self):
         return self.values / self._base_vector.table_margin
 
+    @lazyproperty
+    def unweighted_counts(self):
+        """1D ndarray of unweighted count for each base element and insertion.
```
slobodan-ilic: `"count"` -> `"counts"` in the docstring as well?

src/cr/cube/matrix.py
```diff
@@ -1659,6 +1662,11 @@ def table_std_err(self):
     def table_proportions(self):
         return self._base_vector.table_proportions[self._visible_element_idxs]
 
+    @lazyproperty
+    def unweighted_counts(self):
+        """1D ndarray of unweighted count for each visible element in vector."""
```
slobodan-ilic: same here? or not?

src/cr/cube/matrix.py
```diff
@@ -1741,6 +1745,11 @@ def ordering(self):
         """
         return (self._index, self._index, self)
 
+    @lazyproperty
+    def unweighted_counts(self):
+        """1D ndarray of unweighted count reordered to opposing-vector ordering."""
```
slobodan-ilic: counts?

src/cr/cube/matrix.py
```diff
@@ -1899,6 +1904,11 @@ def proportions(self):
     def table_margin(self):
         return self._table_margin
 
+    @lazyproperty
+    def unweighted_counts(self):
+        """1D ndarray of unweighted count for each valid element in vector."""
```
slobodan-ilic: Ok, I see a pattern now, I'd like an explanation of why we want to use "of unweighted count" instead of "of unweighted counts" :)

src/cr/cube/stripe.py
```diff
@@ -19,15 +19,28 @@
 
 
 class TransformedStripe(object):
-    """Stripe reflecting application of all transforms."""
+    """Stripe reflecting application of all transforms.
```
slobodan-ilic, line 22: I might be wrong in my gut feeling, but I feel a need for a comma after the word "stripe". Either that or a more redundant construction "that reflects", or some similar indirection.

src/cr/cube/stripe.py
```diff
@@ -52,12 +65,25 @@ def rows_including_hidden(self):
...
+
+        Value is an np.int64 scalar except when this stripe is multiple-response (MR).
+        This value is a 1D ndarray of np.int64 for an MR stripe.
+        """
         return self._base_stripe.table_base
 
     @lazyproperty
     def table_margin_unpruned(self):
-        """Hmm, weird 1D ndarray with same float value repeated for each row."""
+        """np.float/int64 scalar or 1D np.float/int64 ndarray of table margin.
```
slobodan-ilic, line 77: Maybe add an explanation as to why we need this at all? Normally, you'd want values that are either pruned, or not, according to transformations. So in which case we'd need this? It's not necessary, but might be instructive for a doc-string reader.

src/cr/cube/stripe.py
```diff
@@ -194,6 +222,7 @@ def __init__(self, rows_dimension, counts, unweighted_counts):
 
     @lazyproperty
     def rows(self):
+        """Sequence of _CatStripeRow for each valid element in rows-dimension."""
```
slobodan-ilic, line 225: we're returning a `tuple` here, not a `Sequence`

src/cr/cube/stripe.py
```diff
@@ -217,6 +249,7 @@ def __init__(self, rows_dimension, means, unweighted_counts):
 
     @lazyproperty
     def rows(self):
+        """Sequence of _MeansStripeRow for each valid element in rows-dimension."""
```
slobodan-ilic, line 252: same here

src/cr/cube/stripe.py
```diff
@@ -239,6 +272,7 @@ def __init__(self, rows_dimension, means, unweighted_counts):
 
     @lazyproperty
     def rows(self):
+        """Sequence of _MeansMrStripeRow for each valid element in rows-dimension."""
```
slobodan-ilic, line 275: Ok, I see a pattern now, is there a reason why you used the `Sequence` instead of `tuple`?

src/cr/cube/stripe.py
```diff
@@ -465,12 +518,16 @@ def count(self):
...
     @lazyproperty
     def pruned(self):
-        """True if this row is "empty"."""
+        """True if this question was answered by exactly zero respondents."""
         return self._unweighted_count == 0 or np.isnan(self._unweighted_count)
 
     @lazyproperty
     def table_proportions(self):
-        """float between 0.0 and 1.0 quantifying this row's share of stripe total."""
+        """float between 0.0 and 1.0 quantifying this row's share of stripe total.
```
slobodan-ilic, line 533: Could we decide on either using the arrow `"->"` or the combination of `"between"` and `"and"` for consistency?

src/cr/cube/stripe.py
```diff
@@ -519,9 +576,9 @@ class _MeansMrStripeRow(_BaseStripeRow):
 
     def __init__(self, element, unweighted_counts, means):
         super(_MeansMrStripeRow, self).__init__(element)
-        # ---unweighted_counts is an int array like [336 5501]---
+        # ---unweighted_counts is a two-value int array like [336 5501]---
```
slobodan-ilic, line 587: Not sure why though... I can see how this comment would be important for the reader, but it lacks context. At least, I cannot figure it without research (into code) _why_ this should ever be exactly two values. Maybe add some context if it's not too verbose?

src/cr/cube/stripe.py
```diff
@@ -557,22 +614,27 @@ def __init__(self, element, counts, unweighted_counts, table_margin):
 
     @lazyproperty
     def count(self):
+        """Number of weighted respondents who seleted this response."""
```
slobodan-ilic, line 625: maybe add type?

src/cr/cube/stripe.py
```diff
@@ -557,22 +614,27 @@ def __init__(self, element, counts, unweighted_counts, table_margin):
...
         return self._counts[0]
 
     @lazyproperty
     def pruned(self):
+        """True if this question was asked of exactly zero actual respondents."""
         return self.table_base == 0
 
     @lazyproperty
     def table_base(self):
+        """Sample size for MR question, how many actual folks were asked."""
```
slobodan-ilic, line 635: type info?

slobodan-ilic: Same for all in this file...

src/cr/cube/matrix.py
```diff
@@ -26,11 +26,33 @@ def __init__(self, unordered_matrix):
...
+        """ -> TransformedMatrix object constructed from values of `cube`.
+
+        `cube` is the `cr.cube.Cube` object containing the data for this matrix. Note
+        that not all the data in `cube` will necessarily be used by this matrix. When
+        `cube` is more than 2-dimensional, it is "sliced" and each slice gets its own
+        matrix (and `_Slice` object).
+
+        `dimensions` is a pair (2-tuple) of (rows_dimension, columns_dimension)
+        Dimension objects. These are based on two of the dimensions of `cube` but may
+        not be the first two and may and often do have transformations applied that are
```
slobodan-ilic, line 38: They _are_ always the last two though, no? Maybe write it like that, it's more explicit. Unless I've forgotten something about how cube works 🤷

src/cr/cube/matrix.py
```diff
@@ -26,11 +26,33 @@ def __init__(self, unordered_matrix):
...
+        not present on the `cube` dimensions from which they derive.
+
+        `slice_idx` is an int offset indicating which portion of `cube` data to use for
+        this matrix.
```
slobodan-ilic, line 42: A trick from an ex UX colleague - don't leave the orphan words for new line - it's mostly for when you leave a single word for the new line, that it gets skipped in reading. Here we have two words, but still, if you divided the sentence into `". . . of cube data"` and `"to use for this matrix."`, it'd be more balanced across lines (even though further away from the 88 limit).

src/cr/cube/matrix.py
```diff
@@ -55,22 +85,70 @@ def rows(self):
...
+        a scalar value for this property.
+        """
         return self._unordered_matrix.table_base
 
     @lazyproperty
     def table_margin(self):
+        """np.float/int64 scalar or a 1D or 2D np.float/int64 ndarray table margin.
+
+        The table margin is the overall sample size of the matrix. This is the weighted
+        count of respondents who were asked both questions and provided a valid response
```
slobodan-ilic, line 123: I don't get what "both" means in this context? Which both questions? Did you mean "who were both asked the questions, and provided a valid response?". Or did I miss something altogether?

src/cr/cube/matrix.py
```diff
@@ -266,24 +354,44 @@ def factory(cls, cube, dimensions, slice_idx):
...
             "must be implemented by each subclass"
         )  # pragma: no cover
 
     @lazyproperty
     def _column_elements(self):
+        """Sequence of `cr.cube.dimension._Element` object for each matrix column."""
         return self.columns_dimension.valid_elements
 
-    @lazyproperty
+    @property
```
slobodan-ilic, line 402: 💯

src/cr/cube/matrix.py
```diff
@@ -472,38 +589,87 @@ def rows(self):
 
     @lazyproperty
     def table_base(self):
+        """np.int64 count of actual respondents who answered both questions.
```
slobodan-ilic, line 607: Oh, now I think I know... both refers to the cross-section from _both_ dimensions? Ok, I'll give the PR another pass, and rethink the docstrings from before.

src/cr/cube/matrix.py
```diff
@@ -472,38 +589,87 @@ def rows(self):
...
+
+        Each dimension of a CAT_X_CAT matrix represents a categorical question. Only
+        responses that answered both those questions appear as entries in the valid
```
slobodan-ilic, line 610: respondents instead of responses?

src/cr/cube/matrix.py
```diff
@@ -472,38 +589,87 @@ def rows(self):
...
         return np.sum(self._unweighted_counts)
 
     @lazyproperty
     def table_margin(self):
+        """Scalar np.float/int64 weighted-N for overall table.
+
+        This is the weighted count of respondents who provided a valid response to
+        both questions. Because both dimensions are CAT, the table-margin value is the
+        same for all cells of the matrix. Value is np.int64 when source cube-result is
+        unweighted.
```
slobodan-ilic, line 623: orphan

src/cr/cube/matrix.py
```diff
@@ -472,38 +589,87 @@ def rows(self):
...
+        """1D ndarray of np.float64 (or np.int64) weighted N for each matrix column.
+
+        These values are required for zscore calculations at vector level but must be
+        calculated at matrix level. The values are `np.int64` when the source
+        cube-result is not weighted.
+        """
         return np.sum(self._counts, axis=0)
 
-    @lazyproperty
+    @property
```
slobodan-ilic, line 685: ❤️

src/cr/cube/matrix.py
```diff
@@ -472,38 +589,87 @@ def rows(self):
...
+        Column-index answers the question "are respondents in this row-category more or
+        less likely than the overall table population to choose the answer represented
+        by this column?". For example, if the row is "Hispanic" and the column is
+        home-ownership, a value of 100 indicates hispanics are no less and no more
+        likely to own their home than the overall population. If that value was 150, it
+        would indicate hispanics are 50% more likely to own their home than the general
+        population (or the population surveyed anyway).
+
+        These values must be known by vectors but cannot be calculated there (the
+        calculation is "cross-vector") so must these values must be passed down from the
```
slobodan-ilic, line 670: double "must"

src/cr/cube/matrix.py
```diff
@@ -472,38 +589,87 @@ def rows(self):
...
+        cube-result is not weighted.
+        """
         return np.sum(self._counts, axis=0)
 
-    @lazyproperty
+    @property
     def _row_generator(self):
+        """Iterable of arguments to row-vector constructor call for each row element.
+
+        Used by `.rows` propert. Cannot be a lazyproperty because an iterator is
```
slobodan-ilic, line 689: propert -> property

src/cr/cube/matrix.py
```diff
@@ -516,6 +682,12 @@ def _row_generator(self):
 
     @lazyproperty
     def _rows_margin(self):
+        """1D np.float/int64 ndarray of weighted-N for each matrix row.
+
+        These "opposing-margin" values are required for vector-level calculations but
+        must be calculated at matrix level. The values are `np.int64` when the source
```
slobodan-ilic, line 708: please add the part about calculation that is "cross-vector", similar to the one above. it's priceless.

src/cr/cube/matrix.py
```diff
@@ -561,6 +735,12 @@ def _variance(self):
 
     @lazyproperty
     def _zscores(self):
+        """2D ndarray of np.float64 std-res value for each cell of matrix.
+
+        A z-score is also known as a *standard score* and is the number of standard
+        deviations above (positive) or below (negative) the population mean a cell's
+        value is.
```
slobodan-ilic, line 741: close to orphan

src/cr/cube/matrix.py
```diff
@@ -614,16 +801,34 @@ def _array_type_std_res(counts, total, colsum, rowsum):
 
 
 class _MrXCatMatrix(_MatrixWithMR):
-    """Represents MR x CAT slices.
+    """Represents MR_X_CAT slices.
+
+    It's `._counts` is a 3D ndarray with axes (rows, selected/not, cols), like:
```
slobodan-ilic, line 803: Its

src/cr/cube/matrix.py
```diff
@@ -614,16 +801,34 @@ def _array_type_std_res(counts, total, colsum, rowsum):
...
+          [ 71  76  49  58]]
 
-    It's similar to CAT x CAT, other than the way it handles columns. For
-    columns - which correspond to the MR dimension - it needs to handle the indexing
-    of selected/not-selected correctly.
+         [[  0   0   0   0]
+          [428 491 290 429]]]
+
+    Each value is np.float64, or np.int64 if the cube-result is unweighted (as in this
+    example).
```
slobodan-ilic, line 818: orphan

src/cr/cube/matrix.py
```diff
@@ -614,16 +801,34 @@ def _array_type_std_res(counts, total, colsum, rowsum):
...
+
+    It's rows are `_MrOpposingCatVector` objects. Its columns are `_OpposingMrVector`
+    objects.
```
slobodan-ilic, line 821: same

scanny:
@slobodan-ilic @ernestoarbitrio @malecki I just pushed fixes in response to the comments. I'd like to get this merged so Ernesto can rebase onto it; I'd really like to avoid having to rebase something this big onto other changes :)

Anyway, if you can re-review or provide any new comments you have we can maybe get this merged tomorrow.

src/cr/cube/matrix.py
```diff
@@ -528,7 +528,7 @@ def _row_elements(self):
         return self.rows_dimension.valid_elements
 
     @lazyproperty
-    def _valid_rows_idxs(self):
+    def _valid_row_idxs(self):
```
slobodan-ilic, line 545: I'm not so sure I like the new name better. Row indexes (as it is changed to) implies that we're dealing with the indexes of a row. The first line of the docstring also uses the same noun "rows". Ideally I'd put in something like "new rows' indexes", but we can't do that in python method definition. But I'll leave it to your best judgement... We also need to take into account who'll ever read this, so the new name _might_ be OK (even though I still dislike it :) ).

src/cr/cube/matrix.py
```diff
@@ -646,8 +646,13 @@ def _baseline(self):
         columns-var question) are included. This ensures that the baseline is not
         distorted by a large number of missing responses to the columns-question.
         """
-        dim_sum = np.sum(self._all_counts, axis=1)[self._valid_rows_idxs]
-        return dim_sum[:, None] / np.sum(dim_sum)
+        # --- uncond_row_margin is a 1D ndarray of weighted total observation count
```
slobodan-ilic, line 663: count or counts?

src/cr/cube/matrix.py
```diff
@@ -2527,7 +2532,51 @@ def values(self):
...
+    # only used for column-oriented things like column-proportions and column-labels.
+    # Still, could be bugs lurking here as to how columns handle means.
+
+    @lazyproperty
+    def base(self):
+        """np.int64 unweighted-N for this vector."""
+        return np.sum(self._unweighted_counts[0])
+
+
+class _MrOpposingCatVector(_CategoricalVector):
```
slobodan-ilic, line 2599: I do like this name better than the old one. Although the old one also made some sense (at least to me when I was designing it). I don't remember what that sense was though, at this time.

Anyways, I think we can be even more precise with this name. This is good for now, but maybe something more mathematically sound would be better. What is this vector actually? Is it an "row of a mr x cat matrix" most of the time? Or is it a "column of a cat x mr matrix"? Or both?

The word that's currently ringing in my mind is "orthogonal", potentially giving a name similar to `MrOtrhogonalToCatVector`, but this might be too cumbersome (if even true).

Leave it to your best judgement, but this is also good for now.

src/cr/cube/matrix.py
```diff
@@ -2646,7 +2651,7 @@ def zscore(self):
         return self._zscore
 
 
-class _MultipleResponseVector(_CategoricalVector):
+class _OpposingMrVector(_CategoricalVector):
     """CAT or MR vector that opposes an MR dimension.
```
slobodan-ilic, line 2666: The commit message for this one is _too_ valuable and well-written to allow it to be lost in `git` history. Please include it in the docstring of the class ❤️

slobodan-ilic: Also, as for the other class, the word "orthogonal" just won't leave me alone. I think it describes our use case better than "opposing", which might mean anything.

src/cr/cube/matrix.py
slobodan-ilic: @scanny I agree with @ernestoarbitrio here, I don't have _any_ context for the word "both" here. If you insist on using that word, please provide at least an explanation as to what it refers to. Maybe in a docstring or something.

src/cr/cube/dimension.py
```diff
@@ -640,11 +638,9 @@ class _Element(object):
     This object resolves the transform cascade for element-level transforms.
     """
 
-    def __init__(self, element_dict, index, element_dicts, element_transforms):
+    def __init__(self, element_dict, index, element_transforms):
```
slobodan-ilic, line 641: ❤️

src/cr/cube/cubepart.py
```diff
@@ -806,8 +806,8 @@ def var_scale_means_row(self):
         return np.nansum(numerator, axis=0) / denominator
 
     @lazyproperty
-    def zscore(self):
-        return np.array([row.zscore for row in self._matrix.rows])
+    def zscores(self):
```
slobodan-ilic, line 809: I don't know if there's a higher level API that uses this, and describes what it is, but a docstring surely would be nice. Again, the comment from the commit message (https://github.com/Crunch-io/crunch-cube/pull/216/commits/128dbbc73e40ecda563c3ee185743804616b8b2c) seems valuable.

src/cr/cube/matrix.py
```diff
@@ -2439,14 +2439,6 @@ def pruned(self):
         """True if this vector contains no samples."""
         return self.base == 0
 
-    @lazyproperty
```
slobodan-ilic, line 1807: ❤️

src/cr/cube/matrix.py
```diff
@@ -2378,7 +2378,7 @@ def base(self):
         a 1D np.int64 ndarray. A vector opposing a CAT dimension has a single base and
         produces a scalar np.int64 value.
         """
-        return np.sum(self._unweighted_counts)
+        return np.sum(self._unweighted_counts, axis=0)
```
slobodan-ilic, line 2396: This is unusual, I wasn't expecting this... Can you give some more context to this? Do we need to update the docstring? It might be that I should review exactly what this does, and that lack of memory is the only reason for this comment, but still :)

src/cr/cube/cubepart.py
```diff
@@ -570,10 +573,10 @@ def scale_median_column(self):
         The median is calculated using the standard algebra applied to the numeric
         values repeated for each related counts value
```
malecki, line 574: this could get us into trouble with weights — how does repeat deal with floats? 
weighted median is a kind of weird case to begin with, but this method isn't fully robust to weighted counts.

src/cr/cube/cubepart.py
malecki: I hadn’t realized this. It makes some sense, but a little surprised that the requester of this property doesn’t expect an ndarray?

src/cr/cube/matrix.py
```diff
@@ -55,22 +86,83 @@ def rows(self):
...
+        dimensions are MR, the return value is a 2D ndarray and there is a distinct
+        table-base value for each "cell" of the matrix. A CAT_X_CAT matrix produces
+        a scalar value for this property.
+        """
         return self.table_base_unpruned[
             np.ix_(self._visible_rows_mask, self._visible_cols_mask)
         ]
 
     @lazyproperty
     def table_base_unpruned(self):
```
malecki, line 103: We should consider moving ‘prune’ to be a computed-hide transform so vectors can only be inserted/removed via the transform route

src/cr/cube/matrix.py
```diff
@@ -55,22 +86,83 @@ def rows(self):
...
+        # those for themselves. So this will require reconstructing that "show-ranges"
+        # requirement and either adding some sort of a `.range` property that returns
+        # a sequence of (min, max) tuples, or maybe just returning margin or base as
+        # tuples when appropriate and having something like a `.margin_is_ranges`
+        # predicate the client can switch on to control their rendering.
         return self._unordered_matrix.table_base
 
     @lazyproperty
     def table_margin(self):
+        """np.float/int64 scalar or a 1D or 2D np.float/int64 ndarray table margin.
```
malecki, line 132: doug-note: should probably be ‘total’, though table is probably not going to hurt us.

src/cr/cube/matrix.py
```diff
@@ -224,20 +317,29 @@ def _inserted_rows(self):
     @lazyproperty
     def _row_order(self):
         """ -> 1D ndarray of int row idx specifying order of unordered-array rows."""
-        # ---Specifying int type prevents failure when there are zero rows---
+        # --- specifying int type prevents failure when there are zero rows ---
         return np.array(self._rows_dimension.display_order, dtype=int)
```
malecki, line 321: that’s … an interesting side effect of dtype. why is this the case? do we ever go as far as calling row_order on an empty table?

src/cr/cube/matrix.py
scanny: added "the" weighted total (for each valid row). This form is consistent with other collection description language.

src/cr/cube/matrix.py
scanny: It's used for rows of MR_X_CAT and for columns of CAT_X_MR. It's an abstraction of that symmetry, like if you're an MR vector and your opposing vectors are CAT then this is the vector class for you.

I discussed preferring "orthogonal" to "opposing" with Mike and he said he knew of no statistical-world convention of using "orthogonal" in this way that we might want to respect as a domain convention, and also that he found "opposing" descriptive. Add that to the fact that we use it broadly now and I'm not going to change it at the moment. I will keep it in the back of my mind though and continue to consider it.

src/cr/cube/cubepart.py
scanny: I guess I figure one of the design principles for the `cr.cube` public interface is that users of this library should have no direct need for array operations since they are using the results for rendering or reporting types of purposes. An example would be whaam, who uses it via the SOA interface. Everything the library can express can be expressed as primitives that could, for example, be directly serialized to JSON. Whatever whaam gets from SOA, it's not ndarrays once it gets over there.

I haven't thought that through 100%, but I have noticed that array types and operations have worked their way into `exporter` code where they don't "natively" belong. We shouldn't have to do any additional analysis operations like dividing one array by another and if the ndarray-ness of a returned value needs to be taken advantage of, I'm thinking that represents work that probably belongs in `cr.cube` (and there is at least a couple of examples of that I can think of).

Now I don't feel strongly enough about this to propose we convert all the external interface to return Python primitives right now, but I think the principle is probably pretty correct. Anyway, that's what I had in mind with this.

Aside from its mutability, an ndarray has some surprising behaviors to someone who's not familiar and I can't think of a good reason that clients of `cr.cube` should have to be familiar with `numpy` (or anything outside native Python) to be able to use it effectively.

src/cr/cube/cubepart.py
scanny: Yeah, I think this whole module needs a thorough documentation pass like we've done here to `matrix.py` and to some extent to `stripe.py`. I had to draw the scope line somewhere just because it's such a big job.

src/cr/cube/matrix.py
scanny: Yeah, that's sounds like a good idea. We could probably add it to the dimension transforms schema, maybe `"prune": true`. I know end-users think of it as show-empty with opposite boolean semantics, but once we're inside the technical realm, "prune" means do something and "don't prune" means "don't do something" so it's less of a brain-twister to interpret the semantics.

An possible advantage of dimension-level specification would be you could say to prune rows but not columns. It would default to `true` so you wouldn't often need to say it in two places. Also an advantage is we wouldn't need to create a new "table" level transforms container to put this in; so far we haven't had a need for transforms at that level.

src/cr/cube/matrix.py
scanny: Adding the `dtype=int` (and the comment) actually came as a fix to a Sentry error that would come up in rashes maybe a year and a half ago. We don't have conditional logic that prevents a table with zero rows or zero columns from being processed, and `._row_order` is a basic attribute that is accessed unconditionally in the course of all `_Slice` computations, so yeah, we need it to work for a table with no valid rows.

I expect the error came up in the course of tabbook exports where folks were selecting a bunch of variables without necessarily looking too closely at them.

src/cr/cube/matrix.py
scanny: Okay, this might have been an extraction too far, but I documented this like so:

```python
        # --- The `axis=0` parameter makes this work for both the X_MR and X_CAT cases.
        # --- A vector opposing an MR dimension (X_MR) has a 2D unweighted_counts like:
        #
        #      [[653  128  120  389  469]
        #       [101   95   96   69   61]]
        #
        # --- where the first axis is selected/unselected and the second dimension is
        # --- a distinct value for each MR-subvar. The sum must include both selected
        # --- and unselected values, but cannot sum across subvars. So the result is
        # --- a 1D array for the X_MR case.
        # ---
        # --- In the X_CAT case, unweighted_counts is a 1D array like:
        #
        #      [246  61  102  318  469]
        #
        # --- Summing across axis-0 gives the right (scalar) result here too, even
        # --- though specifying the axis is superfluous for this case.
```

PR #220, opened 2020-10-01
Initial 🗡🩸 at Numeric Arrays means in 🎲

slobodan-ilic:
This PR uses two new JSON payloads obtained from production. It's generated by using the `simple_alltypes` dataset and adding a new `"numeric array"` variable to it (trough superadmin). It aims to demonstrate how we need to treat measures for variables that include additional dimensions (which changes the original shape of the measure). For example, the `Nub_` measure becomes a `Stripe` when this additional (subvariables) dimension is present. `Stripe_` becomes the matrix.

The work is currently done in a way that expands on the functionality of the measure classes, so that they're aware of their own shape (that might be different than the shape of the original cube, because of the ☝️ reasons).

We could do this more rigorously, and implement a separate measure class for the cases when there are subvariables present or not. But given that the measure classes already handle different shapes (nub, stripe, matrix), I've opted for the simpler solution - just being _aware_ of the shape.


src/cr/cube/cube.py
```diff
@@ -595,6 +598,21 @@ def __init__(self, cube_dict, all_dimensions):
...
+        When we have variables in measures' specification, that are arrays, or when we
+        specify measures such as correlation (that are themselves matrices), this shape
+        needs to change.
+
+        Each subclass is responsible for implementing the correct shape, based on the
+        shape of the provided variable (and if it has subvariables or not), and on the
+        provided measure, and its innate shape.
+        """
+        return self._all_dimensions.shape
+
```
ernestoarbitrio, line 615: keep the alphabetical order for class property

src/cr/cube/cube.py
```diff
@@ -620,6 +638,30 @@ def _flat_values(self):  # pragma: no cover
...
+        # Original shape: (m, n, . . .,)
+        shape = super(_MeanMeasure, self).shape
+
+        if self._subvariables is not None:
+            # There are subvariables in the provided variable => augment shape
+            # Shape now becomes: (m, n, . . ., n_subvars,)
+            shape += (len(self._subvariables),)
+
+        return shape
+
```
ernestoarbitrio, line 664: I'd like to see a unit test for this specific property, maybe parametrised to cover the more meaningful edge cases

src/cr/cube/cube.py
```diff
@@ -263,6 +262,12 @@ def has_means(self):
         """True if cube includes a means measure."""
         return self._measures.means is not None
 
+    @lazyproperty
+    def means_subvariables(self):
+        if not self.has_means:
+            return None
+        return self._measures.means.subvariables
+
```
ernestoarbitrio, line 270: Also this property should be right before the `missing` one. just to respect the alphabetical oder

src/cr/cube/cube.py
slobodan-ilic: Done

src/cr/cube/cube.py
ernestoarbitrio: This name `means_subvariables` is odd to me 🤷‍♂️  , can you explain me what's a subvariable of means? I mean i got from the code that the subvariables belongs to a variable and they're than retrieved in the MeanMeasure, but i don't get te sense of this name ... sorry

tests/unit/test_cube.py
```diff
@@ -519,3 +519,22 @@ def but_the_fraction_is_NaN_for_unfiltered_count_zero(self):
...
+        _base_measure_shape.return_value = expected_shape
+        measure = _MeanMeasure(None, None)
+        assert measure.shape == expected_shape
+
+    def but_it_augments_measure_shape_when_it_has_subvars(self, _base_measure_shape):
+        base_shape = (3, 4)
+        _base_measure_shape.return_value = base_shape
+        measure = _MeanMeasure(None, None, subvariables=["0001", "0002"])
+        expected_shape = base_shape + (2,)  # augmented by n_subvars
+        assert measure.shape == expected_shape
```
ernestoarbitrio, line 536: Can you collapse this 2 tests onto 1? I mean using `@pytest.mark.parametrize` ... something like:

```python
    @pytest.mark.parametrize(
        "base_shape, subvariables, expected_shape",
        (((3, 4), None, (3, 4)), ((3, 4), ["0001", "0002"], (3, 4, 2))),
    )
    def it_knows_its_altered_shape(
        self, base_shape, subvariables, expected_shape, _base_measure_shape
    ):
        _base_measure_shape.return_value = base_shape
        measure = _MeanMeasure(None, subvariables=subvariables)
        assert measure.shape == expected_shape
```

tests/integration/test_numeric_array.py
```diff
@@ -0,0 +1,39 @@
...
+
+from __future__ import absolute_import, division, print_function, unicode_literals
+
+import numpy as np
+
+from cr.cube.cube import Cube
+
+from ..fixtures import NA
+
+NAN = np.nan
```
ernestoarbitrio, line 13: where is this COSTANT used? i don't see any usage in this file ... am I missing something?

src/cr/cube/cube.py
```diff
@@ -603,11 +611,26 @@ def raw_cube_array(self):
         response. Specifically, it includes values for missing elements, any
         MR_CAT dimensions, and any prunable rows and columns.
         """
-        array = np.array(self._flat_values).reshape(self._all_dimensions.shape)
+        array = np.array(self._flat_values).reshape(self.shape)
```
ernestoarbitrio, line 614: What do you think if we change the name `array` in something more specific? I know it's not related to you changes but considering that you touch this line, maybe this can be refactored using a meaningful name.

ernestoarbitrio (changes requested):
Couple of comments + coverage must remain the same (100%) so maybe there's something to be covered in our test suite 😄

tests/integration/test_numeric_array.py
slobodan-ilic: Good catch! It was used in the middle of the process but I ended up removing it. Fixed now.

tests/unit/test_cube.py
slobodan-ilic: done

src/cr/cube/cube.py
slobodan-ilic: Done, I've used the `raw_cube_array` (same as the name of the property) instead. I'm open to better ideas.

tests/fixtures/numeric_arrays/means-grouped-by-single-categorical.json
```diff
@@ -0,0 +1,217 @@
...
+                                "?": -8
+                            },
+                            {
+                                "?": -8
+                            }
+                        ]
+                    ],
+                    "metadata": {
+                        "derived": true,
+                        "references": {},
```
malecki, line 191: Hey, we *desperately* need the references to be passed along here. In the “movie ratings” example — we know the ratings, but now we have no idea what the movies were! Or even the variable name to know that it’s movies and not magazines! Argh!

slobodan-ilic: We have the `"subvariables"` a couple of lines below. Is that not enough?

tests/fixtures/numeric_arrays/means-grouped-by-single-categorical.json
```diff
@@ -0,0 +1,217 @@
...
+                        "type": {
+                            "class": "numeric",
+                            "integer": true,
+                            "missing_reasons": {
+                                "NaN": -8,
+                                "No Data": -1
+                            },
+                            "missing_rules": {},
+                            "subvariables": [
```
slobodan-ilic, line 200: @malecki here

src/cr/cube/cube.py
```diff
@@ -312,6 +311,12 @@ def is_weighted(self):
         """True if cube response contains weighted data."""
         return self._measures.is_weighted
 
+    @lazyproperty
+    def means_subvariables(self):
+        if not self.has_means:
```
ernestoarbitrio, line 316: Considering that we're trying to keep cr.cube well documented I'd like to see a docstring here :D

src/cr/cube/cube.py
```diff
@@ -620,6 +643,34 @@ def _flat_values(self):  # pragma: no cover
 class _MeanMeasure(_BaseMeasure):
     """Statistical mean values from a cube-response."""
 
+    def __init__(self, cube_dict, all_dimensions, subvariables=None):
+        super(_MeanMeasure, self).__init__(cube_dict, all_dimensions)
+        self._subvariables = subvariables
+
+    @lazyproperty
+    def subvariables(self):
+        return self._subvariables
```
ernestoarbitrio, line 652: docstring also here.

src/cr/cube/stripe.py
```diff
@@ -581,6 +584,8 @@ def count(self):
 
     @lazyproperty
     def mean(self):
+        if self._subvariables is not None:
+            return tuple(self._mean)
         return self._mean
```
ernestoarbitrio, line 589: docstrin here

ernestoarbitrio (changes requested):
some further minor comments

src/cr/cube/cube.py
slobodan-ilic: done

src/cr/cube/cube.py
slobodan-ilic: done

src/cr/cube/stripe.py
slobodan-ilic: done

slobodan-ilic:
Closing in favor of #238

PR #224, opened 2020-10-28
Standard error margin

slobodan-ilic:
Mostly add expectations, but also implement the feature.


src/cr/cube/cubepart.py
```diff
@@ -239,6 +239,8 @@ class _Slice(CubePartition):
     dimensions which can be crosstabbed in a slice.
     """
 
+    z_alpha = 1.959964
```
ernestoarbitrio, line 242: Is this the confidence level, or better is this the zscore value given the confidence level = 95%? If yes the name `z_alpha` seems fishy/shady to me. Can we try to find a better name?

src/cr/cube/cubepart.py
```diff
@@ -291,6 +293,13 @@ def columns_dimension_type(self):
     def columns_margin(self):
         return np.array([column.margin for column in self._matrix.columns]).T
 
+    @lazyproperty
+    def columns_moe(self):
+        """Returns the margin of error (MoE) for cell percentages
+        `moe = z(0.25) * 100 * std_error`
```
ernestoarbitrio, line 299: i don't understand the `z(0.25)` and even why are you multiplying by 100. can you give me more context?

src/cr/cube/cubepart.py
```diff
@@ -1039,7 +1050,7 @@ def scale_mean(self):
 
     @lazyproperty
     def scale_median(self):
-        """ -> np.int64, the median of scales
+        """-> np.int64, the median of scales
```
ernestoarbitrio, line 1053: can you remove the comma from the docstring? I think it's a mistake of mine, but considering that your black touches this line ... :D

src/cr/cube/cubepart.py
```diff
@@ -1207,7 +1218,7 @@ def _numeric_values(self):
 
     @lazyproperty
     def _numeric_values_mask(self):
-        """ -> np.ndarray, boolean elements for each element in rows dimension."
+        """-> np.ndarray, boolean elements for each element in rows dimension."
```
ernestoarbitrio, line 1221: also here :D

src/cr/cube/cubepart.py
```diff
@@ -764,6 +771,10 @@ def table_margin(self):
     def table_margin_unpruned(self):
         return self._matrix.table_margin_unpruned
 
+    @lazyproperty
+    def table_moe(self):
+        return self.z_alpha * 100 * self.table_std_err
+
```
ernestoarbitrio, line 777: The docstring here could be the same of the `columns_moe` but with table percentages reference:
```
"""Returns the margin of error (MoE) for table percentages"""
```

ernestoarbitrio (changes requested):
Left some comments ...

src/cr/cube/cubepart.py
slobodan-ilic: Fixed all, and resolved conversations

ernestoarbitrio (approved):
Code LGTM to me :)

src/cr/cube/cubepart.py
```diff
@@ -291,16 +295,23 @@ def columns_dimension_type(self):
     def columns_margin(self):
         return np.array([column.margin for column in self._matrix.columns]).T
 
+    @lazyproperty
+    def columns_moe(self):
+        """Returns the margin of error (MoE) for col percentages
+        `moe = Z_975 * 100 * std_error` (the * 100 part accounts for percentages)
+        """
```
scanny, line 302: Rename to `column_percentages_moe` so it is explicit and also appears immediately following `column_percentages`, both here in the code and in the documentation.

Prefer `"float margin of error (MoE) for column percentages."` for the summary line. A property, like an attribute, doesn't "return" anything, it "is" some value. Although there are many examples (like the next property below) left to clean up, we reserve the `"Returns ...` prefix for methods and functions.

Blank line before continuation when the docstring extends beyond the first line: https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings

`moe = Z_975 * 100 * std_error` just restates the implementation. This docstring appears in the documentation and so should communicate to the user what they are getting from (the contract for) this property, _in user-space terms_. It should also probably state that 3.5% is represented as 3.5 (and not .035). Also it should probably mention that the return value can be NaN and under what circumstances. And we should state clearly whether a returned NaN is a `float("NaN")` or `np.nan` because those two have different behaviors in certain comparisons.

The `(the * 100 ...)` bit can appear as a comment and seems useful. Comments speak in _developer-space_ and topics related to the implementation are appropriate there.

src/cr/cube/cubepart.py
```diff
@@ -239,6 +239,10 @@ class _Slice(CubePartition):
     dimensions which can be crosstabbed in a slice.
     """
 
+    # quantile of the normal cdf at .975 because the confidence
```
scanny, line 242: Don't make me look up what `cdf` is. Spell it out and then show the abbreviation in parentheses, like `Common Denominator Frombus (CDF)`. Also, if you would, prefix comments with `--- ` to make them stand out from commented-out code. Like:

```python
# --- Quantile of the normal Cubic Dimension Fromulus (CDF) at .975 because the
# --- confidence interval is ±.025 on each side.
```

What are the units of the ±.025 value? Is that standard-deviations? Best to be explicit. Use these opportunities to educate the uninitiated reader.

src/cr/cube/cubepart.py
```diff
@@ -291,16 +295,23 @@ def columns_dimension_type(self):
...
+    def columns_moe(self):
+        """Returns the margin of error (MoE) for col percentages
+        `moe = Z_975 * 100 * std_error` (the * 100 part accounts for percentages)
+        """
+        return self.Z_975 * 100 * self.columns_std_err
+
     @lazyproperty
     def columns_std_dev(self):
-        """Returns the standard deviation for cell percentages
+        """Returns the standard deviation for col percentages
```
scanny, line 307: Good catch, remainder of fixup would be something like:

```python
"""np.float64 standard-deviation of column percentages."""
```

src/cr/cube/cubepart.py
```diff
@@ -337,17 +348,15 @@ def inserted_row_idxs(self):
...
-        """
+              0     1     2     3      4       5       6
+        0   inf   inf   inf   inf    inf    -2.9     inf
+        1   inf   inf   inf   inf    inf    -4.3     inf
+        2   2.5   1.3   3.3  -0.70  -7.25   -6.52    2.25
+        3   inf   inf   inf   inf    inf    -2.51    inf
+        4  -1.16  2.20  5.84  1.78  -8.48   -5.92    0.93
+        5   inf   inf   inf   inf    inf     9.70    inf
+
+        Only the insertions residuals are showed in a inf masked array"""
```
scanny, line 359: 👍  this scattered alignment with tabs in it has been bugging me for months :)

Final `"""` should appear on its own line. From the PEP257 link above:

> Unless the entire docstring fits on a line, place the closing quotes on a line by themselves.

src/cr/cube/cubepart.py
```diff
@@ -601,7 +610,7 @@ def scale_means_rows_margin(self):
 
     @lazyproperty
     def scale_median_column(self):
-        """ -> np.int64 ndarray of the columns scale median
+        """-> np.int64 ndarray of the columns scale median
```
scanny, line 613: We tried out this ` -> ` shorthand for `"Returns ..."` because it was compact and consistent with Python3 type-hints. But the new version of `Black` won't allow the leading space and then the eye struggles to differentiate it from the `"""`. So we're back to using `Returns` when applicable (methods and functions). This is a property though, so the arrow should just go away, leaving `"""np.int64 ndarray of ..."""`.

src/cr/cube/cubepart.py
```diff
@@ -764,6 +773,13 @@ def table_margin(self):
     def table_margin_unpruned(self):
         return self._matrix.table_margin_unpruned
 
+    @lazyproperty
+    def table_moe(self):
+        """Returns the margin of error (MoE) for table percentages
+        `moe = Z_975 * 100 * std_error` (the * 100 part accounts for percentages)
+        """
+        return self.Z_975 * 100 * self.table_std_err
```
scanny, line 781: same remarks as on `column_moe`.

tests/integration/test_headers_and_subtotals.py
```diff
@@ -1127,6 +1127,65 @@ def it_calculate_col_residuals_for_subtotals(self):
                 ],
             ],
         )
+        np.testing.assert_almost_equal(
```
scanny, line 1130: this expectation should be extracted. It's long enough to disrupt reading the test.

tests/integration/test_multiple_response.py
```diff
@@ -375,11 +375,58 @@ def test_various_measures_from_r_rows_margin():
         ],
     ]
 
+    expected_col_moe = [
```
scanny, line 378: all of the expectations for this test should be extracted. Together they are like three screenfuls of noise in the test. We're doing this incrementally as we encounter legacy tests like this one.

scanny (changes requested):
A few changes, mostly just updates on the docstring and test conventions we've evolved to while you were away :)

PR #225, opened 2020-10-29
Base measure mapping [#175213915]

ernestoarbitrio:
The authoritative keywords for keyword-measures are the ones we are already using in `zoom` codebase as defined in `cr.exporter.enums.MEASURE`. 

This PR contains a mapping of those measure-keywords to `_Slice` `_Strand` property-names for the measure allowed in the `SingleSidedMovingAvg` object.

For furhter refernece: https://github.com/Crunch-io/zoom/pull/2861, basically this is a reversed mapping of this one:
image


src/cr/cube/noa/smoothing.py
```diff
@@ -146,4 +154,4 @@ def _smoothed_values(self):
     @lazyproperty
     def _window(self):
         """int, the window parameter specified in the measure expression"""
-        return self._measure_expression.get("window", 3)
+        return self._measure_expression.get("window", 2)
```
slobodan-ilic, line 157: It seems at first glance that you don't have a test that tests a default value (if it's not provided in the expression).

tests/integration/test_smoothing.py
```diff
@@ -12,6 +12,39 @@
...
+    )
+    def it_provides_smoothed_col_index_for_compatible_cubes(
+        self, fixture, window, expectation
+    ):
+        slice_ = Cube(fixture).partitions[0]
+        col_index = slice_.evaluate(
+            {
+                "function": "one_sided_moving_avg",
+                "base_measure": "col_index",
+                "window": window,
```
slobodan-ilic, line 41: Maybe make one test where you don't provide the `"window"` parameter?

src/cr/cube/noa/smoothing.py
```diff
@@ -60,8 +60,16 @@ def values(self):
 
     @lazyproperty
     def _base_measure(self):
-        """str, the base_measure parameter specified in the measure expression"""
-        return self._measure_expression.get("base_measure", "column_percentages")
+        """str base_measure parameter specified in the measure expression"""
+        base_measure = self._measure_expression.get("base_measure", "col_percent")
```
slobodan-ilic, line 64: I also don't see a test exercising the default value here. Maybe add one?

slobodan-ilic (approved):

PR #228, opened 2020-11-06
fix: capture UserWarning in tests that trigger it

scanny:
The smoothing work generated a large number of UserWarning messages from tests that were not captured and therefore appeared as noise in the pytest output.

Capture all of these and inspect them in each test to make sure the right one is being triggered.

Correct several tests that were improperly emitting warnings or were emitting the wrong warning for the test case.

Remove unnecessary fixtures for smoothing-failure cases because those can simply be checked against the unsmoothed output.


scanny:
@ernestoarbitrio can you review this PR and merge it to master once approved? No release, just want to base the remaining work I'm doing on this so want it on master.

ernestoarbitrio (approved):

PR #233, opened 2020-11-19
Sort by value 174027230

scanny:
De-vectorize matrix.py and add sort-by-value.


tests/unit/test_matrix.py
```diff
@@ -798,3 +798,28 @@ def it_knows_its_unweighted_counts(self):
...
+                ],
+                [  # -- row 1 ------------
+                    [4, 3],  # -- col 0 --
+                    [5, 2],  # -- col 1 --
+                    [6, 1],  # -- col 2 --
+                    # --------------------
+                ],
+            ]
+        )
+        np.testing.assert_equal(
```
slobodan-ilic, line 2527: Is there a way that we could progress to using the `assert` syntax instead of the `np.testisng.assert`? Maybe with using `tolist()` if necessary. But it would certainly be more along the lines of where we're going with Crunch. And more readable

src/cr/cube/matrix.py
```diff
@@ -95,7 +95,31 @@ def table_base(self):
...
+        # --- CAT_X_MR slice produces a 1D table-base (each column has its own N) ---
+        if self._columns_dimension.dimension_type == DT.MR_SUBVAR:
+            return base_table_base[self._column_order]
+
+        # --- MR_X_CAT slice produces a 1D table-base (each row has its own N) ---
+        if self._rows_dimension.dimension_type == DT.MR_SUBVAR:
+            return base_table_base[self._row_order]
+
+        # --- CAT_X_CAT produces scalar table-base ---
+        table_base = self._cube_result_matrix.table_base
```
slobodan-ilic, line 241: we could return in a single expression, no need for var initialization

src/cr/cube/cubepart.py
```diff
@@ -505,7 +511,43 @@ def row_percentages(self):
...
+        (aka. rows_margin), a number between 0.0 and 1.0.
+        """
+        rows_margin = self.rows_margin
+        with np.errstate(divide="ignore", invalid="ignore"):
+            # --- numpy broadcasting only works "vertically", so in the 1D margin case,
+            # --- we need to "rotate" (transpose) the counts such that rows are running
+            # --- "vertically", and then re-transpose the result to get back
+            # --- (rows, cols) orientation.
+            if rows_margin.ndim == 1:
+                return (self.counts.T / self.rows_margin).T
```
slobodan-ilic, line 507: use just `rows_margin`, not the `self.rows_margin`, if we instantiated it already.

src/cr/cube/matrix.py
```diff
@@ -447,9 +447,11 @@ def blocks(cls, cube_result_matrix, measure_propname=None):
...
     def _base_values(self):
         """2D ndarray of "body" values from cube-result matrix."""
-        raise NotImplementedError(
-            "`%s` must implement `._base_values`" % type(self).__name__
-        )
+        if self._measure_propname is None:
+            raise NotImplementedError(
+                "`%s` must implement `._base_values`" % type(self).__name__
+            )
+        return getattr(self._cube_result_matrix, self._measure_propname)
```
slobodan-ilic: I'm still reading this commit by commit, so this might not be relevant, but:
Could we implement something more explicit here (rather than `getattr`)? I'm thinking something like the `__get__` method on the target class, so we end up with:
```
return self._cube_result_matrix[self._measure_propname]
```
and let the `__get__` mechanism deal with the explicitness, the dictionary of existing measures, handling the proper exceptions, etc.

src/cr/cube/matrix.py
```diff
@@ -279,7 +279,16 @@ def zscores(self):
         deviations above (positive) or below (negative) the population mean a cell's
         value is.
         """
-        raise NotImplementedError
+        # --- punt except on CAT_X_CAT (if any MR dimensions it gets np.nans) ---
```
slobodan-ilic, line 358: What happens with `CA_SUBVAR`?

src/cr/cube/cubepart.py
```diff
@@ -844,7 +844,20 @@ def table_base(self):
...
+        does not affect a scalar value but when the return value is an ndarray, the
+        shape may be different than the array returned by `.table_base`.
+
+        A multiple-response (MR) dimension produces an array of table-base values
+        because each element (subvariable) of the dimension represents a logically
+        distinct question which may not have been asked of all respondents. When both
+        dimensions are MR, the return value is a 2D ndarray. A CAT_X_CAT slice produces
+        a scalar value for this property.
+        """
+        return self._assembler.table_base_unpruned
```
slobodan-ilic, line 825: This pattern keeps repeating itself, where we only give back a single prop from the `_assembler`. It feels a little fishy, and as if it's supposed to go away. Do you have something like that envisioned in the future iterations?

slobodan-ilic (approved):
Couple of small comments but 💯 overall

src/cr/cube/matrix.py
```diff
@@ -2,157 +2,318 @@
...
+        return self._assemble_vector(
+            self._cube_result_matrix.columns_margin,
+            self._column_subtotals,
+            self._column_order,
+        )
+
+    @lazyproperty
+    def inserted_column_idxs(self):
+        """tuple of int index of each subtotal column in slice."""
+        return tuple(i for i, col_idx in enumerate(self._column_order) if col_idx < 0)
```
ernestoarbitrio, line 117: I'd like to see a comment here to explain why you're considering only the negative col_idx

src/cr/cube/matrix.py
```diff
@@ -2,157 +2,318 @@
...
+
+    @lazyproperty
+    def inserted_row_idxs(self):
+        """tuple of int index of each subtotal row in slice."""
+        return tuple(i for i, row_idx in enumerate(self._row_order) if row_idx < 0)
```
ernestoarbitrio, line 122: maybe same here

src/cr/cube/matrix.py
```diff
@@ -163,2555 +324,1553 @@ def table_margin_unpruned(self):
...
+        display_order = (
+            ExplicitOrderCollator.display_order(dimension, empty_idxs)
+            if collation_method == CM.EXPLICIT_ORDER
+            else PayloadOrderCollator.display_order(dimension, empty_idxs)
         )
 
+        # --- Returning as np.array suits its intended purpose, which is to participate
+        # --- in an np._ix() call. It works fine as a sequence too for any alternate
+        # --- use. Specifying int type prevents failure when there are zero elements.
+        return np.array(display_order, dtype=int)
```
ernestoarbitrio, line 466: what kind of problem you could have with an empty array? I mean if display_order is [] or () you shouldn'y have problems with float64 by default

tests/unit/test_matrix.py
scanny: Yeah, that's a good idea. I wasn't sure at first how best to assert these but I like the `.tolist()` method. It gets the values, shape right and usually works on the type in practice.

I've updated the unit tests in the matrix subpackage. Not all could be changed because `.tolist()` doesn't give you "almost_equal", so we'll need to figure something else out for that. Also it doesn't work on `NaN` results because `np.nan` is not equivalent to `np.nan` (and also not not-equal, interestingly). So we'll have to noodle what to do with those.

src/cr/cube/matrix.py
scanny: done

src/cr/cube/cubepart.py
scanny: yep, good catch

src/cr/cube/matrix.py
scanny: yeah, this went away in the last commit or so. I was pretty sure it would, so I didn't sweat it while I was getting the thing roughed out. Basically, the caller gets the "base_measure" and sends it in, rather than delegating that to the subtotaler. That turns out to be necessary in measure consolidation, so this unpleasantness is gone now :)

src/cr/cube/matrix.py
scanny: We don't distinguish CA_SUBVAR from CAT for matrix purposes. Do we need to in this case? If so, it wasn't in the existing code and wasn't covered by the existing tests.

src/cr/cube/cubepart.py
scanny: No, actually. This is the intended pattern for measures and marginals (and table totals and labels). All that work is completely delegated to the `Assembler` object (which probably needs a slightly different name, since "assembling" describes _how_ it does it, not what it provides). Note that not all _Slice methods are measures and marginals, there are quite a few dimension characteristics for example, and something like `.column_percentages` probably doesn't need to come from assembler. So this pattern is evident on _many_ `_Slice` properties, but not all of them. The role of `_Slice` is to operate as the interface object, integrating the efforts of several collaborators to provide a unified public interface to the caller. We could not just collapse `Assembler` into `_Slice`, for example.

One reason all the measure and marginal computation needs to be in (and below) `Assembler` is that it needs to be accessible to internal objects, like for smoothing, sort-by-value, pairwise comparison, etc. Previously some of those extended the `_Slice` public interface for private purposes, which is not a good thing. The only one accessing `_Slice` properties should be the client.

This pattern will actually become more prevalent with the measure consolidation work.

src/cr/cube/matrix.py
scanny: done

src/cr/cube/matrix.py
scanny: done

src/cr/cube/matrix.py
scanny: This array is used for indexing (in an `np.ix_()` call) so its values need to be integers (floats can't be used for indexing). The default `numpy` type is `float`, so if you don't specify a `dtype` and `display_order` is an empty sequence, numpy has no examples from which to infer type, and you end up with an array with `dtype == float`, which raises an error (or used to at least) in the later `np.ix_()` call.

PR #234, opened 2020-11-30
CI rebasing branch for sort by value

slobodan-ilic:
This PR represents the attempt to fixate the functionality developed by @scanny , by extensive use of unit tests, and the practice of continuous rebasing.


tests/unit/test_matrix.py
```diff
@@ -76,3 +76,39 @@ def dimension_(self, request):
     @pytest.fixture
     def _SumSubtotals_(self, request):
         return class_mock(request, "cr.cube.matrix._SumSubtotals")
+
+
+class Describe_BaseCubeResultMatrix(object):
+    @pytest.mark.parametrize(
+        ("is_mr_aug", "has_means", "factory_method"),
```
slobodan-ilic, line 123: `factory_method` -> `factory_method_name`

tests/unit/test_matrix.py
```diff
@@ -76,3 +76,39 @@ def dimension_(self, request):
...
+    )
+    def it_calls_the_correct_factory_method_for_appropriate_matrix_type(
+        self, request, cube_, is_mr_aug, has_means, factory_method
+    ):
+        # --- set cube properties that dictate logic
+        cube_.is_mr_aug = is_mr_aug
+        cube_.has_means = has_means
+
+        # --- create factory method mock
+        factory = method_mock(request, _BaseCubeResultMatrix, factory_method)
```
slobodan-ilic, line 143: `factory` -> `factory_method`

tests/unit/test_matrix.py
```diff
@@ -76,3 +76,39 @@ def dimension_(self, request):
...
+        factory.return_value = "zoom"
```
slobodan-ilic, line 144: Use `_BaseCubeResultMatrix` object instead, as an instance mock (name it `cube_result_matrix_`).

tests/unit/test_matrix.py
```diff
@@ -76,3 +76,39 @@ def dimension_(self, request):
...
+        ("is_mr_aug", "has_means", "factory_method"),
+        (
+            (True, None, "_mr_aug_matrix_factory"),
+            (True, True, "_mr_aug_matrix_factory"),
+            (True, False, "_mr_aug_matrix_factory"),
+            (False, True, "_means_matrix_factory"),
+            (None, True, "_means_matrix_factory"),
+            (False, False, "_regular_matrix_factory"),
+            (None, False, "_regular_matrix_factory"),
+            (False, None, "_regular_matrix_factory"),
```
slobodan-ilic, line 132: Reduce to truth table (2 x 2).

tests/unit/test_matrix.py
```diff
@@ -76,3 +76,39 @@ def dimension_(self, request):
...
+        # --- set cube properties that dictate logic
+        cube_.is_mr_aug = is_mr_aug
+        cube_.has_means = has_means
+
+        # --- create factory method mock
+        factory = method_mock(request, _BaseCubeResultMatrix, factory_method)
+        factory.return_value = "zoom"
+
+        # --- assert correct method is executed and its result returned
+        assert _BaseCubeResultMatrix.factory(cube_, "foo", "bar") == "zoom"
```
slobodan-ilic, line 147: separate out the call

slobodan-ilic, line 147: Mock the appropriate objects for calling the `factory` method (like `(dimension_, dimension_)`.

slobodan-ilic: have a call on a separate line

tests/unit/test_matrix.py
```diff
@@ -107,6 +108,50 @@ def it_calls_the_correct_factory_method_for_appropriate_matrix_type(
...
+            (
+                3,
+                DT.MR,
+                0,
+                ([["cnt"]], [["unw"]], [["w_miss"]]),
+                ("cnt", "unw", "w_miss"),
+            ),
+        ),
+    )
+    def it_knows_how_to_extract_counts_for_matrix_creation(
```
slobodan-ilic, line 182: Create 3 different tests for each execution path.

slobodan-ilic: For slicing examples, use actual `np` to check if the slicing is correct.

src/cr/cube/matrix.py
```diff
@@ -159,4 +164,46 @@ def _mr_aug_matrix_factory(cls, cube, dimensions, slice_idx):
...
-        raise NotImplementedError
+        (
+            counts,
+            unweighted_counts,
+            counts_with_missings,
+        ) = cls._extract_counts_for_matrix_creation(cube, slice_idx)
+
+        dimension_types = cube.dimension_types[-2:]
+
+        if dimension_types == (DT.MR, DT.MR):
```
slobodan-ilic: Consider stacked `if else`

```
MatrixCls = (
    _MrMrMatrix
    if dimension_types == (DT.MR, DT.MR) else
    _MrXCatMat...
```

tests/unit/test_matrix.py
```diff
@@ -152,8 +159,50 @@ def it_knows_how_to_extract_counts_for_matrix_creation(
...
+            ((DT.MR, DT.CAT, DT.CAT), _CatXCatMatrix, _BaseCubeResultMatrix),
+            ((DT.MR, DT.MR), _MrXMrMatrix, _CatXCatMatrix),
+            ((DT.MR, DT.CAT), _MrXCatMatrix, _CatXCatMatrix),
+            ((DT.CAT, DT.MR, DT.CAT), _MrXCatMatrix, _CatXCatMatrix),
+            ((DT.CAT, DT.MR), _CatXMrMatrix, _CatXCatMatrix),
+            ((DT.CAT, DT.CAT, DT.MR), _CatXMrMatrix, _CatXCatMatrix),
+            ((DT.MR, DT.CAT, DT.MR), _CatXMrMatrix, _CatXCatMatrix),
+        ),
+    )
+    def it_can_create_regular_matrix(
```
slobodan-ilic, line 207: Change this to make more sense according to our discussion with @scanny , and according to the changes in the comments.

tests/unit/test_matrix.py
```diff
@@ -229,3 +229,16 @@ def and_its_blocks_know_their_internal_representation(self, request):
         blocks = _BaseSubtotals.blocks(fake_cube, fake_measure)
         _blocks.assert_called_once()
         assert blocks == fake_blocks
+
+    def it_knows_how_to_assemble_its_blocks(self, request):
+        bs = _BaseSubtotals(None, None)
```
scanny, line 276: Even if it ends up causing line wrapping, I generally find it most readable to derive an instance name directly from the classname, like `base_subtotals` as an instance of `_BaseSubtotals` in this case. Since this is a base-class, `subtotals` might be apt, like "an instance of any subclass of `_BaseSubtotals`.

scanny: Also, I'm not sure I remember exactly why, but I've found it best to put the instance construction as the last line of the fixturing, just above the actual call. I suppose it's because often there are other fixture components that are used as parameters in the call and it's nice to always know where to look for that line which appears in almost every unit test.

tests/unit/test_matrix.py
```diff
@@ -229,3 +229,16 @@ def and_its_blocks_know_their_internal_representation(self, request):
...
+        bs._base_values = property_mock(request, _BaseSubtotals, "_base_values")
```
scanny, line 277: You don't want to be assigning mocks to properties of the instance. The mocks already "do" that. I'm not sure why this is working :)

So you would have something like this instead. Note there is no need to name mocks if you don't refer to them later; flake8 will complain if you do :)

```
property_mock(request, _BaseSubtotals, "_base_values", return_value=[1])
property_mock(request, _BaseSubtotals, "_subtotal_columns", return_value=[2])
property_mock(request, _BaseSubtotals, "_subtotal_rows", return_value=[3])
property_mock(request, _BaseSubtotals, "_intersections", return_value=[4])

assert subtotals._blocks == [[[1], [2]], [[3], [4]]]
```

tests/unit/test_matrix.py
```diff
@@ -242,3 +243,16 @@ def it_knows_how_to_assemble_its_blocks(self, request):
             [bs._base_values, bs._subtotal_columns],
             [bs._subtotal_rows, bs._intersections],
         ]
+
+
+class Describe_SumSubtotals(object):
+    @pytest.mark.parametrize(("prop_name",), (("foo",), ("bar",)))
```
scanny, line 430: These propnames would be measure names, so "weighted_counts" and "unweighted_counts" might be better choices than "foo" and "bar".

tests/unit/test_matrix.py
```diff
@@ -242,3 +243,16 @@ def it_knows_how_to_assemble_its_blocks(self, request):
...
+    def it_returns_correct_base_values(self, _cube_result_matrix_prop_, prop_name):
+        ss = _SumSubtotals(_cube_result_matrix_prop_, prop_name)
```
scanny, line 432: _SumSubtotals is called with a cube-result-matrix _instance_, not with a property.

```python
def it_knows_the_base_values_to_help(self, cube_result_matrix_, prop_name):
    setattr(cube_result_matrix, prop_name, [1, 2, 3])
    assert _SumSubtotals(cube_result_matrix, prop_name)._base_values == [1, 2, 3]
```

So in fact, in order for this to work, the prop-names will need to be real ones defined (but not necessarily implemented yet) on `_BaseCubeResultMatrix`.

tests/unit/test_matrix.py
```diff
@@ -242,3 +243,16 @@ def it_knows_how_to_assemble_its_blocks(self, request):
...
+        assert ss._base_values == getattr(_cube_result_matrix_prop_, prop_name)
+
+    # fixture components ---------------------------------------------
+
+    @pytest.fixture
+    def _cube_result_matrix_prop_(self, request):
+        return property_mock(request, Assembler, "_cube_result_matrix")
```
scanny, line 487: and this will need to become an instance mock. There is no `._cube_result_matrix` property on `_SumSubtotals` I don't think. It's an instance variable isn't it?

tests/unit/test_matrix.py
```diff
@@ -256,3 +256,9 @@ def it_returns_correct_base_values(self, _cube_result_matrix_prop_, prop_name):
     @pytest.fixture
     def _cube_result_matrix_prop_(self, request):
         return property_mock(request, Assembler, "_cube_result_matrix")
+
+
+class Describe_CatXCatMatrix(object):
+    def it_provides_its_unweighted_counts_to_help(self):
```
scanny, line 508: This one doesn't get "_to_help" because the `.unweighted_counts` property is interface, not a helper.

tests/unit/test_matrix.py
```diff
@@ -256,3 +256,9 @@ def it_returns_correct_base_values(self, _cube_result_matrix_prop_, prop_name):
...
+        fake_counts = Mock()
+        assert _CatXCatMatrix(None, None, fake_counts).unweighted_counts == fake_counts
```
scanny, line 510: I would use a literal for the counts, like [1, 2, 3] or something. They could just be inlined here I think, no need for a name, although you could do:

```python
assert _CatXCatMatrix(None, None, unweighted_counts=[1, 2, 3]).unweighted_counts == [1, 2, 3]
```

src/cr/cube/matrix.py
```diff
@@ -91,8 +91,32 @@ def blocks(cls, cube_result_matrix, measure_propname=None):
...
+
+    @lazyproperty
+    def _subtotal_columns(self):
+        """Subtotal columns must be implemented by each subclass."""
+        raise NotImplementedError
+
+    @lazyproperty
+    def _subtotal_rows(self):
+        """Subtotal rows must be implemented by each subclass."""
         raise NotImplementedError
```
scanny: If these will have no eventual implementation, you should probably complete the (non-)implementation now since there won't be a trigger to do it later and "interface-defining" base-class methods like this are not usually tested directly. The best implementation I've found so far is this:

```python
@lazyproperty
def _subtotal_columns(self):
    """ ... """
    raise NotImplementedError(
        "`%s` must implement `._subtotal_columns` % type(self).__name__
    )
```

That way you get an error like "`_SumSubtotals` must implement `._subtotal_columns`" at just the right time (when it's time to implement that) and there's no digging around to work out where it needs implementing. Using `type(self).__name__` doesn't work with old-style Python2 classes (those that don't inherit from `object`) but we shouldn't be encountering any of those. Those would need `self.__class__.__name__` instead.

src/cr/cube/matrix.py
```diff
@@ -128,6 +159,10 @@ def _base_values(self):
         """2D np.float64 ndarray of table-stderr for each cell of cube-result matrix."""
         return getattr(self._cube_result_matrix, self._measure_propname)
 
+    def _subtotal_column(self, subtotal):
+        """Return (n_rows,) ndarray of np.nan values."""
+        return np.sum(self._base_values[:, subtotal.addend_idxs], axis=1)
```
scanny, line 739: You don't really want to be implementing two methods in the same commit. Sometimes you write multiple tests (when the method has two or more very different execution paths, like when it has an exception path), but generally it's one method/property per commit. So you'd be mocking `._subtotal_column` and it's return value(s) in this one.

tests/unit/test_matrix.py
```diff
@@ -244,19 +245,83 @@ def it_knows_how_to_assemble_its_blocks(self, request):
...
+
+    # fixture components ---------------------------------------------
+
+    @pytest.fixture
+    def _subtotal_column_(self, request):
+        return method_mock(
+            request,
+            _BaseSubtotals,
+            "_subtotal_column",
+            return_value=np.array([1, 2, 3]),
```
scanny, line 400: It's generally inadvisable to place return values on fixture components. That limits their reuse (and they shouldn't exist unless they're being used more than once, inlined in the test otherwise). Also it separates the return value from the test requiring you to scroll up and down to see what should be coming back. Generally just make a simple mock here and assign a return value in each test that uses it.

tests/unit/test_matrix.py
```diff
@@ -244,19 +245,83 @@ def it_knows_how_to_assemble_its_blocks(self, request):
...
+    def it_knows_its_subtotal_columns(
+        self,
+        _column_subtotals_prop_,
+        _nrows_prop_,
+        _subtotal_column_,
+        nrows,
+        column_subtotals,
+        expected,
+    ):
+        _column_subtotals_prop_.return_value = column_subtotals
```
scanny, line 304: The column-subtotals returned by `._column_subtotals` is a `cr.cube.dimension._Subtotals` object. It's probably okay to mock those with a tuple of `_Subtotal` instance mocks, but I don't think a list is going to get it done.

scanny: This one might be good to use two tests for, at least to get started. You can always combine them later if they're similar enough, but starting with the empty case is probably easiest and will allow you to make progress. These can be something like:

```
it_produces_an_empty_column_subtotals_matrix_when_there_are_no_column_subtotals()

# --- and ---

it _produces_an_nrows_x_n_col_subtotals_matrix_when_there_are_column_subtotals()
```

One advantage of that approach is you don't have to squeeze your fixtures into a compact set of parameters, so the fixture is less abstract and easier to read.

scanny (commented):
Here's a few comments to help things along :)

tests/unit/test_matrix.py
```diff
@@ -0,0 +1,51 @@
...
+    def it_knows_the_unweighted_counts(
+        self,
+        _cube_result_matrix_prop_,
+        cube_result_matrix_,
+        _SumSubtotals_,
+        _assemble_matrix_,
+    ):
+        _cube_result_matrix_prop_.return_value = cube_result_matrix_
+        _SumSubtotals_.blocks.return_value = [["A", "B"], ["C", "D"]]
+        _assemble_matrix_.return_value = [["un"], ["weighted"], ["counts"]]
```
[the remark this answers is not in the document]
slobodan-ilic: this was already done (probably you)

tests/unit/test_matrix.py
```diff
@@ -32,12 +34,33 @@ def it_knows_the_unweighted_counts(
...
+        _BaseCubeResultMatrix_.factory.return_value = cube_result_matrix_
+        slice_idx = 42
+        assembler = Assembler(cube_, (dimension_, dimension_), slice_idx)
+
+        cube_result_matrix = assembler._cube_result_matrix
+
+        _BaseCubeResultMatrix_.factory.assert_called_once_with(
+            cube_, (dimension_, dimension_), slice_idx
+        )
+        assert cube_result_matrix == cube_result_matrix_
```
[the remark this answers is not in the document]
slobodan-ilic: this is also already done

tests/unit/test_matrix.py
```diff
@@ -206,3 +208,24 @@ def extract_counts_(self, request):
...
         )
+
+
+class Describe_BaseSubtotals(object):
+    def it_initializes_properly(self, request):
+        fake_cube = Mock()
+        fake_measure = Mock()
+        base_subtotals = _BaseSubtotals(fake_cube, fake_measure)
+        assert base_subtotals._cube_result_matrix == fake_cube
+        assert base_subtotals._measure_propname == fake_measure
```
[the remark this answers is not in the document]
slobodan-ilic: done

tests/unit/test_matrix.py
```diff
@@ -206,3 +208,24 @@ def extract_counts_(self, request):
...
+
+    def and_its_blocks_know_their_internal_representation(self, request):
+        fake_blocks = Mock()
+        fake_cube = Mock()
+        fake_measure = Mock()
```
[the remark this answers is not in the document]
slobodan-ilic: Got it. I think this is now removed. Notice for future taken.

tests/unit/test_matrix.py
slobodan-ilic: done

tests/unit/test_matrix.py
slobodan-ilic: yes, figured it out just before reading the comment... this was a first day blurp. fixed now...

tests/unit/test_matrix.py
slobodan-ilic: sure. but this already got fixed

tests/unit/test_matrix.py
slobodan-ilic: sure, but this probably got fixed by you :)

slobodan-ilic:
Closing in favor of #233

src/cr/cube/matrix.py
slobodan-ilic: Great how this got resolved by itself when I addressed your previous comments. It's inspiring!

tests/unit/test_matrix.py
slobodan-ilic: this also got fixed. notion taken

tests/unit/test_matrix.py
slobodan-ilic: i think i figured it out ok now..

PR #235, opened 2020-11-30
Selected categories label property available at cubepart level. 

ernestoarbitrio:


malecki:
For other reviewers: 
At first I thought this looked large, but I would say this is the minimum appropriate size for the dimensions needed. In the full form, it’s 5 × 4 × 3 × 2, and in reduced selected form, 4 × 3 × 3

src/cr/cube/cubepart.py
```diff
@@ -122,6 +122,11 @@ def population_fraction(self):
         """population fraction of the cube"""
         return self._cube.population_fraction
 
+    @lazyproperty
+    def selected_categories_label(self):
+        """Set of selected categories for all the dimensions."""
+        return {}
```
slobodan-ilic, line 128: I think this property shouldn't be here. I'm not exactly sure which class it's in (looking at diff only), but my guess is that that class shouldn't have that property. Only the class that can actually return something useful should have it (and then the legit return value would be `None`). And then we'd inspect it from the exporter, if it's exactly that combination of dimensions (or whatever the condition is) and _know_ that we need to print it out.

tests/unit/test_cubepart.py
```diff
@@ -225,6 +225,10 @@ def it_provides_the_transforms_dict_to_help(self, transforms, expected_value):
         """Handles defaulting of transforms arg."""
         assert CubePartition(None, transforms)._transforms_dict == expected_value
 
+    def it_knows_its_selected_categories_labels(self):
+        # --- default of {} is overridden by subclasses when appropriate ---
```
slobodan-ilic, line 229: Yeah, this makes no sense to me at first sight. We should _only_ have it in derived classes. Higher classes don't know about that case - it's too specific.

src/cr/cube/cubepart.py
scanny: Also, `{}` is an empty `dict`, not a set. An empty set is `set()`, although a `lazyproperty` should only ever be returning an (immutable) `frosenset()`.

src/cr/cube/cubepart.py
```diff
@@ -754,6 +759,16 @@ def scale_std_err_row(self):
...
 
+    @lazyproperty
+    def selected_categories_label(self):
+        """Set of selected categories for all the dimensions."""
+        labels = []
+        for d in self._dimensions:
+            for s in d.selected_categories:
+                if s.get("name"):
+                    labels.append(s.get("name"))
+        return set(labels)
```
scanny, line 770: I think this gets it done without the mutation:

```python
return frozenset(
    s["name"]
    for d in self.dimensions
    for s in d.selected_categories
    if s.get("name")
)

File: src/cr/cube/dimension.py [hunk 4]
```diff
@@ -381,6 +381,16 @@ def prune(self):
...
 
+    @lazyproperty
+    def selected_categories(self):
+        """List of selected categories specified for this dimension."""
+        selected_categories = self._dimension_dict["references"].get(
+            "selected_categories"
+        )
+        if not selected_categories or selected_categories == [{}]:
+            return []
+        return selected_categories
```
scanny, line 392: In general, a lazyproperty should not return a mutable object. Prefer tuple for sequence return types.

src/cr/cube/dimension.py
```diff
@@ -381,6 +381,16 @@ def prune(self):
             return True
         return False
 
+    @lazyproperty
+    def selected_categories(self):
+        """List of selected categories specified for this dimension."""
+        selected_categories = self._dimension_dict["references"].get(
+            "selected_categories"
+        )
+        if not selected_categories or selected_categories == [{}]:
```
scanny, line 390: Why is `[{}]` a legitimate value for selected_categories? If `dimension.references.selected_categories` is an array of categories, `[]` is a valid value, but `{}` is not a valid category and should be filtered out by the producer, not left to clients to deal with.

src/cr/cube/cubepart.py
scanny: Also, I'm thinking `selected_category_labels` is grammatically more correct. This is a collection and so should be plural, not a single "label".

src/cr/cube/cubepart.py
scanny: Okay, a little more on this:

1. Name should be `selected_category_names`. It is a collection, so name is plural, not singular, and plural goes on _last_ word.
2. The return type should be `tuple` such that the order of the names is preserved. This also makes the return value immutable which is a must for lazyproperties.

src/cr/cube/cubepart.py
scanny: A method with no implementation (by design, in a base class) should raise `NotImplementedError` like this:

```python
raise NotImplementedError(
    "`%s` must implement `.selected_category_labels`
    % self.__class__.__name__
)
```

This way, if this method is called on `_Slice` for example, and `_Slice` has not implemented it, you get the error message: "`_Slice` must implement `.selected_category_labels`", which is very clear about what needs to be done.

This allows you to mock the method on the base class, while protecting against that method implementation actually ever being inherited directly by a subclass.

tests/integration/test_cubepart.py
```diff
@@ -735,6 +735,11 @@ def it_knows_its_insertions(self):
             ),
         )
 
+    def it_knwos_its_selected_categories(self):
```
scanny, line 738: knows

tests/unit/test_cubepart.py
scanny: There's generally no need to test an "interface-defining" method on a base class that raises `NotImplementedError`.

src/cr/cube/cubepart.py
scanny: After chatting with Mike, we're thinking `.selected_category_labels` is a bit too general and could mean a lot of things. Better would be `.selected_ca_category_labels`, which this should be unless we can think of something better, but also the docstring should be robust, with minimally a paragraph that explains that this only applies in the "transforming a CA into MR" case and explaining how that works and when it arises in a business context.

src/cr/cube/cubepart.py
```diff
@@ -123,9 +123,12 @@ def population_fraction(self):
         return self._cube.population_fraction
 
     @lazyproperty
-    def selected_categories_label(self):
+    def selected_ca_category_labels(self):
         """Set of selected categories for all the dimensions."""
-        return {}
+        raise NotImplementedError(
+            "`%s` must implement `.selected_ca_category_labels`"
```
slobodan-ilic, line 129: I think that this message should be explicit about the fact that the _child_ classes must implement this, and not this class. So if it says that _this_ class must implement it, it's technically wrong.

scanny: @slobodan-ilic `CubePartition` is never instantiated directly (except in tests) and so it will _always_ be a subclass that triggers this message (which should only be seen by developers anyway).

@ernestoarbitrio just for future reference, I did a little research yesterday and it looks like `type(self).__name__` is preferable over `self.__class__.__name__`, although they both work fine. No need to change this one if you're not in there otherwise, but this pattern comes up from time to time so we should probably use the `type()` version going forward.

scanny: @ernestoarbitrio why are we saying this should be implemented on all `_CubePartition` subclasses? Isn't it a characteristic that is particular to a `_Slice`?

src/cr/cube/cubepart.py
scanny: @ernestoarbitrio what about elaborating the docstring to educate the unfamiliar reader (all of us after about two weeks) about what selected CA categories are and the gist of the whole "make a CA into an MR" bit?

src/cr/cube/cubepart.py
ernestoarbitrio: nope, because this property will be called on exporter without cehcking on the partition type beeing a slice a strand or a nub. So we need it in all the subclasses.

slobodan-ilic (approved):

PR #238, opened 2020-12-22
Numeric array handling

ernestoarbitrio:
Cr.cube features to handle correctly numeric array cube responses


tests/integration/test_numeric_array.py
```diff
@@ -0,0 +1,40 @@
...
+from ..fixtures import NA
+
+
+# pylint: disable=no-self-use
+class TestNumericArrays:
+    """Test-suite for numeric-arrays behaviors."""
+
+    def test_num_arr_x_cat_with_missings(self):
+        """Test means on numeric array, grouped by single categorical dimension."""
+        slice_ = Cube(NA.NUM_ARR_X_CAT_WITH_MISSINGS).partitions[0]
```
slobodan-ilic, line 20: The name of the fixture wrongly implies that we have a cross product of something (e.g. in the dimensions), when there's none. A better name would be `NUM_ARR_MEANS_GROUPED_BY_CAT`. So we know exactly what kind of cube we're dealing with.

tests/integration/test_numeric_array.py
```diff
@@ -0,0 +1,40 @@
...
+
+        np.testing.assert_almost_equal(
+            slice_.means, [[87.66666667, 93.33333333, 1.0], [52.5, 50.0, 45.0]]
+        )
+        np.testing.assert_almost_equal(slice_.column_base, [5, 4, 2])
+
+    def test_num_arr_weighted_x_cat_with_missings(self):
+        """Test means on numeric array, grouped by single categorical dimension."""
+        slice_ = Cube(NA.NUM_ARR_WEIGHTED_X_CAT_WITH_MISSINGS).partitions[0]
```
slobodan-ilic, line 29: Same here, it's not really a cross product.

tests/integration/test_numeric_array.py
```diff
@@ -0,0 +1,40 @@
...
+
+        np.testing.assert_almost_equal(
+            slice_.means, [[91.625, 94.375, 1.0], [50.71428571, 50.0, 45.0]]
+        )
+        np.testing.assert_almost_equal(slice_.column_base, [5, 4, 2])
+
+    def test_means_without_grouping_by_dimensions(self):
+        """Test means on no-dimensions measure of numeric array."""
+        strand_ = Cube(NA.NUM_ARR_MEANS_NO_DIMS).partitions[0]
```
slobodan-ilic, line 38: This name is better, but can be improved, maybe like: `NUM_ARR_MEANS_NO_GROUPING`.

tests/integration/test_numeric_array.py
```diff
@@ -0,0 +1,40 @@
...
+    """Test-suite for numeric-arrays behaviors."""
+
+    def test_num_arr_x_cat_with_missings(self):
+        """Test means on numeric array, grouped by single categorical dimension."""
+        slice_ = Cube(NA.NUM_ARR_X_CAT_WITH_MISSINGS).partitions[0]
+
+        np.testing.assert_almost_equal(
+            slice_.means, [[87.66666667, 93.33333333, 1.0], [52.5, 50.0, 45.0]]
+        )
+        np.testing.assert_almost_equal(slice_.column_base, [5, 4, 2])
```
slobodan-ilic, line 25: Just by the first look, maybe we could use a different summary here, just to be explicit? Something like `mean_column_base`. Or even `mean_unweighted_column_base`. This is potentially important, because we're going to have different measures in the cube (maybe multiple), and will potentially also need different bases for them.

tests/integration/test_numeric_array.py
```diff
@@ -0,0 +1,40 @@
...
+# pylint: disable=no-self-use
+class TestNumericArrays:
+    """Test-suite for numeric-arrays behaviors."""
+
+    def test_num_arr_x_cat_with_missings(self):
+        """Test means on numeric array, grouped by single categorical dimension."""
+        slice_ = Cube(NA.NUM_ARR_X_CAT_WITH_MISSINGS).partitions[0]
+
+        np.testing.assert_almost_equal(
+            slice_.means, [[87.66666667, 93.33333333, 1.0], [52.5, 50.0, 45.0]]
```
slobodan-ilic, line 23: When specifying the expectation with literal values like this, maybe use the same trick we used in the `cr.server` test, where you specified for each entry what it relates to. So here, it would be something like:
```
[
    #  Gender:
    #  M, F, Missing
    [87, 93, 1.0],  # Num Arr subvar: 0001
    [52, 50, 45],  # Num Arr subvar: 0002
]
```

slobodan-ilic: Same for other tests

tests/unit/test_cube.py
```diff
@@ -470,18 +471,183 @@ def and_it_accepts_a_JSON_format_cube_response(self, cube_response, expected_val
...
+
+    @pytest.mark.parametrize(
+        "fixture, expectation",
+        (
+            (CR.MEANS_CAT_X_CAT_HS, None),
+            (CR.MR_X_CAT_HS, None),
+            (NA.NUM_ARR_MEANS_NO_DIMS, ["0001", "0002"]),
+        ),
+    )
+    def it_knows_if_it_has_means_subvariales(self, fixture, expectation):
```
slobodan-ilic, line 636: This doesn't seem like a unit test. It uses fixtures and the mechanism to load them from the file.

tests/unit/test_cube.py
```diff
@@ -470,18 +471,183 @@ def and_it_accepts_a_JSON_format_cube_response(self, cube_response, expected_val
...
+        cube_response,
+        expected_value,
+    ):
         with pytest.raises(TypeError) as e:
             Cube(cube_response)._cube_dict
 
         assert str(e.value) == expected_value
 
+    @pytest.mark.parametrize(
+        "cube_response, expected_value",
```
slobodan-ilic, line 485: This is unreadable. I'd prefer if we tested this in a different way. I.e. if the same path is tested by integration tests, I'd say to get rid of this test.

If we would like to test if the dimensions are inflated properly, there's no need to replicate the entire `"measures"` objects in the response. We can just mock that with a property mock, and use a smaller piece of the cube response to test if the dimension is added. We don't even need to provide the cube with the response, but just test if the inflation is correctly executed (i.e. we could just test if a single element is being appended or inserted).

src/cr/cube/cube.py
```diff
@@ -410,16 +421,47 @@ def _ca_as_0th(self):
     @lazyproperty
     def _cube_dict(self):
         """dict containing raw cube response, parsed from JSON payload."""
+        cube_dict = copy.deepcopy(self._cube_response)
+        # ---cube inflation---
+        # ---In case of numeric arrays, we need to inflate the rows dimension
```
slobodan-ilic, line 426: It's not rows, according to the code. It's columns :) because we're appending.

src/cr/cube/cube.py
```diff
@@ -410,16 +421,47 @@ def _ca_as_0th(self):
...
+        # ---according to the mean subvariables. For each subvar the rows dimension
+        # ---will have a new element related to the subvar metadata.
+        if self._mean_subvariables:
```
slobodan-ilic, line 429: This entire block would be suited for a separate static method, because it's specific enough. Something like:
```
if self._mean_subvariables:
    dimensions = self._prepare_num_arr_dimensions(dimensions)
```
which would then have a nice docstring, and be properly placed in the code, and not obfuscate the readability of something like `_cube_dict`.

tests/integration/test_numeric_array.py
ernestoarbitrio: not sure because the `column_base` is a measure that we have for all the kind of cube.

src/cr/cube/cube.py
```diff
@@ -431,6 +473,22 @@ def _is_single_filter_col_cube(self):
         """bool determines if it is a single column filter cube."""
         return self._cube_dict["result"].get("is_single_col_cube", False)
 
+    @lazyproperty
+    def _mean_subreferences(self):
+        """List of mean subreferences, tipically for numeric arrays."""
+        cube_response = self._cube_response
+        cube_measures = cube_response.get("result", {}).get("measures", {})
+        metadata = cube_measures.get("mean", {}).get("metadata", {})
+        return metadata.get("references", {}).get("subreferences", [])
```
slobodan-ilic, line 482: In one case we have `None` for the invalid value, while here we'll have `[]`. We need to make this uniform. Both values are fine with me, although `[]` is probably better because of the uniform type.

Also, I'd maybe avoid this particular method altogether, because you're only going to use it in the inflation procedure. It doesn't _do_ anything, it "just" fetches some values if they're there...

src/cr/cube/cube.py
```diff
@@ -645,6 +717,9 @@ class _UnweightedCountMeasure(_BaseMeasure):
     @lazyproperty
     def _flat_values(self):
         """tuple of int counts before weighting."""
+        # ---If valid_count are expressed in the cube dict, returns its data---
+        if self._cube_dict["result"]["measures"].get("valid_count", {}).get("data"):
+            return tuple(self._cube_dict["result"]["measures"]["valid_count"]["data"])
```
slobodan-ilic, line 722: This is a little confusing. Our canonical way of naming things is that "weighted" is implicit, and that "unweighted" is explicit. So I'd rather change the keys altogether, and call the weighted version of the valid counts just `"valid_count"`, and the unweighted `"valid_count_unweighted"`, so that we don't confuse ourselves.

And then here, in this property, you can use the "unweighted" explicitly, which will correspond with the class name. And in the other one (maybe) you'd use just valid counts.

src/cr/cube/stripe.py
```diff
@@ -558,12 +560,13 @@ def unweighted_count(self):
 class _MeansStripeRow(_BaseStripeRow):
     """Stripe-row for a non-MR 1D cube-result with means."""
 
-    def __init__(self, element, unweighted_count, mean):
```
slobodan-ilic, line 561: Are we even using this anywhere? Doesn't seem like it. If so - please get rid of it :)

slobodan-ilic (changes requested):
Per comments. But looks good overall.

src/cr/cube/stripe.py
ernestoarbitrio: mmm i don't understand this comment :D

src/cr/cube/cube.py
```diff
@@ -517,6 +493,32 @@ def _measures(self):
         """
         return _Measures(self._cube_dict, self._all_dimensions)
 
+    @lazyproperty
+    def _numeric_array_dimensions(self):
```
slobodan-ilic, line 497: The name should be singular, as it's only a single dimension object.

src/cr/cube/cube.py
```diff
@@ -517,6 +493,32 @@ def _measures(self):
...
+        """Column dimensions object inflated according to the mean subvariables."""
+        column_dimensions = {
```
slobodan-ilic, line 499: same for this var name

tests/integration/test_numeric_array.py
```diff
@@ -15,26 +15,38 @@
...
 
         np.testing.assert_almost_equal(
-            slice_.means, [[91.625, 94.375, 1.0], [50.71428571, 50.0, 45.0]]
+            slice_.means,
+            [
+                #  Movies:
+                #  S1, S2, S3
+                [91.625, 94.375, 1.0],  # Gender: Male
+                [50.71428571, 50.0, 45.0],  # Gender: Female
+            ],
```
slobodan-ilic, line 44: 💯

tests/unit/test_cube.py
```diff
@@ -482,85 +481,33 @@ def but_it_raises_on_other_cube_response_types(
...
-                            }
-                        },
-                    }
-                },
+                ["001", "002"],
+                [{"alias": "A", "name": "A"}, {"alias": "B", "name": "B"}],
+                [
+                    {"id": 0, "value": {"references": {"alias": "A", "name": "A"}}},
+                    {"id": 1, "value": {"references": {"alias": "B", "name": "B"}}},
+                ],
```
slobodan-ilic, line 492: Much better

tests/unit/test_cube.py
```diff
@@ -648,6 +608,14 @@ def _cube_dict_prop_(self, request):
...
         return property_mock(request, Cube, "_cube_response")
 
+    @pytest.fixture
+    def _mean_subreferences_prop_(self, request):
+        return property_mock(request, Cube, "_mean_subreferences")
+
+    @pytest.fixture
+    def _mean_subvariables_prop_(self, request):
+        return property_mock(request, Cube, "_mean_subvariables")
+
```
slobodan-ilic, line 618: 💯

src/cr/cube/cube.py
```diff
@@ -422,38 +422,14 @@ def _ca_as_0th(self):
...
-                        "value": {
-                            "references": {
-                                "alias": subrefs[i].get("alias") if subrefs else None,
-                                "name": subrefs[i].get("name") if subrefs else None,
-                            }
-                        },
-                    },
-                )
-            dimensions.append(rows_dimension)
+            dimensions.append(self._numeric_array_dimensions)
```
slobodan-ilic, line 427: Since we only have a single append here, I'd prefer a more functional expression here (and not mutation). Something like:
```
dimensions = cube_dict.get(. . .) + [self._numeric_array_dimension]
```
(provided that you change the name of that property).

src/cr/cube/cube.py
```diff
@@ -312,6 +316,13 @@ def is_weighted(self):
         """True if cube response contains weighted data."""
         return self._measures.is_weighted
 
+    @lazyproperty
+    def means_subvariables(self):
+        """List of subvar names for numeric arrays, None otherwise."""
+        if not self.has_means:
+            return None
```
slobodan-ilic, line 323: This is what I was referring to. I know that it's the API method (property), and that it's not the same, but we should avoid confusing an unsuspecting reader of the code :)

src/cr/cube/cube.py
```diff
@@ -431,6 +449,22 @@ def _is_single_filter_col_cube(self):
         """bool determines if it is a single column filter cube."""
         return self._cube_dict["result"].get("is_single_col_cube", False)
 
+    @lazyproperty
+    def _mean_subreferences(self):
```
slobodan-ilic, line 453: Do we really need both subreferences and subvariables separately? I think we could ditch at least one of those properties :)

src/cr/cube/cube.py
ernestoarbitrio: actaully we can't do this cause otherwise we don't mutate the cube_dict object

src/cr/cube/cube.py
ernestoarbitrio: ok

src/cr/cube/cube.py
```diff
@@ -410,16 +414,23 @@ def _ca_as_0th(self):
     @lazyproperty
     def _cube_dict(self):
         """dict containing raw cube response, parsed from JSON payload."""
+        cube_dict = copy.deepcopy(self._cube_response)
+        if self._mean_subvariables:
+            dimensions = cube_dict.get("result", {}).get("dimensions", {})
```
slobodan-ilic, line 419: This is interesting, but we _have_ to have both the `"result"` and the `"dimensions"` in it. And if we don't, then for the result it's ok to use `{}`, but for the dimensions it has to be `[]`, because you're appending in the next line. So, either do that, or access the values directly with square brackets (and if you want to do a check if everything is valid, do so at construction time (fail sooner rather than later)).

ernestoarbitrio: u right

src/cr/cube/cube.py
```diff
@@ -410,16 +414,23 @@ def _ca_as_0th(self):
...
-            cube_dict = (
-                cube_response
-                if isinstance(cube_response, dict)
-                else json.loads(cube_response)
+            cube_response = (
+                response if isinstance(response, dict) else json.loads(response)
             )
             # ---cube is 'value' item in a shoji response---
-            return cube_dict.get("value", cube_dict)
+            return cube_response.get("value", cube_response)
```
slobodan-ilic, line 433: Yeah, some validation definitely wouldn't hurt here, but maybe best to leave it for later (because it needs to be a proper validator class, and not a bunch of ifs thrown around).

src/cr/cube/cube.py
```diff
@@ -459,6 +486,32 @@ def _measures(self):
         """
         return _Measures(self._cube_dict, self._all_dimensions)
 
+    @lazyproperty
+    def _numeric_array_dimension(self):
+        """Column dimensions object inflated according to the mean subvariables."""
```
slobodan-ilic, line 491: dimensions -> dimension

Also, the object itself is not inflated, it's used for inflating other stuff. And we need to sy that it can also return an invalid result if there's no numeric array defined. By the way, should we return a `None` in such case explicitly (like at the start of the method)? I think we should.

src/cr/cube/cube.py
```diff
@@ -459,6 +486,32 @@ def _measures(self):
...
+        column_dimension = {
+            "references": {"alias": "mean", "name": "mean"},
+            "type": {"elements": [], "class": "enum", "subtype": {"class": "variable"}},
+        }
+        subrefs = self._mean_subreferences
+        # ---cube inflation---
```
slobodan-ilic, line 497: This is not yet inflation. It happens in the original `dict` code, and this comment should probably go there. We still need some commenting here, but here, try to explain _what_ we're building, rather than _what_ we need to _do_ with it. And then explain what we need to do, in the place we're actually doing it.

src/cr/cube/cube.py
```diff
@@ -645,6 +705,15 @@ class _UnweightedCountMeasure(_BaseMeasure):
     @lazyproperty
     def _flat_values(self):
         """tuple of int counts before weighting."""
+        # ---If valid_count are expressed in the cube dict, returns its data---
```
slobodan-ilic, line 708: This is great. Maybe put it _under_ the if statement (I know it's generally supposed to be above it, but I think here it will serve us better if it's under). And also, add some more context, like, when this might happen and why (for our future selves).

src/cr/cube/cube.py
```diff
@@ -645,6 +705,15 @@ class _UnweightedCountMeasure(_BaseMeasure):
...
+        if (
+            self._cube_dict["result"]["measures"]
```
slobodan-ilic, line 710: I'm not sure, but is it possible _not_ to have measures? I honestly don't know :)

src/cr/cube/stripe.py
```diff
@@ -579,6 +579,7 @@ def count(self):
 
     @lazyproperty
     def mean(self):
+        """Return stripe of means."""
```
slobodan-ilic, line 582: This is a property, we don't need the "return" because we're trying to describe what it _is_ rather than what it _does_ (because it's a property :) )

slobodan-ilic (changes requested):
Another batch of minor comments

src/cr/cube/cube.py
ernestoarbitrio: ???

src/cr/cube/cube.py
ernestoarbitrio: i don't think it's possible

slobodan-ilic (approved):
Looks good now. If we notice issues in production, we can always fix as we go. And this will certainly need fixing once we merge @scanny 's work. I'm satisfied with this, it can be merged and released.

src/cr/cube/cube.py
```diff
@@ -488,17 +492,20 @@ def _measures(self):
 
     @lazyproperty
     def _numeric_array_dimension(self):
-        """Column dimensions object inflated according to the mean subvariables."""
+        """Column dimension object according to the mean subvariables."""
+        if not self._mean_subvariables:
+            return None
```
slobodan-ilic, line 497: Check if this decreases tests coverage, I'm not 100% sure 😄

PR #239, opened 2021-01-04
Extend cr.cube to deal with num array dimensions and tabbook export

ernestoarbitrio:


src/cr/cube/cube.py
```diff
@@ -415,13 +415,19 @@ def _ca_as_0th(self):
...
+        if self._cube_idx_arg:
+            cube_dict["cube_idx"] = self._cube_idx_arg
         if self._mean_subvariables:
             dimensions = cube_dict.get("result", {}).get("dimensions", [])
             # ---cube inflation---
             # ---In case of numeric arrays, we need to inflate the columns dimension
             # ---according to the mean subvariables. For each subvar the col dimension
             # ---will have a new element related to the subvar metadata.
-            dimensions.append(self._numeric_array_dimension)
+            (
```
slobodan-ilic, line 426: This is not an assignment (nothing on the left side). I would therefore prefer the traditional structure (with `if/else`) instead of the ternary expression.

src/cr/cube/cube.py
```diff
@@ -673,7 +679,22 @@ def raw_cube_array(self):
         )
         # ---must be read-only to avoid hard-to-find bugs---
         raw_cube_array.flags.writeable = False
-        return raw_cube_array
+        return raw_cube_array.T if self._require_array_transposition else raw_cube_array
+
+    @lazyproperty
+    def _require_array_transposition(self):
```
slobodan-ilic, line 685: Consider using `requires`, as a statement instead of `require` (in imperative).

src/cr/cube/cube.py
```diff
@@ -673,7 +679,22 @@ def raw_cube_array(self):
...
+        """True of False if the raw cube array need to be transposed.
```
slobodan-ilic, line 686: Sentence structure needs a little correction, maybe like:
`"""True if raw cube array needs transposition, False otherwise."""`

src/cr/cube/cube.py
```diff
@@ -415,13 +415,19 @@ def _ca_as_0th(self):
     def _cube_dict(self):
         """dict containing raw cube response, parsed from JSON payload."""
         cube_dict = copy.deepcopy(self._cube_response)
+        if self._cube_idx_arg:
+            cube_dict["cube_idx"] = self._cube_idx_arg
```
slobodan-ilic, line 419: Do we really need to set this specifically on the dictionary? I.e. don't we already have it in `self._cube_idx_arg`? And if so, why can't we use the `self._cube_idx_arg` in the other place (where we need to decide)?

src/cr/cube/cube.py
```diff
@@ -673,7 +679,22 @@ def raw_cube_array(self):
         )
         # ---must be read-only to avoid hard-to-find bugs---
         raw_cube_array.flags.writeable = False
-        return raw_cube_array
+        return raw_cube_array.T if self._require_array_transposition else raw_cube_array
```
slobodan-ilic, line 682: This is probably going to get more complicated than a simple `x.T` (when we have more dimensions in the future). So maybe add a `"TODO:"` here... or `"NOTE:"`.

src/cr/cube/cube.py
```diff
@@ -673,7 +679,22 @@ def raw_cube_array(self):
...
+    def _require_array_transposition(self):
+        """True of False if the raw cube array need to be transposed.
+
+        When a Cube is part of a cubeset and one of the dimension type is a NUM_ARRAY,
+        it's cube_array values have to be transposed.
+        This business rule needs to drive correctly tabbook exports when a numeric array
+        dimension is expressed.
+        """
+        return (
+            self._cube_dict.get("cube_idx") is not None
```
slobodan-ilic, line 694: Why can't we just check `self._cube_idx_arg` here?

tests/unit/test_cube.py
```diff
@@ -340,7 +340,8 @@ def it_can_inflate_itself(self, request):
                         },
                         {"other": "dim"},
                     ]
-                }
+                },
+                "cube_idx": 1,
```
slobodan-ilic, line 344: Who needs to provide this part of the JSON? I it's us (from within cube), then it shouldn't live in the fixture.

slobodan-ilic (changes requested):
A couple of comments. Mostly regarding style and typos, but there's also one architectural concern (about where and how se set the index information).

src/cr/cube/cube.py
ernestoarbitrio: Cause in python logic expression like `None and True` returns None` and i'd like to be consistent with a boolean return value

src/cr/cube/cube.py
slobodan-ilic: Also, what happens with other things (not just numbers), like labels and valid indexes? Is it enough to "just" transpose the `raw_cube_array`?

ernestoarbitrio: that's a valid question and i think this can be checked in the exporter :D

src/cr/cube/cube.py
```diff
@@ -653,18 +653,23 @@ def weighted_counts(self):
...
         """
         return (
-            _WeightedCountMeasure(self._cube_dict, self._all_dimensions)
+            _WeightedCountMeasure(
+                self._cube_dict, self._all_dimensions, self._cube_idx_arg
+            )
             if self.is_weighted
-            else _UnweightedCountMeasure(self._cube_dict, self._all_dimensions)
+            else _UnweightedCountMeasure(
+                self._cube_dict, self._all_dimensions, self._cube_idx_arg
```
slobodan-ilic, line 661: Do we need it in all these instantiations? Or just in `MeanMeasure`? It's gonna be `None` in all of those that we don't need, right?

src/cr/cube/cube.py
```diff
@@ -679,19 +684,23 @@ def raw_cube_array(self):
...
+        """True if raw cube array needs transposition, False otherwise.
 
         When a Cube is part of a cubeset and one of the dimension type is a NUM_ARRAY,
         it's cube_array values have to be transposed.
         This business rule needs to drive correctly tabbook exports when a numeric array
         dimension is expressed.
         """
         return (
-            self._cube_dict.get("cube_idx") is not None
+            self._cube_idx_arg is not None
```
slobodan-ilic, line 703: Ah, this is so much nicer!!!

src/cr/cube/cube.py
ernestoarbitrio: we do need, otherwise we don't swap the summary counts in the exporter:
``` 
E                       5     Dark Knight           73.60           87.67           52.50
E                       6      Fight Club           82.50           93.33           50.00
E                       7  Meets the Par.           23.00            1.00           45.00
E       -               8    Unweighted N              11               7               4
E       +               8    Unweighted N              11               5               4
```

slobodan-ilic (approved):

PR #240, opened 2021-01-11
Measure consolidation 176212559

scanny:
Jan-20-2021 STATUS: Code complete, ready for merging to master. Note there is a companion branch/PR on zoom that is required in order to pin to this new version.

### DIFF
--- a/src/cr/cube/matrix/cubemeasure.py
+++ b/src/cr/cube/matrix/cubemeasure.py
@@ -167,6 +167,18 @@ class _CatXCatUnweightedCubeCounts(_BaseUnweightedCubeCounts):
         return np.sum(self.unweighted_counts, axis=1)
 
     @lazyproperty
+    def table_base(self):
+        """np.int64 count of actual respondents who answered both questions.
+
+        Each dimension of a CAT_X_CAT matrix represents a categorical question. Only
+        responses that include answers to both those questions appear as entries in the
+        valid elements of those dimensions. The sum total of all valid answers is the
+        sample size, aka "N". The term "base" derives from its use as the denominator
+        for the table-proportions measure.
+        """
+        return np.sum(self.unweighted_counts)
+
+    @lazyproperty
     def table_bases(self):
         """2D np.int64 ndarray of table-proportion denominator for each matrix cell."""
         return np.broadcast_to(self.table_base, self._unweighted_counts.shape)
--- a/src/cr/cube/matrix/cubemeasure.py
+++ b/src/cr/cube/matrix/cubemeasure.py
@@ -389,6 +389,19 @@ class _CatXMrWeightedCubeCounts(_BaseWeightedCubeCounts):
         return self.rows_margin
 
     @lazyproperty
+    def rows_margin(self):
+        """2D np.float64 ndarray of weighted-N for each cell of this matrix.
+
+        A matrix with an MR columns dimension has a distinct rows-margin for each cell.
+        This is because not all column responses (subvars) are necessarily offered to
+        each respondent. The weighted-count for each X_MR cell is the sum of its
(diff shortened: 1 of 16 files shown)


src/cr/cube/matrix/measure.py
```diff
@@ -159,6 +161,30 @@ def _subtotal_columns(self):
...
+        subtotal_rows = SumSubtotals.subtotal_rows(self._base_values, self._dimensions)
+
+        # --- in the "no-row-subtotals" case, short-circuit with a (0, ncols) array
+        # --- return value, both because that is the right answer, but also because the
+        # --- non-empty columns-base array cannot be broadcast into that shape. dtype
+        # --- must be `int` to avoid changing type of assembled array to float.
+        if subtotal_rows.shape[0] == 0:
+            return subtotal_rows
+
+        return np.broadcast_to(
```
slobodan-ilic, line 218: Are we sure this won't broadcast to a wrong shape if the number of rows and columns is the same?

src/cr/cube/matrix/cubemeasure.py
```diff
@@ -234,7 +234,14 @@ def factory(cls, cube, dimensions, slice_idx):
     @lazyproperty
     def column_bases(self):
         """2D np.float64 ndarray of column-proportion denominator for each cell."""
-        raise NotImplementedError
+        return np.broadcast_to(self.columns_margin, self.weighted_counts.shape)
```
slobodan-ilic, line 391: same here, if we end up with square shape, could this do the wrong thing?

slobodan-ilic (approved):

src/cr/cube/matrix/measure.py
scanny: Because this is broadcasting "vertically", no. The trailing dimension of `subtotal_rows` will always match the length of the columns-marginal (`.columns_base` I think in this case, the code moved so might be `.columns_margin`, but same applies). So if `subtotal_rows` is (3, 3) and `columns_margin` is (3,), the trailing dimension matches (and always will here). Since that is the only dimension in the margin, the (3,) margin is broadcast to have 3 rows and a shape of (3, 3).

```python
>>> import numpy as np
>>> A = np.arange(9).reshape(3, 3)
>>> A
array([[0, 1, 2],
       [3, 4, 5],
       [6, 7, 8]])
>>> B = np.array([1,2,3])
>>> B
array([1, 2, 3])
>>> np.broadcast_to(B, A.shape)
array([[1, 2, 3],
       [1, 2, 3],
       [1, 2, 3]])
```

src/cr/cube/matrix/cubemeasure.py
scanny: Same principle applies here. The interesting other example is when you want to broadcast a _rows_ marginal, i.e a 1D array that is to be interpreted as a "column". In that case, the array needs to be inflated so each value appears in its own "row" of length 1. Then the trailing dimensions don't match anymore, but since the marginal's trailing dimension is 1, the size (column count) from the measure array is used:

```python
>>> A
array([[0, 1, 2],
       [3, 4, 5],
       [6, 7, 8]])
>>> B = np.array([1,2,3])[:, None]
>>> B
array([[1],
       [2],
       [3]])
>>> np.broadcast_to(B, A.shape)
array([[1, 1, 1],
       [2, 2, 2],
       [3, 3, 3]])
```

You can see this pattern appearing in `_RowUnweighted` and `_RowWeighted._subtotal_columns` properties.

One of the side benefits of having the measures segregated like this is that the cases are very narrow, like we don't have to deal with figuring out what the form of the marginal is ("is this a 'row' or 'column' in this case?") and the code can be more straightforward, focused on a single case.

PR #242, opened 2021-01-18
Fix valid_idxs to handle transposition in case of num array, expose valid count measure for missing calc

ernestoarbitrio:
Fix valid_idxs to handle transposition in case of num array, expose valid count measure for missing calc

### DIFF
--- a/src/cr/cube/cube.py
+++ b/src/cr/cube/cube.py
@@ -516,8 +516,13 @@ class Cube(object):
 
     @lazyproperty
     def _valid_idxs(self):
-        return np.ix_(
+        valid_idxs = np.ix_(
             *tuple(d.valid_elements.element_idxs for d in self._all_dimensions)
+        )
+        return (
+            valid_idxs[::-1]
+            if self._measure(self.is_weighted).requires_array_transposition
+            else valid_idxs
         )
 
 
@@ -664,12 +669,10 @@ class _BaseMeasure(object):
         raw_cube_array.flags.writeable = False
         # NOTE: The transpose operation cannot be enough in future when we'll have more
         # than 2 dimensions.
-        return (
-            raw_cube_array.T if self._requires_array_transposition else raw_cube_array
-        )
+        return raw_cube_array.T if self.requires_array_transposition else raw_cube_array
 
     @lazyproperty
-    def _requires_array_transposition(self):
+    def requires_array_transposition(self):
         """True if raw cube array needs transposition, False otherwise.
 
         When a Cube is part of a cubeset and one of the dimension type is a NUM_ARRAY,
--- a/src/cr/cube/cube.py
+++ b/src/cr/cube/cube.py
@@ -381,8 +381,9 @@ class Cube(object):
     @lazyproperty
     def valid_counts(self):
         """ndarray of valid counts, valid elements only."""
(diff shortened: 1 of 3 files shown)


src/cr/cube/cube.py
```diff
@@ -516,9 +539,14 @@ def _slice_idxs(self):
 
     @lazyproperty
     def _valid_idxs(self):
-        return np.ix_(
+        valid_idxs = np.ix_(
             *tuple(d.valid_elements.element_idxs for d in self._all_dimensions)
         )
+        return (
+            valid_idxs[::-1]
```
slobodan-ilic, line 546: This is going to get more elaborate in the future, probably. We'll need a way to tell to the cube _which_ dimensions we want transposed and which not. Just keep it in mind.

src/cr/cube/cube.py
```diff
@@ -363,6 +378,14 @@ def unweighted_counts(self):
         """
         return self._measures.unweighted_counts.raw_cube_array[self._valid_idxs]
 
+    @lazyproperty
+    def valid_count(self):
+        """ndarray of valid counts, valid elements only."""
+        # --- In case of ndim > 2 the sum should be done on the second axes to get the
+        # --- correct sequence of valid count (e.g. CA_SUBVAR).
+        axes = 1 if len(self._all_dimensions) > 2 else 0
+        return np.sum(self._measures.means.valid_count[self._valid_idxs], axis=axes)
```
slobodan-ilic, line 387: We shouldn't sum here. Because we lose information that way. Either add a "summary" property, for the valid counts, or do the summation from the exporter code. I'd prefer the summary method.

We might still sometimes need the verbose valid count, and to suggest that they're always collapsed is not correct.

src/cr/cube/cube.py
```diff
@@ -132,6 +132,16 @@ def population_fraction(self):
         """
         return self._cubes[0].population_fraction
 
+    @lazyproperty
+    def total_n(self):
```
slobodan-ilic, line 136: The name `total` is a little bit ambiguous here. It doesn't say explicitly what it is. Maybe consider something like num_rows perhaps? I'm not against `total`, I'd just like more explicitness. Like, is it the total number of responses considered? Maybe just add that in the docstring...

slobodan-ilic (changes requested):
Left some comments. The most important one is for the summary vs. just counts.

src/cr/cube/cube.py
ernestoarbitrio: ok i'll write a todo there

src/cr/cube/cube.py
ernestoarbitrio: ok

src/cr/cube/cube.py
```diff
@@ -363,6 +378,23 @@ def unweighted_counts(self):
         """
         return self._measures.unweighted_counts.raw_cube_array[self._valid_idxs]
 
+    @lazyproperty
+    def valid_counts(self):
+        """ndarray of valid counts, valid elements only."""
+        if self.has_means and self._measures.means.valid_counts.any():
```
slobodan-ilic, line 384: Having means doesn't necessarily mean that it will also have valid counts.

slobodan-ilic (approved):

PR #244, opened 2021-01-26
Columns base properties for numeric array matrices

ernestoarbitrio:
- `columns_base` props for NumericArray matrix types
- `table_base` for numeric array stripe

For reference please see the activities section in here: https://www.pivotaltracker.com/story/show/176085256

### DIFF
--- a/src/cr/cube/cube.py
+++ b/src/cr/cube/cube.py
@@ -307,7 +307,7 @@ class Cube(object):
     @lazyproperty
     def missing(self):
         """Get missing count of a cube."""
-        return self._measures.missing_count if self._show_missing else 0
+        return self._measures.missing_count if self._has_missing else 0
 
     @lazyproperty
     def name(self):
@@ -391,7 +391,7 @@ class Cube(object):
         """ndarray of summary valid counts"""
         # --- In case of ndim > 2 the sum should be done on the second axes to get the
         # --- correct sequence of valid count (e.g. CA_SUBVAR).
-        if self.valid_counts.any() and self._show_missing:
+        if self.valid_counts.any() and self._has_missing:
             axes = 1 if len(self._all_dimensions) > 2 else 0
             return np.sum(self.valid_counts, axis=axes)
         return np.empty(0)
@@ -454,6 +454,11 @@ class Cube(object):
                 "Unsupported type <%s> provided. Cube response must be JSON "
                 "(str) or dict." % type(self._cube_response_arg).__name__
             )
+
+    @lazyproperty
+    def _has_missing(self):
+        """True if cube dimensions doesn't contain MR and NUM_ARRAY else False."""
+        return not all(d in self.dimension_types for d in [DT.MR, DT.NUM_ARRAY])
 
     @lazyproperty
     def _is_single_filter_col_cube(self):
@@ -546,11 +551,6 @@ class Cube(object):
         if self.ndim < 3 and not self._ca_as_0th:
             return (0,)
         return range(len(self.dimensions[0].valid_elements))
-
-    @lazyproperty
-    def _show_missing(self):
-        """True if cube dimensions doesn't contain MR and NUM_ARRAY else False."""
-        return not all(d in self.dimension_types for d in [DT.MR, DT.NUM_ARRAY])
 
     @lazyproperty
     def _valid_idxs(self):
@@ -737,7 +737,6 @@ class _BaseMeasure(object):
             .get("valid_count_unweighted", {})
             .get("data", [])
         )
-
         if not valid_counts:
             return np.empty(0)
         valid_counts_array = np.array(valid_counts).reshape(self._all_dimensions.shape)
--- a/src/cr/cube/cube.py
+++ b/src/cr/cube/cube.py
@@ -389,11 +389,11 @@ class Cube(object):
     @lazyproperty
     def valid_counts_summary(self):
         """ndarray of summary valid counts"""
-        # --- In case of ndim > 2 the sum should be done on the second axes to get the
-        # --- correct sequence of valid count (e.g. CA_SUBVAR).
         if self.valid_counts.any() and self._has_missing:
-            axes = 1 if len(self._all_dimensions) > 2 else 0
-            return np.sum(self.valid_counts, axis=axes)
+            # --- In case of ndim > 2 the sum should be done on the second axes to get
+            # --- the correct sequence of valid count (e.g. CA_SUBVAR).
+            axis = 1 if len(self._all_dimensions) > 2 else 0
+            return np.sum(self.valid_counts, axis=axis)
         return np.empty(0)
 
     @lazyproperty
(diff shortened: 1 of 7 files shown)


src/cr/cube/cube.py
```diff
@@ -547,19 +547,30 @@ def _slice_idxs(self):
             return (0,)
         return range(len(self.dimensions[0].valid_elements))
 
+    @lazyproperty
+    def _show_missing(self):
```
slobodan-ilic, line 551: Why is this called `"_show_missing"`? It implies that we have some kind of "show" settings, which are different from context to context. Crunch-cube is strictly a 2nd order analysis library, and shouldn't be concerned with the presentation. A more appropriate name would be something that implies a "has" concept, rather than "show".

src/cr/cube/cube.py
```diff
@@ -547,19 +547,30 @@ def _slice_idxs(self):
...
+        """True if cube dimensions doesn't contain MR and NUM_ARRAY else False."""
+        return not all(d in self.dimension_types for d in [DT.MR, DT.NUM_ARRAY])
```
slobodan-ilic, line 553: This would be easier to understand if you put explicit return paths for concrete types (there's only two of them, so no harm done), something like:
```
if DT.MR in self.dimension_types:
    return False
if DT.NUM_ARRAY in self.dimension_types:
    return False
return True

File: src/cr/cube/cube.py [hunk 3]
```diff
@@ -547,19 +547,30 @@ def _slice_idxs(self):
...
-            if self._measure(self.is_weighted).requires_array_transposition
-            else valid_idxs
-        )
+
+        def _reshape_idxs(valid_indices):
+            if self._measure(self.is_weighted).requires_array_transposition:
+                if len(self._all_dimensions) == 3:
+                    # ---In case of 3D array and a numeric array is involved we have to
+                    # ---change the order of the valid idxs, from [0,1,2] to [1,2,0]
+                    return valid_indices[1], valid_indices[-1], valid_indices[0]
```
slobodan-ilic, line 566: I'm not sure that this is always true. We _could_ specify a different transform, in which case you would have to check it (the transform) in order to generate the right shape. We should also put a note here that we're expecting this in the future.

src/cr/cube/cube.py
```diff
@@ -726,6 +737,7 @@ def valid_counts(self):
             .get("valid_count_unweighted", {})
             .get("data", [])
         )
+
```
slobodan-ilic, line 740: Accidental blank line? I don't think it's needed here, as the previous line only has a single character, so everything is pretty visible. I'm all for adding blank lines when they increase visibility though :)

src/cr/cube/matrix/assembler.py
```diff
@@ -85,16 +85,16 @@ def column_weighted_bases(self):
     def columns_base(self):
         """1D/2D np.int64 ndarray of unweighted-N for each slice column/cell."""
         # --- an MR_X slice produces a 2D column-base (each cell has its own N) ---
-        if self._rows_dimension.dimension_type == DT.MR_SUBVAR:
+        columns_base = self._cube_result_matrix.columns_base
+        rows_dim_type = self._rows_dimension.dimension_type
+        if rows_dim_type in (DT.MR_SUBVAR, DT.NUM_ARRAY) and columns_base.ndim > 1:
```
slobodan-ilic, line 90: Is there a case when the first part of the `and` is satisfied, but the second is not? I'm curious to see if we have an array where the columns base `ndim` would be 1.

slobodan-ilic (changes requested):
Per comments

src/cr/cube/matrix/assembler.py
ernestoarbitrio: yes num_array no group should have the first condition true and the second false

src/cr/cube/cube.py
ernestoarbitrio: this is not the same condition. I have to return false only if all the dimension types matched
In your example you return false right after an MR is in the dimension types that is not correct. This rule has the goal of aviod show missing summary row only if the cube is MRxNUMARR and vice versa

src/cr/cube/cube.py
ernestoarbitrio: i think that these 2 conditions matches that inversion operation:
```python
    if self._measure(self.is_weighted).requires_array_transposition:
        if len(self._all_dimensions) == 3:
```
I do agree that ATM this is an hack waiting for the dim_order ... but i'm not sure i follow your concern

src/cr/cube/cube.py
```diff
@@ -455,6 +455,11 @@ def _cube_response(self):
                 "(str) or dict." % type(self._cube_response_arg).__name__
             )
 
+    @lazyproperty
+    def _has_missing(self):
```
slobodan-ilic, line 459: Now that I read it, it doesn't really have to do anything with "has missing", because it doesn't even check the measures. We should figure out a better name for this. I now why you wanted to use "show", but it's still wrong. Let me think about it.

src/cr/cube/cube.py
```diff
@@ -455,6 +455,11 @@ def _cube_response(self):
...
+        """True if cube dimensions doesn't contain MR and NUM_ARRAY else False."""
+        return not all(d in self.dimension_types for d in [DT.MR, DT.NUM_ARRAY])
```
slobodan-ilic, line 461: I have a problem interpreting the structure of `not all(. . .)`. I think it'd be easier to interpret as:
```
return DT.MR not in self.dimension_types and DT.NUM_ARRAY not in self.dimension_types
```
but that might be obsolete 🤷

src/cr/cube/cube.py
```diff
@@ -391,7 +391,7 @@ def valid_counts_summary(self):
         """ndarray of summary valid counts"""
         # --- In case of ndim > 2 the sum should be done on the second axes to get the
         # --- correct sequence of valid count (e.g. CA_SUBVAR).
-        if self.valid_counts.any():
+        if self.valid_counts.any() and self._has_missing:
             axes = 1 if len(self._all_dimensions) > 2 else 0
```
slobodan-ilic, line 395: This should say `axis`, not `axes`.

src/cr/cube/cube.py
```diff
@@ -391,7 +391,7 @@ def valid_counts_summary(self):
         """ndarray of summary valid counts"""
         # --- In case of ndim > 2 the sum should be done on the second axes to get the
```
slobodan-ilic, line 392: This should go under the `if` condition, because it doesn't apply to it, but to the first line _in_ it.

src/cr/cube/cube.py
ernestoarbitrio: sorry but even this one is not the same of what i wrote, I'll try to simplify it in a further commit

slobodan-ilic (approved):

PR #245, opened 2021-01-28
Sort by column percent 176603042

scanny:
Install sort-by-value machinery and wire it up for column-percent.

This is so-far limited to sort-rows-by-column-percent-column, just because there's no call yet for sorting columns by column-percent. That can be added with minimal additional work when and if the need arises.

Note this does not include sort-rows-by-the-All-percent-marginal that typically appears at the right side of such a crosstab. That will require some additional noodling on how we handle marginals (and particularly marginal-ranges) in the new measure architecture.

### DIFF
--- a/src/cr/cube/collator.py
+++ b/src/cr/cube/collator.py
@@ -387,7 +387,7 @@ class SortByValueCollator(_BaseCollator):
         specified measure in each subtotal, except that any NaN values drop to the end
         of the subtotal group.
         """
-        raise NotImplementedError
+        return () if self._descending else self._subtotal_idxs
 
     @lazyproperty
     def _descending(self):
--- a/src/cr/cube/collator.py
+++ b/src/cr/cube/collator.py
@@ -281,3 +281,27 @@ class PayloadOrderCollator(_BaseAnchoredCollator):
         return tuple(
             (idx, idx, element_id) for idx, element_id in enumerate(self._element_ids)
         )
+
+
+class SortByValueCollator(_BaseCollator):
+    """Produces an idx ordering based on values in a vector.
+
+    The items in `element_values` must correspond directly the the (valid) elements of
+    `dimension`, both in number and sequence (payload order). The items in
+    `subtotal_values` must also correspond directly in number and sequence with the
+    subtotals defined on `dimension`.
+
+    In general, the anchors used to position inserted subtotals lose their meaning when
+    the dimension is sorted by-value. In sort-by-value cases, subtotals are grouped at
+    the top (when sort direction is descending (default)) or the bottom (when direction
+    is ascending), while also being sorted by the specified value.
+    """
+
+    @classmethod
+    def display_order(cls, dimension, element_values, subtotal_values, empty_idxs):
+        """Return sequence of int element-idxs ordered by sort on `element_values`.
+
+        The returned tuple contains a signed-int value for each vector and subtotal of
+        `dimension` that is not hidden, sorted (primarily) by the element value.
+        """
+        raise NotImplementedError
--- a/src/cr/cube/collator.py
+++ b/src/cr/cube/collator.py
@@ -17,7 +17,11 @@ from cr.cube.util import lazyproperty
 
 
 class _BaseCollator(object):
-    """Base class for all collator objects, providing shared properties."""
+    """Base class for all collator objects, providing shared properties.
+
+    `empty_idxs` is a tuple of int element-index for each element vector with a zero
+    unweighted N (N = 0).
+    """
 
     def __init__(self, dimension, empty_idxs):
         self._dimension = dimension
@@ -63,10 +67,13 @@ class _BaseAnchoredCollator(_BaseCollator):
         The returned indices are "signed", with positive indices applying to base
         vectors and negative indices applying to inserted vectors. Both work for
         indexing in their respective unordered collections.
+
+        `empty_idxs` identifies vectors with N=0, which may be "pruned", depending on
+        a user setting in the dimension.
         """
         return cls(dimension, empty_idxs)._display_order
 
-    @property
+    @lazyproperty
     def _display_order(self):
         """tuple of int element-idx for each element in assembly order.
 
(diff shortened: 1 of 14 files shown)


tests/integration/test_collator.py
```diff
@@ -57,3 +62,56 @@ def it_knows_the_display_order_for_a_dimension(
...
 
         assert display_order == expected_value
+
+
+class DescribeSortByValueCollator(object):
+    """Partial-integration test suite for `SortByValueCollator` object."""
+
+    @pytest.mark.xfail(reason="WIP", raises=NotImplementedError, strict=True)
+    @pytest.mark.parametrize(
+        "order, xtop, xbot, element_vals, empty_idxs, expected_value",
```
slobodan-ilic, line 71: I can't figure the meaning of `xtop` by just reading this test.

src/cr/cube/collator.py
```diff
@@ -331,4 +331,65 @@ def _display_order(self):
         in the `"exclude": [...]` array of the order transform. Base elements appear in
         value-sorted order within their grouping.
         """
+        hidden_idxs = self._hidden_idxs
+        return tuple(
+            idx
+            for idx in (
+                self._top_subtotal_idxs
+                + self._top_exclusion_idxs
```
slobodan-ilic, line 342: "exclusion" sounds right in the construct "exclusion from sort", but on its own it might suggest that the elements are somehow excluded completely, which is the first thing I thought. I then went back to the docstring and figured out what it meant. Maybe consider a slightly different approach to naming? Something like `_top_not_sorted_idxs`?

tests/integration/test_cubepart.py
```diff
@@ -402,6 +402,36 @@ def and_it_respects_explicit_order_transform_for_measures(
...
+                    "measure": "col_percent",
+                    "direction": "ascending",
+                    # --- element-ids are 1, 2, 3, 999 ---
+                    "exclude": {"top": [999], "bottom": [1]},
+                }
+            }
+        }
+        slice_ = _Slice(Cube(CR.CAT_4_X_CAT_5), 0, transforms, None, 0)
+
+        expected = [
```
slobodan-ilic, line 425: Maybe a non-sorted version would be useful for comparison (for the test reader) alongside this one.

tests/integration/test_matrix.py
```diff
@@ -1457,3 +1457,54 @@ def it_computes_mr_x_mr_table_stderr(self):
                 ]
             ),
         )
+
+    @pytest.mark.parametrize(
```
slobodan-ilic, line 1462: I don't see that this has an `xfail` at this particular commit (`d4981c133f75edd514cedc873656353cb3680daa`).

src/cr/cube/matrix/assembler.py
```diff
@@ -710,6 +710,23 @@ def _rows_dimension(self):
...
+        Negative values represent inserted-vector locations. Returned sequence reflects
+        insertion, hiding, pruning, and ordering transforms specified in the
+        rows-dimension.
+        """
+        CollatorCls = (
+            ExplicitOrderCollator
+            if self._columns_dimension.collation_method == CM.EXPLICIT_ORDER
+            else PayloadOrderCollator
+        )
+        return CollatorCls.display_order(
```
slobodan-ilic, line 658: is the signature going to change for future collators? maybe worth noting if it is...

src/cr/cube/matrix/cubemeasure.py
```diff
@@ -312,6 +312,15 @@ def columns_base(self):
...
         return np.sum(self._unweighted_counts, axis=1)
 
+    @lazyproperty
+    def columns_pruning_base(self):
+        """1D np.int64 ndarray of unweighted-N for each matrix column.
+
+        These values include both the selected and unselected counts of the MR rows
+        dimension.
+        """
+        return np.sum(self._unweighted_counts, axis=(0, 1))
```
slobodan-ilic, line 322: This seems odd, you're adding across subvariables dimension. Maybe a better choice here would be to use `not all` across that (zeroth) subvariables dimension. It's gonna amount to the same thing, since we're only using it to show/hide values, but maybe worth considering.

slobodan-ilic (approved):
left a couple of comments

tests/integration/test_collator.py
scanny: I figured since these values appear in the `"exclude": {"top": xtop, "bottom": xbot}` structure in the test body, that I could manage to get these parameters to fit on a single line and keep the test shorter. I don't generally try to make the tests explain the concepts of the code they exercise, reserving that effort for the docstrings in the code itself. But I can think about that more for future tests.

src/cr/cube/collator.py
scanny: Yeah, that might be an idea. The exclusion language came in right at the start when this was specced, as apparently the users' language for this concept and I didn't question it too much. It's probably too much to change at this point because we've consistently applied that term from the schema and validation all the way through to here, so probably more work than we can afford at this point and I'm not sure there really is a purely intuitively-descriptive name for the concept.

tests/integration/test_cubepart.py
scanny: I don't care much for compound tests, but I can add some comments so a reader can tell which parts of the expectation show what aspect.

src/cr/cube/matrix/assembler.py
scanny: I expect it to remain the same for these two "anchored" orders (payload and explicit). All other collators are handled by their own distinct class, so it can handle variation in the call signature.

src/cr/cube/matrix/cubemeasure.py
scanny: Yeah, that's an interesting point. I was blowing through all this pretty fast just because there was so much ground to cover, but in retrospect I'm inclined to think that this property should be something more like `.pruning_mask` and return an array of boolean rather than integers that are elsewhere compared to zero to determine that boolean value. That would be better because these counts would never occur outside this method where they could be misinterpreted.

I don't have time for that change today, but maybe I'll sneak it in next time I'm in here.

PR #247, opened 2021-02-05
Transpose row vs col in case of num arrays + revert slobodan dim order changes

ernestoarbitrio:


src/cr/cube/cube.py
```diff
@@ -506,36 +503,35 @@ def _measures(self):
...
             # ---name of the numeric array mean measure subreferences.
-            column_dimension["type"].get("elements", []).append(
+            row_dimension["type"].get("elements", []).append(
                 {
                     "id": i,
                     "value": {
                         "references": {
                             "alias": subrefs[i].get("alias") if subrefs else None,
                             "name": subrefs[i].get("name") if subrefs else None,
-                        },
```
slobodan-ilic, line 533: We should still leave this here, it conforms to all the other subvars. We're not gonna use it though, but we should keep it...

src/cr/cube/cube.py
```diff
@@ -506,36 +503,35 @@ def _measures(self):
 
     @lazyproperty
     def _numeric_array_dimension(self):
-        """Column dimension object according to the mean subvariables."""
+        """Row dimension object according to the mean subvariables."""
```
slobodan-ilic, line 506: I think we need to use "rows dimension" wording, rather than "row dimension". Plural is an important part of how we  view the measures and everything...

src/cr/cube/cube.py
```diff
@@ -506,36 +503,35 @@ def _measures(self):
...
         if not self._mean_subvariables:
             return None
         subrefs = self._mean_references.get("subreferences", [])
-        column_dimension = {
+        row_dimension = {
```
slobodan-ilic, line 510: also here

src/cr/cube/cube.py
```diff
@@ -716,15 +712,12 @@ def raw_cube_array(self):
     def requires_array_transposition(self):
         """True if raw cube array needs transposition, False otherwise.
 
-        When a Cube is part of a cubeset and one of the dimension type is a NUM_ARRAY,
-        it's cube_array values have to be transposed.
-        This business rule needs to drive correctly tabbook exports when a numeric array
-        dimension is expressed.
+        When one of the dimension type is a NUM_ARRAY, it's cube_array values have to be
```
slobodan-ilic, line 715: it's => its

src/cr/cube/cube.py
```diff
@@ -716,15 +712,12 @@ def raw_cube_array(self):
...
+        transposed. This business rule needs to drive correctly tabbook and deck exports
+        when a numeric array dimension is expressed.
         """
-        return (
-            self._cube_idx_arg is not None
-            and len(self._all_dimensions) >= 2
-            and any([d.dimension_type == DT.NUM_ARRAY for d in self._all_dimensions])
+        return len(self._all_dimensions) >= 2 and any(
+            [d.dimension_type == DT.NUM_ARRAY for d in self._all_dimensions]
```
slobodan-ilic, line 720: could we use `DT.NUM_ARRAY in self.dimension_types`?

src/cr/cube/cube.py
```diff
@@ -712,13 +713,12 @@ def raw_cube_array(self):
...
 
-        When one of the dimension type is a NUM_ARRAY, it's cube_array values have to be
+        When one of the dimension type is a NUM_ARRAY, its cube_array values have to be
         transposed. This business rule needs to drive correctly tabbook and deck exports
         when a numeric array dimension is expressed.
         """
-        return len(self._all_dimensions) >= 2 and any(
-            [d.dimension_type == DT.NUM_ARRAY for d in self._all_dimensions]
-        )
+        dimension_types = [d.dimension_type for d in self._all_dimensions]
```
slobodan-ilic, line 720: I don't see why this is necessary. I _think_ that `NUM_ARRAY` should appear in `apparent_dimensions`, and thus its type in `self.dimension_types`. Is this not so?

slobodan-ilic (approved):

src/cr/cube/cube.py
ernestoarbitrio: it is true, but we don't have dimension_types property accessible there 🤷🏽‍♂️

PR #248, opened 2021-02-08
New architecture strand 176841194

scanny:
~Spike of Strand re-architecture. Next step is to apply continuous rebasing to add unit tests.~

~As of Feb 12, 2021, the grooming is perhaps half done, I'll pick it up when I return from PTO.~

Completed on Mar 1, 2021 and ready for code-review.


src/cr/cube/stripe/measure.py
```diff
@@ -198,6 +198,39 @@ def scale_mean(self):
         numeric value does not contribute to either the numerator or denominator of that
         computation.
         """
+        # --- value is None when no row-element has been assigned a numeric value ---
+        if not self._numeric_values.size:
```
ernestoarbitrio, line 202: I'd like to be explicit here. I know that this works when size is 0 but i do prefer to have an explicit expression
e.g. `if self._numeric_values.size == 0 `

src/cr/cube/stripe/measure.py
```diff
@@ -198,6 +198,39 @@ def scale_mean(self):
...
+            return None
+
+        # --- Also when the total count is zero, unlikely but possible and would lead to
+        # --- a divide-by-zero error in this case.
+        if not self._total_weighted_count:
```
ernestoarbitrio, line 207: also here

src/cr/cube/stripe/measure.py
```diff
@@ -216,7 +216,19 @@ def scale_median(self):
         This value is `None` when no rows have a numeric-value assigned. Responses
         without a numeric value are not considered.
         """
-        raise NotImplementedError
+        # --- value is None when no row-element has been assigned a numeric value ---
+        if not self._numeric_values.size:
```
ernestoarbitrio, line 220: same explicit condition here `== 0`

src/cr/cube/stripe/measure.py
```diff
@@ -238,7 +238,11 @@ def scale_stddev(self):
         units of scaled-counts and indicates the dispersion of the scaled-count
         distribution from its mean (scale-mean).
         """
-        raise NotImplementedError
+        # --- value is None when no row-element has been assigned a numeric value ---
+        if not self._numeric_values.size:
```
ernestoarbitrio, line 242: same here

src/cr/cube/cubepart.py
```diff
@@ -1533,18 +1518,6 @@ def unweighted_counts(self):
...
-        if np.all(np.isnan(self._numeric_values)):
-            return None
-
-        numeric_values = self._numeric_values[self._numeric_values_mask]
-        counts = self._counts_as_array[self._numeric_values_mask]
-
-        return np.nansum(counts * pow((numeric_values - self.scale_mean), 2)) / np.sum(
-            counts
-        )
-
```
ernestoarbitrio: why get rid of this property?

tests/integration/test_scale_means.py
```diff
@@ -562,8 +547,8 @@ def test_univariate_with_hs_var_scale_means_row():
         "rows_dimension": {"insertions": {}},
     }
     strand = Cube(CR.ECON_BLAME_WITH_HS, transforms).partitions[0]
-    np.testing.assert_almost_equal(strand.var_scale_mean, [1.1363901131679894])
+    assert strand.scale_mean == pytest.approx(2.1735206)
```
ernestoarbitrio, line 550: I see that considering that `var_scale_mean`  no longer exists (maybe for a reason but i don't know yet :D) ... BTW there the name of the test is wrong cause we're testing for scale means and not for its variance

ernestoarbitrio (changes requested):
I left some comments, and more importand there's a missing row in `dimensions.py` not covered by tests.

src/cr/cube/stripe/assembler.py
```diff
@@ -0,0 +1,289 @@
...
+
+    @lazyproperty
+    def _display_order(self):
+        """tuple of signed int idx for each row of stripe.
+
+        Negative values represent inserted-vector locations. Returned sequence reflects
+        insertion, hiding, pruning, and ordering transforms specified in the
+        rows-dimension.
+        """
+        raise NotImplementedError(
```
ernestoarbitrio, line 256: I would add a pragma no cover here

src/cr/cube/stripe/cubemeasure.py
```diff
@@ -0,0 +1,310 @@
...
+        means = cube.counts
+        MeansCls = (
+            _MrCubeMeans if rows_dimension.dimension_type == DT.MR else _CatCubeMeans
+        )
+        return MeansCls(rows_dimension, means)
+
+    @lazyproperty
+    def means(self):
+        """1D np.float64 ndarray of mean for each stripe row."""
+        raise NotImplementedError("`%s` must implement `.means`" % type(self).__name__)
```
ernestoarbitrio, line 77: no cover also here

ernestoarbitrio:
A general comment: add `pragma: no cover` near the `NotImplementedError` ... this way we should go back to 100% of cov

src/cr/cube/cubepart.py
```diff
@@ -1472,7 +1476,7 @@ def table_percentages(self):
 
     @lazyproperty
     def table_proportions(self):
-        return tuple(row.table_proportions for row in self._stripe.rows)
+        return np.array([row.table_proportions for row in self._stripe.rows])
```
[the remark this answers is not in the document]
scanny: Yeah, one of the transitions during this re-architecture was to "embrace" the idea that the interface was numpy-oriented. A year or two ago I had the idea that all interface properties should be plain-Python objects (e.g. tuple), but it turns out that some items can't be that (like the minimum base-size mask) and so better to be consistent and just make them all measures numpy types. I believe I got them all as I encountered them in these commits.

tests/integration/test_cubepart.py
```diff
@@ -792,6 +792,27 @@ def it_provides_values_for_cat_with_means_and_insertions(self):
         assert strand.title == "Untitled"
         assert strand.unweighted_counts == (409, 113, 139, 409, 252)
 
+    def it_provides_values_for_univariate_numeric(self):
+        strand = Cube(CR.NUM, population=9001).partitions[0]
```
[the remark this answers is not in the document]
scanny: The convention is like `CAT_X_CAT` for a 2D, and `CAT_X_CAT_X_CAT` for 3D, so a 1D is represented as `CAT` (or `NUM` in this case). It the few cases we have a 0D, we can just say `NUM_0D` or something. We have many more 1D fixtures than 0D so I think the consistency is more important to apply there.

tests/integration/test_cubepart.py
```diff
@@ -792,6 +792,22 @@ def it_provides_values_for_cat_with_means_and_insertions(self):
         assert strand.title == "Untitled"
         assert strand.unweighted_counts == (409, 113, 139, 409, 252)
 
+    def it_provides_values_for_univariate_datetime(self):
+        strand = Cube(CR.DATE, population=9001).partitions[0]
```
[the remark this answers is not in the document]
scanny: I find the `UNIVARIATE_` prefix redundant. I've left it where I've found it when the fixture is otherwise serviceable just because I didn't want to get side-tracked with too much fixture renovation on this already quite large branch.

tests/integration/test_cubepart.py
```diff
@@ -817,6 +832,94 @@ def it_provides_values_for_cat_with_means_and_insertions(self):
...
+                83.85587,
+                81.74092,
+                57.64860,
+                82.98627,
+                78.11554,
+                76.16142,
+                76.43885,
+                71.04146,
+                82.79688,
+                54.20534,
```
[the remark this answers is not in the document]
scanny: Yeah, we could, but I think the real solution here is to prune down that fixture a little so it has a smaller number of points. We only need maybe 5 values for the tests to be valid. I've shortened it and these each now fit on one line.

tests/integration/test_cubepart.py
```diff
@@ -861,7 +861,7 @@ def it_provides_values_for_univariate_cat(self):
         assert strand.table_proportions == pytest.approx([0.6666667, 0.3333333])
         assert strand.title == "Registered Voters"
         assert strand.unweighted_bases == (15, 15)
-        assert strand.unweighted_counts == (10, 5)
+        assert strand.unweighted_counts.tolist() == [10, 5]
```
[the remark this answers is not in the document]
scanny: `strand.unweighted_counts` is now type `ndarray` which doesn't directly allow comparison. When you say `assert array_x == array_y`, the `array == something` expression forms another array of boolean, and that has no truth value (not == True and also not == False). So the `.tolist()` is pretty good because it verifies the type as `ndarray` (tuple or list have no `.tolist()` method) as well as allow for a convenient `list` comparand.

src/cr/cube/stripe/cubemeasure.py
```diff
@@ -128,10 +128,20 @@ def unweighted_counts(self):
...
 
+    def __init__(self, rows_dimension, weighted_counts):
+        super(_BaseWeightedCubeCounts, self).__init__(rows_dimension)
+        self._weighted_counts = weighted_counts
+
     @classmethod
     def factory(cls, cube, rows_dimension, ca_as_0th, slice_idx):
         """Return _BaseWeightedCubeCounts subclass instance appropriate to `cube`."""
-        raise NotImplementedError
+        if ca_as_0th:
```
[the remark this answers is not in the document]
scanny: I'm open to a better name, but after walking the same path of confusion myself, I've ended up unable to come up with a better name. The payload from zz9 is interpreted differently when the first (0th) dimension is CA, and this is what this flag indicates.

I think what we _might_ be able to do is to encapsulate all the `ca_as_0th` rules inside `Cube` such that no one else needs to keep track of it. I think this would mean moving the responsibility for slicing the raw cube-measure payloads into Cube, which would be nice because that job is presently done very far away down in the cube-measures, several layers down, and that means that `ca_as_0th` has to make that entire trip, which is why it appears in so many calls.

I think that's actually a good direction to pursue next time we need to do some renovation in `cr.cube.cube.Cube`.

src/cr/cube/stripe/measure.py
scanny: yeah, ok. will do

src/cr/cube/stripe/measure.py
scanny: done too.

src/cr/cube/cubepart.py
scanny: This property is used only by the `_ScaledCounts` measure and has moved to that class as `._scale_variance`. There were no clients for this so there is no interface property for it in `_Strand`. If we end up with a client for this then we can add `_Strand.scale_variance` and implement it as `return self._assembler.scale_variance` which in turn would call `self._measures.scaled_counts.scale_variance`.

As far as I can tell, this method should not have been public. Also, it was a smell that we needed helper methods like this for scaled-count computations rather than delegating that to it's own object. But all that is resolved by this measure-consolidation work.

tests/integration/test_scale_means.py
scanny: fixed the name. We wouldn't test a helper method in an integration test because implementation is subject to change and the integration test should still pass if it does. As far as I can tell, `_Strand.var_scale_means` was always a helper, even though it was named as interface.

In any case, `_ScaledCounts._scale_variance` is now tested in the unit tests for that measure.

src/cr/cube/stripe/cubemeasure.py
scanny: These are all patched up except for three skip and xfail tests that need the numeric-array changes in your branch. Be sure to check those and get them back working again when you get rebased onto this one.

src/cr/cube/cubepart.py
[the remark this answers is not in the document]
ernestoarbitrio: k

ernestoarbitrio (approved):

PR #251, opened 2021-02-12
Cube sum measure 176498756

ernestoarbitrio:
- Add 1st order cube sum measure
- Get rid of the `has_mean` concept


scanny (changes requested):
Here are some comments to get started with. I wasn't able to separate your work from my branch for much of it because you have a merge commit at `f176cb8` and that is carrying around a bunch of my pre-merge commits on this branch even though that branch has been merged to master.

I think what you need to do to fix that is:

```bash
git rebase --onto master f176cb8 cube-sum-last-176498756
```

so the diff isn't polluted with all my changes obscuring yours.

I can take another look at it tomorrow once you've gotten that cleared up.

src/cr/cube/cube.py
```diff
@@ -44,6 +47,9 @@ def __init__(self, cube_responses, transforms, population, min_base):
         self._population = population
         self._min_base = min_base
 
+    def __getitem__(self, index):
+        return self._cubes[index]
```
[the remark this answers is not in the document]
ernestoarbitrio: this is something i need in exporter, but at this point i decided to move `_cubes` to `cubes`  and access to the first cube in the exporter using this prop.

src/cr/cube/cube.py
```diff
@@ -232,13 +233,22 @@ def __repr__(self):
         except Exception:
             return super(Cube, self).__repr__()
 
+    @lazyproperty
+    def available_measures(self):
+        """Tuple of available measures in the cube response."""
+        return tuple(self._cube_response.get("result", {}).get("measures", {}).keys())
```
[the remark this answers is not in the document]
ernestoarbitrio: yeah nice consideration

src/cr/cube/cube.py
```diff
@@ -326,6 +345,11 @@ def ndim(self):
         """int count of dimensions for this cube."""
         return len(self.dimensions)
 
+    @lazyproperty
+    def n_responses(self):
+        """Total (int) number of responses considered."""
+        return self._cube_response["result"].get("n", 0)
```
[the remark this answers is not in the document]
ernestoarbitrio: Basically this property is used to calculate the missing value (as a range) in the exporter summary. This field in the payload express the number of respondents ... so I can change this name but this name came during a noodling session on num arrays. So for now i would leave it as is

src/cr/cube/cube.py
```diff
@@ -160,9 +161,9 @@ def iter_cubes():
                     population=self._population,
                     mask_size=self._min_base,
                 )
-                # --- numeric-mean cubes require inflation to restore their
+                # --- numeric-measures cubes require inflation to restore their
                 # --- rows-dimension, others don't
-                yield cube.inflate() if self._is_numeric_mean else cube
+                yield cube.inflate() if self._is_numeric_measure else cube
```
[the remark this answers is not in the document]
ernestoarbitrio: Maybe you right that the name is flaky but IMHO this cannot be reverted back. The reason is that now the inflation needs to be done even for `sum` `ste_dev`  or whatever other numeric measure (AKA Quantity of Interests measures) we will have. So `is_numeric_mean`  is meaningless too 🤷🏽‍♂️

src/cr/cube/matrix/assembler.py
```diff
@@ -254,6 +252,15 @@ def rows_margin(self):
...
         )
 
+    @lazyproperty
+    def sum(self):
+        """2D optional np.float64 ndarray of sum for each cell.
+        Raises `ValueError` if the cube-result does not include a sum cube-measure.
+        """
+        return self._assemble_matrix(
+            NanSubtotals.blocks(self._cube_result_matrix.sum, self._dimensions)
+        )
```
scanny, line 262: Nothing new should be using `._cube_result_matrix` because that's the legacy way. This should be something like:

`self._assemble_matrix(self._measures.sum.blocks)`

See how `Assembler.weighted_counts` works and model your approach after that. Basically:

1. `Assembler calls `._assemble_matrix` on a collection of blocks (2x2 list of lists).
2. blocks are provided by the appropriate measure, so probably there is a new `_Sum` measure added to `cr.cube.matrix.measures` and that provides the blocks. The blocks are body values, subtotal-rows, subtotal-columns, and intersections.
3. The measure computes the blocks with the help of a cube-measure (like maybe a new `_CubeSums` object, possibly a `_BaseCubeSums` object and `CatXCat, CatXMr, MrXCat, and MrXMr` subclasses, depending on the variety required by other dimensions. The measure also computes the subtotatals, often with the help of a Subtotals collaborator like perhaps `NanSubtotals` in this case.

src/cr/cube/stripe/cubemeasure.py
```diff
@@ -97,9 +104,66 @@ class _MrCubeMeans(_BaseCubeMeans):
...
+            raise ValueError(
+                "`.means` is undefined for a cube-result without a means measure"
+            )
         return self._means[:, 0]
 
 
+# === SUM ===
+
+
+class _BaseCubeSum(_BaseCubeMeasure):
```
scanny, line 117: I think this would be `_BaseCubeSums` (not `Sum`).

src/cr/cube/stripe/cubemeasure.py
```diff
@@ -97,9 +104,66 @@ class _MrCubeMeans(_BaseCubeMeans):
...
+    """Base class for means cube-measure variants."""
```
scanny, line 118: Looks like unrevised copy-paste from means cube-measure.

src/cr/cube/stripe/cubemeasure.py
```diff
@@ -210,6 +277,32 @@ def unweighted_counts(self):
...
+        Each row in a CAT stripe has the same base (the table-base).
+        """
+        return self._unweighted_counts
+
+    @lazyproperty
+    def pruning_base(self):
+        """1D np.int64 ndarray of unweighted-N for each matrix row.
+
+        Because this matrix has no MR dimension, this is simply the unweighted count for
+        each row.
```
scanny, line 296: I notice you're leaving a lot of unrevised copy-paste errors like these that relate to the place you copied them from but not this. I recommend you make updating the docstring the first order of business when you copy a "template" from somewhere else, so you're less likely to forget.

src/cr/cube/stripe/measure.py
```diff
@@ -305,6 +310,33 @@ def _weighted_counts(self):
...
+        """
+        return NanSubtotals.subtotal_values(self.base_values, self._rows_dimension)
+
+    @lazyproperty
+    def _cube_sum(self):
+        """_BaseCubeMeans subclass instance for this measure.
+        Provides the means measures from the cube-result, encapsulating variation based
+        on dimensionality.
+        """
+        return self._cube_measures.cube_sum
```
scanny, line 337: There's only one caller for this, it would be better inlined.

src/cr/cube/stripe/measure.py
```diff
@@ -305,6 +310,33 @@ def _weighted_counts(self):
         return self._weighted_cube_counts.weighted_counts[self._has_numeric_value]
 
 
+class _Sum(_BaseSecondOrderMeasure):
+    """Provides the means measure for a stripe.
+    Relies on the presence of a means cube-measure in the cube-result.
+    """
```
scanny, line 316: blank line between first and later lines of docstring.

src/cr/cube/stripe/measure.py
```diff
@@ -305,6 +310,33 @@ def _weighted_counts(self):
...
+    @lazyproperty
+    def base_values(self):
+        """1D np.float64 ndarray of sum for each row."""
+        return self._cube_sum.sum
+
+    @lazyproperty
+    def subtotal_values(self):
+        """1D ndarray of np.nan for each row-subtotal.
+        Sum values cannot be subtotaled and each subtotal value is unconditionally
+        np.nan.
```
scanny, line 327: blank line between first and later lines of docstring.

src/cr/cube/stripe/measure.py
```diff
@@ -305,6 +310,33 @@ def _weighted_counts(self):
...
+        """
+        return NanSubtotals.subtotal_values(self.base_values, self._rows_dimension)
+
+    @lazyproperty
+    def _cube_sum(self):
+        """_BaseCubeMeans subclass instance for this measure.
+        Provides the means measures from the cube-result, encapsulating variation based
+        on dimensionality.
+        """
```
scanny, line 336: blank line between first and later lines of docstring (PEP 257).

src/cr/cube/cube.py
```diff
@@ -47,29 +47,48 @@ def __init__(self, cube_responses, transforms, population, min_base):
...
+                    cube_idx=idx if self._is_multi_cube else None,
+                    transforms=self._transforms_dicts[idx],
+                    population=self._population,
+                    mask_size=self._min_base,
+                )
+                # --- numeric-measures cubes require inflation to restore their
+                # --- rows-dimension, others don't
+                yield cube.inflate() if self._is_numeric_measure else cube
+
+        return tuple(iter_cubes())
```
scanny, line 81: Why does `.cubes` need to be public now?

scanny (changes requested):
Ok, you have some pretty big changes to do on the matrix version because this needs to be a measure, not the legacy cube_result_matrix style. I'll take another look once you've got that sorted out.

src/cr/cube/cube.py
ernestoarbitrio: i need to check for the first cube of the cubeset in the exporter. So the options are 1. make this public, 2. make CubeSet subscriptable. I chose this way :D

src/cr/cube/matrix/assembler.py
ernestoarbitrio: I think i did it now, included the `Mean`  measure

src/cr/cube/cube.py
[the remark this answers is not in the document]
scanny: Okay, I'll buy that, and actually I think `._is_numeric_measure` is quite a good name if that's what we're using this for. But you're going to need to update the docstring for `._is_numeric_measure` because it still says `True when CubeSet is special-case "numeric-mean" case requiring inflation."

Also, the logic in there seems odd to me, is it true that whenever the rows-dimension cube measure is sum or std_dev that it has 0 dimensions (is a Nub)? If so, then a comment in there would be good. Since there were no changes to that property other than the name I assumed nothing had changed, but it sounds like that same logic now has broader applicability.

src/cr/cube/cube.py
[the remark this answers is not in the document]
scanny: Ah, okay, this was just moved, got it :)

src/cr/cube/cube.py
```diff
@@ -779,17 +814,34 @@ def _flat_values(self):
...
+    def _flat_values(self):
+        """Return tuple of mean values as found in cube response.
+        Mean data may include missing items represented by a dict like
+        {'?': -1} in the cube response. These are replaced by np.nan in the
+        returned value.
+        """
+        return tuple(
+            np.nan if type(x) is dict else x
+            for x in self._cube_dict["result"]["measures"]["sum"]["data"]
+        )
```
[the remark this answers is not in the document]
scanny: btw, I believe Greg is doing this in his branch, since he added something to do with flat values, so you might want to take a look at his branch before you merge it in, to get a sense of where the conflicts might lie.

src/cr/cube/cube.py
scanny: What specifically do you mean by "... check for the first cube of the cubeset ..."? You need to check that it's there? Or you need to inspect some attributes for some reason? Please describe more fully so I can understand. There are a number of cases where work is done in exporter that should be done here, so that's what I'm looking to understand.

src/cr/cube/cubepart.py
```diff
@@ -1412,6 +1412,14 @@ def smoothed_dimension_dict(self):
         """dict, row dimension definition"""
         return self._rows_dimension._dimension_dict
 
+    @lazyproperty
+    def sum(self):
+        """1D np.float64 ndarray of sum for each row of strand.
+        Raises ValueError when accessed on a cube-result that does not contain a sum
+        cube-measure.
+        """
```
scanny, line 1420: same

src/cr/cube/stripe/assembler.py
```diff
@@ -134,6 +134,14 @@ def scale_stderr(self):
         """
         return self._measures.scaled_counts.scale_stderr
 
+    @lazyproperty
+    def sum(self):
+        """1D np.float64 ndarray of sum for each row.
+        The array is filled with `np.nan` when the cube-result does not include a sum
+        cube-measure.
+        """
```
scanny, line 142: The array is not filled with `np.nan` when there's no cube-sum measure, is it? It raises `ValueError` further down in that case. If docstring goes past the summary line it needs a blank line between.

src/cr/cube/stripe/cubemeasure.py
```diff
@@ -28,6 +28,11 @@ def cube_means(self):
         """_BaseCubeMeans subclass object for this stripe."""
         return _BaseCubeMeans.factory(self._cube, self._rows_dimension)
 
+    @lazyproperty
+    def cube_sum(self):
+        """_BaseCubeMeans subclass object for this stripe."""
```
scanny, line 33: `_BaseCubeSums` subclass, not `_BaseCubeMeans`.

src/cr/cube/stripe/cubemeasure.py
```diff
@@ -85,6 +88,10 @@ class _CatCubeMeans(_BaseCubeMeans):
     @lazyproperty
     def means(self):
         """1D np.float64 ndarray of mean for each stripe row."""
+        if self._means is None:
+            raise ValueError(
+                "`.means` is undefined for a cube-result without a means measure"
+            )
```
scanny, line 94: As I think more about this, I'm thinking that the exception-raising should move upwards, like into the factory, rather than have it down here where it needs to be repeated for each subclass. There's no way the factory can produce a valid `CubeMeans` object if `cube.means` is `None`, right? So might as well move it up there and reduce the number of times these four lines appear repeated in the code.

I'm _almost_ thinking that we should move it further, right up to `Cube.means`, but that would complicate testing for whether the cube-result had a means measure; probably better to just move it up the one level to the factory, and then any `_BaseCubeMeans` subclass can rely on its `._means` instance variable to contain a valid cube-means array.

src/cr/cube/cubepart.py
```diff
@@ -927,6 +922,15 @@ def smoothed_dimension_dict(self):
         # from the matrix later on.
         return self._columns_dimension._dimension_dict
 
+    @lazyproperty
+    def sum(self):
```
scanny, line 926: This should really be `sums`, shouldn't it? All of the other measures are plural, and this is a measure, and in general it contains multiple values, doesn't it? Pretty sure this should be `sums`, and all the related measure bits below would be plural too, like `_Sums`, `_BaseCubeSums`, etc.

tests/integration/test_cubepart.py
```diff
@@ -318,25 +330,31 @@ def it_provides_values_for_cat_hs_x_mr(self):
...
-        np.testing.assert_array_almost_equal(slice_.rows_margin, np.array([np.nan]))
-        np.testing.assert_array_almost_equal(
-            slice_.columns_margin, np.array([np.nan] * len(slice_.counts[0, :]))
+        # This fixture has both cube_counts and cube_means measure, for this reason
+        # both measures are available at cubepart level.
+        assert slice_.means == pytest.approx(
+            np.array([[24.4393575, 37.3212274, np.nan, 55.4857195, 73.0242765]]),
+            nan_ok=True,
         )
+        assert slice_.counts == pytest.approx(np.array([[189, 395, 584, 606, 310]]))
```
scanny, line 339: I think it's well to establish these asserts being in alphabetical order by property name, like all the other ones. It might seem like there is a logical ordering possible, now, while it's small, but as more property-value asserts are added that just comes undone. Better to be consistent with all the other tests like this from the start and then folks know what to expect and can find what they're looking for quickly.

tests/integration/test_stripe.py
```diff
@@ -62,7 +62,9 @@ def it_provides_values_for_univariate_cat_means(self):
         assert assembler.scale_stddev is None
         assert assembler.scale_stderr is None
         assert assembler.table_base_range == pytest.approx([661, 661])
-        assert assembler.table_margin_range == pytest.approx([86.4992, 86.4992])
+        # for a cube with numeric measure like mean, table margin and table base are the
+        # same because they are both calculated on the counts.
```
scanny, line 66: do you mean "... are both calculated on the (unweighted) valid-counts."?

scanny (changes requested):
Ok, I think we're almost there. I made it all the way through, so just another quick look once you've made all your changes and I think this one will be ready to go :)

src/cr/cube/cube.py
ernestoarbitrio: ok reverted

tests/integration/test_stripe.py
ernestoarbitrio: yup

src/cr/cube/cube.py
```diff
@@ -380,7 +408,7 @@ def unweighted_counts(self):
     @lazyproperty
     def valid_counts(self):
         """ndarray of valid counts, valid elements only."""
-        valid_counts = self._measure.valid_counts
+        valid_counts = self._base_measure.valid_counts
         if valid_counts.any():
             return valid_counts[self._valid_idxs]
         return np.empty(0)
```
scanny, line 414: As far as I can tell, `.valid_counts` should not be an interface method because it has no callers outside this class.

There is a broader problem though, because we shouldn't need to construct `_BaseMeasure` directly just to redundantly access a shared measure property (redundant in the sense it will be independently computed in a `_WeightedCounts` instance and perhaps an `_UnweightedCounts` as well.

What we need to do instead is this:

1. Make `_ValidCounts` a full-fledged measure of its own, that knows how to find its JSON values in the payload and shares the `.raw_cube_array` property from `_BaseMeasure` etc. `_ValidCounts` becomes the authority on what these values are and whether or not they exist in the payload; its `._flat_values` property returns `None` if the payload does not include a valid-counts measure.

2. All the logic for substituting valid-counts for unweighted and weighted counts comes out of the `._flat_values` properties of those classes and moves into `Cube.unweighted_counts` and `Cube.weighted_counts`, which become something like:

   ```python
    @lazyproperty
    def unweighted_counts(self):
        unweighted_counts = (
            self._measures.valid_counts
            if self._measures.valid_counts is not None
            else self._measures.unweighted_counts
        )
        return unweighted_counts.raw_cube_array[self._valid_idxs]
   ```

   And `Cube.weighted_counts` likewise.

3. `Cube.valid_counts_summary` uses `self._measures.valid_counts` in a similar way to get what it needs.

src/cr/cube/cube.py
```diff
@@ -407,6 +435,11 @@ def _all_dimensions(self):
         """
         return AllDimensions(dimension_dicts=self._cube_dict["result"]["dimensions"])
 
+    @lazyproperty
+    def _base_measure(self):
+        """_BaseMeasure object for this cube."""
+        return _BaseMeasure(self._cube_dict, self._all_dimensions, self._cube_idx_arg)
```
scanny, line 441: Then this property should go away.

src/cr/cube/cube.py
```diff
@@ -231,13 +233,25 @@ def __repr__(self):
         except Exception:
             return super(Cube, self).__repr__()
 
+    @lazyproperty
+    def available_measures(self):
+        """frozenset of available measures in the cube response."""
+        return frozenset(
+            self._cube_response.get("result", {}).get("measures", {}).keys()
+        )
```
scanny, line 241: This needs to be a collection of `Enum` members, not a collection of `str` values. The interface we want to present to callers is:

```python
from cr.cube.enums import MEASURE

if MEASURE.COUNTS in cube.available_measures:
    # --- do the counts thing ---
```

Otherwise we have coupled our clients to the internal schema we use, which is bad juju.

src/cr/cube/cube.py
scanny, line 414: This use of an empty array to signal "no value" is improper, because it is an "in-band" value, i.e. it is the same type as a valid value, and who's to say that one day (or even already) an empty array might be a valid return value.

When signaling the failure case of an optional value, use an out-of-band value, which is exactly what `None` is. It's also easier and faster to check, with `if valid_counts is None:`.

I'm pretty sure these lines are going away anyway, but the principle applies broadly, not just in this instance.

src/cr/cube/cube.py
```diff
@@ -457,36 +490,34 @@ def _is_single_filter_col_cube(self):
         return self._cube_dict["result"].get("is_single_col_cube", False)
 
     @lazyproperty
-    def _mean_references(self):
-        """Dict of mean references, tipically for numeric arrays."""
+    def _numeric_measure_references(self):
+        """Dict of numeric measure references, tipically for numeric means or sum."""
```
scanny, line 494: "typically"

src/cr/cube/cube.py
```diff
@@ -457,36 +490,34 @@ def _is_single_filter_col_cube(self):
...
         """
-        return (
-            self._measures.means
-            if self._measures.means is not None
-            else self._measures.weighted_counts
-            if self._measures.weighted_counts is not None
-            else self._measures.unweighted_counts
-        )
+        numeric_measures = [measure.value for measure in NUMERIC_MEASURES]
+        return list(self.available_measures.intersection(numeric_measures))
```
scanny, line 520: As mentioned a few times, we should not be trafficking in str values. The type of this should be a collection of MEASURE enum members.

Also, this should be immutable. No @lazyproperty should be mutable, otherwise it can return different values on different calls. `frozenset` may be convenient, that has ready set operations and can be iterated, it just can't be directly indexed, so you would need `list(self._available_numeric_measures)[0]` to get a random one, like up above in `._numeric_mean_subvariables`.

Also, I think this should be `._available_numeric_measures`, to be more descriptive. It is not _all_ numeric values, just the ones available in this cube-result.

The `QUANTITY_OF_INTEREST_MEASURES` reference is outdated.

If you need the str value elsewhere, like up on line 509, you can just use the `.value` property on the enum member, so that line would become:

```python
metadata = cube_measures.get(self._available_numeric_measures[0].value, {}).get("metadata", {})
```

src/cr/cube/cube.py
```diff
@@ -500,20 +531,20 @@ def _measures(self):
     @lazyproperty
     def _numeric_array_dimension(self):
         """Rows dimension object according to the mean subvariables."""
```
scanny, line 533: should this docstring also say "numeric-measure" rather than "mean"?

src/cr/cube/cube.py
```diff
@@ -737,28 +778,47 @@ def _flat_values(self):
...
         if self._result is None:
             return None
+        return np.array(
+            tuple(np.nan if type(x) is dict else x for x in self._result["data"]),
+            dtype=np.float64,
+        ).flatten()
+
+    @lazyproperty
+    def _result(self):
+        return self._cube_dict.get("result", {}).get("measures", {}).get("mean")
```
scanny, line 792: docstring. Also, this seems a poor name. I would expect `._result` to return `self._cube_dict["result"]`. Maybe something like `._measure_values`.

scanny: or maybe `._measure_payload`, since it seems it is a dict rather than a list of values.

src/cr/cube/cubepart.py
```diff
@@ -923,6 +923,21 @@ def smoothed_dimension_dict(self):
         # from the matrix later on.
         return self._columns_dimension._dimension_dict
 
+    @lazyproperty
+    def sums(self):
+        """2D optional np.float64 ndarray of sum value for each table cell.
+
+        Cell value is `np.nan` for each cell corresponding to an inserted subtotal
```
scanny, line 930: Are we sure that a subtotal of `sums` should be `np.nan`? I'm inclined to think not, and that they _do_ add, unlike `means`.

src/cr/cube/matrix/cubemeasure.py
```diff
@@ -68,6 +78,164 @@ def _slice_idx_expr(cls, cube, slice_idx):
...
+            _MrXMrCubeMeans
+            if dimension_types == (DT.MR, DT.MR)
+            else _MrXCatCubeMeans
+            if dimension_types[0] == DT.MR
+            else _CatXMrCubeMeans
+            if dimension_types[1] == DT.MR
+            else _CatXCatCubeMeans
+        )
+        if cube.means is None:
+            raise ValueError("cube-result does not contain cube-means measure")
```
scanny, line 105: This guard should probably go right up top in this method since none of the rest makes any sense if there is no `cube.means`.

src/cr/cube/matrix/cubemeasure.py
```diff
@@ -925,40 +1089,15 @@ def _array_type_std_res(self, counts, total, rowsum, colsum):
...
         return (
-            _MrXMrMatrix
+            _NumArrayXMrMatrix
+            if dimension_types == (DT.NUM_ARRAY, DT.MR)
+            else _NumArrayXCatMatrix
+            if dimension_types[0] == DT.NUM_ARRAY
+            else _MrXMrMatrix
             if dimension_types == (DT.MR, DT.MR)
             else _MrXCatMatrix
             if dimension_types[0] == DT.MR
```
scanny, line 1103: This is only called once and should be inlined. It made sense to be separate back when the factories for means and counts etc. were all together, but not anymore.

src/cr/cube/matrix/cubemeasure.py
```diff
@@ -968,7 +1107,7 @@ def _regular_matrix_class(dimension_types):
         )
 
     @staticmethod
-    def _regular_matrix_counts_slice(cube, slice_idx):
+    def _regular_matrix_slice(cube, slice_idx):
```
scanny, line 1110: The `regular_` prefix no longer makes any sense here because there is no "non-regular" counterpart. Also, this is not a "matrix-slice", it is the "slice-expression" that digs out the right portion of the overall cube values. So something like "._cube_slice_expression" is more appropriate.

src/cr/cube/matrix/cubemeasure.py
```diff
@@ -1001,8 +1134,12 @@ def _sliced_counts(cls, cube, slice_idx):
...
         tabulation.
         """
-        i = cls._regular_matrix_counts_slice(cube, slice_idx)
-        return (cube.counts[i], cube.unweighted_counts[i], cube.counts_with_missings[i])
+        i = cls._regular_matrix_slice(cube, slice_idx)
+        return (
+            cube.counts[i],
+            cube.unweighted_counts[i],
+            cube.counts_with_missings[i],
+        )
```
scanny, line 1142: I find `i` is misleading here, making one think it is an `int` as it commonly would be. Maybe `slice_expr = cls._cube_slice_expression(cube, slice_idx)` instead.

src/cr/cube/matrix/cubemeasure.py
```diff
@@ -1719,73 +1856,7 @@ def _table_proportion_variances(self):
...
 
 
-# === LEGACY MEANS MATRIX OBJECTS ===
-
-
-class _CatXCatMeansMatrix(_CatXCatMatrix):
-    """CAT_X_CAT matrix for means measure.
-
-    A means matrix has an array of mean values instead of a `counts` array.
-    """
```
scanny, line 1729: Nice to see these in red ink :)

src/cr/cube/matrix/measure.py
```diff
@@ -524,6 +545,15 @@ def _subtotal_rows(self):
...
 
 
+class _Sums(_BaseSecondOrderMeasure):
+    """Provides the sum measure for a matrix."""
+
+    @lazyproperty
+    def blocks(self):
+        """2D array of the four 2D "blocks" making up this measure."""
+        return NanSubtotals.blocks(self._cube_measures.cube_sum.sums, self._dimensions)
+
```
scanny, line 555: Not sure about the `NanSubtotals` bit here, better check with Mike.

scanny (changes requested):
Okay @ernestoarbitrio, I made it all the way through, unfortunately I had quite a few new comments, as you've seen, but hoping again that we're pretty much there. I'll take another look on Monday and we'll see if we can push.

src/cr/cube/cube.py
ernestoarbitrio: I leave it cause i though it was ok for you after that this property was introduced in the greg's PR 🤷🏽‍♂️  ... I'll change it

src/cr/cube/cubepart.py
ernestoarbitrio: u right

src/cr/cube/matrix/measure.py
ernestoarbitrio: yes sums should add

src/cr/cube/cube.py
```diff
@@ -791,8 +855,10 @@ class _WeightedCountMeasure(_BaseMeasure):
     def _flat_values(self):
         """Optional 1D np.ndarray of np.float64 numeric counts after weighting."""
         unweighted_counts = self._cube_dict["result"]["counts"]
-        weighted_counts = self._cube_dict["result"]["measures"]["count"]["data"]
-        if unweighted_counts == weighted_counts:
+        weighted_counts = (
+            self._cube_dict["result"]["measures"].get("count", {}).get("data")
+        )
+        if unweighted_counts == weighted_counts or weighted_counts is None:
```
scanny, line 861: This doesn't seem right. The presence of flat values for weighted-counts have nothing to do with unweighted counts. If a `result.measures.count` payload item appears, we return it here. Otherwise we return None. It containing equivalent values to `result.count` does not change the fact that these are present.

Basically, if `cube-count` appears in the query, then this property has a non-None value.

src/cr/cube/dimension.py
```diff
@@ -57,6 +57,17 @@ def apparent_dimensions(self):
...
+    @lazyproperty
+    def require_transposition(self):
+        """True if dimensions need transposition, False otherwise.
+
+        Transposition is needed when one of the dimension type is a NUM_ARRAY and the
+        dimensions are more than 1. This business rule needs to drive correctly tabbook
+        and deck exports when a numeric array dimension is expressed.
+        """
+        dimension_types = [d.dimension_type for d in self.apparent_dimensions]
+        return len(self.apparent_dimensions) >= 2 and DT.NUM_ARRAY in dimension_types
```
scanny, line 69: If this is a temporary hack while we wait for `dim_order`, I'm a little uncomfortable with allowing the hack to spread to the `cr.cube.dimension` module where we're likely to forget to remove it when the time comes. I think this would be better appearing only in `cr.cube.cube.Cube` and marked as a hack and when it is expected to come out.

src/cr/cube/cube.py
```diff
@@ -543,12 +593,13 @@ def _slice_idxs(self):
...
     @lazyproperty
     def _valid_idxs(self):
+        """Tuple of int64 ndarrays of the valid elements idx for each dimension."""
         valid_idxs = np.ix_(
             *tuple(d.valid_elements.element_idxs for d in self._all_dimensions)
         )
 
         def _reshape_idxs(valid_indices):
-            if self._measure.requires_array_transposition:
+            if self._all_dimensions.require_transposition:
```
scanny, line 602: Since this is a temporary hack while we wait for `dim_order`, I think it's a mistake to pollute `AllDimensions` when we can easily contain the hackery to this module.

Just create `Cube.requires_array_transposition` (even if it is redundant with `_BaseMeasure.requires_array_transposition`) and mark it as a hack with a comment and when it is expected to go away.

src/cr/cube/cube.py
```diff
@@ -675,30 +748,6 @@ def raw_cube_array(self):
...
-    @lazyproperty
-    def requires_array_transposition(self):
-        """True if raw cube array needs transposition, False otherwise.
-
-        When one of the dimension type is a NUM_ARRAY, its cube_array values have to be
-        transposed. This business rule needs to drive correctly tabbook and deck exports
-        when a numeric array dimension is expressed.
-        """
-        dimension_types = [d.dimension_type for d in self._all_dimensions]
-        return len(self._all_dimensions) >= 2 and DT.NUM_ARRAY in dimension_types
```
scanny, line 687: This should stay here rather than spreading the hackery to `AllDimensions`.

src/cr/cube/enums.py
```diff
@@ -75,7 +75,21 @@ class MEASURE(enum.Enum):
...
+
+
+class CUBE_MEASURE(enum.Enum):
+    """Enumerated values representing cube measures."""
+
+    COUNT = "count"
+    MEAN = "mean"
+    STDDEV = "stddev"
+    SUM = "sum"
+    VALID_COUNT_UNWEIGHTED = "valid_count_unweighted"
```
scanny, line 92: Nice.

tests/integration/test_cube.py
```diff
@@ -104,12 +104,7 @@ def it_provides_pruned_array_for_CA_CAT_x_CA_SUBVAR(self):
     def it_provides_valid_counts_for_NUM_ARRAY_GROUPED_BY_CAT(self):
         cube = Cube(NA.NUM_ARR_MEANS_GROUPED_BY_CAT)
 
-        np.testing.assert_array_equal(cube.valid_counts, [[3, 2], [3, 1], [1, 1]])
-
-    def and_it_returns_empty_array_if_valid_counts_are_not_available(self):
-        cube = Cube(CR.CAT_X_CAT)
-
-        np.testing.assert_array_equal(cube.valid_counts, [])
+        np.testing.assert_array_equal(cube.counts, [[3, 2], [3, 1], [1, 1]])
```
scanny, line 108: use `assert` syntax whenever touching a test.

scanny (changes requested):
Okay, we're getting there. A couple new issues. I can take another look today if you're done in time.

src/cr/cube/cube.py
```diff
@@ -245,9 +245,11 @@ def counts(self):
 
     @lazyproperty
     def counts_with_missings(self):
-        """ndarray of weighted or unweighted cube counts."""
+        """ndarray of weighted, unweighted or valid cube counts."""
```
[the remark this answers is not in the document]
ernestoarbitrio: col index depends on counts for sure, but valid counts ATM (by spec) are only available for numeric measures whene col_index has no sense to be calculated

src/cr/cube/cube.py
ernestoarbitrio: This is always been this way. If you take a look on `master` it has the check on the unweighted counts either.

src/cr/cube/dimension.py
ernestoarbitrio: Considering that this property is needed even in `_BaseMeasure` in the shape, i'd prefer to have it in something where is effectively related. We are transposing the dimensions so... dimensions would be an appropriate module.

src/cr/cube/cube.py
ernestoarbitrio: Basically, if wcounts and ucounts are the same means that counts are not weighted and so this check i think

ernestoarbitrio: Ok i investigate on that and the problem is that all the fixtures have `result.measures.count` so this means that all the cube have weighted counts. This is no longer optional and that's not true. So I'd prefer to keep it this way until we figure out what the best thing to do.

src/cr/cube/dimension.py
```diff
@@ -57,6 +57,30 @@ def apparent_dimensions(self):
...
+        NUM_ARR_X_MR_SUBVAR_X_MR_CAT the order should be (1,2,0) that is basically
+        swapping the MR (2 dimensions) with the NUM_ARRAY dimension. In case of 2
+        dimensions the dimension order correspond simpy to the reverse of the original
+        dimension order.
+        """
+        # NOTE: this is a temporary hack that goes away when we introduce the dim_order
+        # concept. We should receive the actual order directly in the cube_response.
+        # So, all this logic will be deleted.
+        dimension_types = tuple(d.dimension_type for d in self._dimensions)
+        dim_order = tuple(idx for idx in range(len(self._dimensions)))
```
scanny, line 75: can just be `tuple(range(len(self._dimensions)))`

scanny (approved):
LGTM :) Just one small tweak and then it can go. Nice job encapsulating the `dim-order` so the temporary hack is painless to the rest of the code and everything is already set up to use dim-order once it arrives :)

src/cr/cube/dimension.py
ernestoarbitrio: yes ... sorry i changed it but not pushed ... damn ... DONE now

PR #253, opened 2021-02-19
Subdiff v2 176856023

gergness:
Adds the subtotal "difference" insertions where some categories can be subtracted off. Builds off of Steve's new-architecture-strand branch.


tests/integration/test_headers_and_subtotals.py
```diff
@@ -374,6 +375,247 @@ def test_subtotals_proportions_2d_cube_with_hs_on_two_dim_by_cell(self):
         ]
         np.testing.assert_almost_equal(slice_.table_proportions, expected)
 
+    def test_subtotal_diffs_1d(self):
+        strand = Cube(CR.CAT_SUBDIFF).partitions[0]
+        counts = strand.counts
+        np.testing.assert_array_equal(counts, (81, 211, 212, 131, 79, 51))
```
slobodan-ilic, line 381: The issue with this test is that I can't immediately see _what_ is it that I'm supposed to be seeing :) So, if it's the insertion 211 - 79 - 51, that should be stated in the test explicitly. Same for all the other expectations. Also, if the numbers are round, it's better to use the `assert counts == [x, y, z]` than `np.testing.assert_array_equal`, because it's the go-to way of the crunchdb team, and we need to fix all the cube tests to use it as well.

slobodan-ilic: An even better approach would be to pass the transformations explicitly, when creating the cube. But I've forgotten if that's possible for insertions?

tests/integration/test_headers_and_subtotals.py
```diff
@@ -374,6 +375,247 @@ def test_subtotals_proportions_2d_cube_with_hs_on_two_dim_by_cell(self):
...
+            [0.0904977, 0.27, 0.1491228, 0.2261146, 0.3870968, 0.0487805, 0.0672269],
+            [0.0791855, 0.22, 0.1842105, 0.2070064, 0.0322581, 0.0406504, 0.0336134],
+            [0.0271493, 0.09, 0.122807, 0.1019108, 0.1182796, 0.0243902, 0.0252101],
+            [0.1063348, 0.31, 0.3070175, 0.3089172, 0.1505376, 0.0650407, 0.0588235],
+        ]
+        np.testing.assert_almost_equal(slice_.column_proportions, expected)
+
+    def test_subtotals_proportions_2d_cube_with_diffs_and_totals_on_both_by_row(self):
+        slice_ = Cube(CR.CAT_SUBTOT_SUBDIFF_X_CAT_SUBTOT_SUBDIFF).partitions[0]
+        expected = [
```
slobodan-ilic, line 552: These large expectations tend to render the code unreadable. Could you put them into a file? There are examples of it throughout the test suite.

slobodan-ilic: In general I think that a criterion should be "if it's more than a single screen hight - move it to a file". But I think @scanny has an even better rule, but I don't remember it now.

tests/integration/test_headers_and_subtotals.py
```diff
@@ -3810,3 +3811,540 @@ def it_provide_residual_test_stats_including_hs(self):
...
+        np.testing.assert_equal(slice_.zscores[:, 0], np.full(6, np.nan))
+
+    def test_cat_subtot_subdiff_x_cat_subtot_subdiff_counts(
+        self, row_col_subtot_subdiff_insertions
+    ):
+        slice_ = Cube(
+            CR.CAT_SUBDIFF_X_CAT_SUBDIFF, transforms=row_col_subtot_subdiff_insertions
+        ).partitions[0]
+        # --- Subdiff in row 0 equals row=1 - (row=4 + row=5) ---
+        np.testing.assert_array_equal(
```
gergness, line 3973: I'm fighting with `black` formatting here. There's gotta be a better way, right? But black kept changing to this.

tests/integration/test_headers_and_subtotals.py
```diff
@@ -3810,3 +3811,540 @@ def it_provide_residual_test_stats_including_hs(self):
...
+            ],
+        )
+
+    def test_cat_subdiff_x_cat_subdiff_table_prop(self, row_col_subdiff_insertions):
+        slice_ = Cube(
+            CR.CAT_SUBDIFF_X_CAT_SUBDIFF, transforms=row_col_subdiff_insertions
+        ).partitions[0]
+
+        np.testing.assert_almost_equal(
+            slice_.table_proportions, slice_.counts / slice_.table_base
```
gergness, line 3922: I notice these `/` aren't python2 compatible. Advice on what to change to?
https://github.com/Crunch-io/crunch-cube/pull/253/checks?check_run_id=1942716810#step:5:142

tests/integration/test_headers_and_subtotals.py
gergness: nevermind, I think I can change from `count[1, ]` to `count[1, :]` and it is better

src/cr/cube/matrix/subtotals.py
```diff
@@ -209,6 +209,13 @@ class SumDiffSubtotals(SumSubtotals):
 
     def _intersection(self, row_subtotal, column_subtotal):
         """Sum and Diff for this row/column subtotal intersection."""
+        # --- Intersections of subtotal differences are undefined ---
+        if (
+            len(column_subtotal.subtrahend_idxs) > 0
+            and len(row_subtotal.subtrahend_idxs) > 0
+        ):
+            return np.nan
```
gergness, line 217: This doesn't work because unweighted counts are currently stored as integers which don't have a `nan`. Fix should probably be to make unweighted counts as floats

tests/integration/test_headers_and_subtotals.py
gergness: fixed via `from __future__ import division`

src/cr/cube/matrix/subtotals.py
gergness: Next commit is a way to fix, but I'm not sure it's the correct way.

tests/unit/matrix/test_measure.py
```diff
@@ -1064,7 +1087,7 @@ def it_computes_its_blocks_to_help(self, request, dimensions_):
             "_unweighted_cube_counts",
             return_value=unweighted_cube_counts_,
         )
-        SumSubtotals_ = class_mock(request, "cr.cube.matrix.measure.SumSubtotals")
+        SumSubtotals_ = class_mock(request, "cr.cube.matrix.measure.SumDiffSubtotals")
```
scanny, line 1090: This mock's name should also change, to `SumDiffSubtotals_` in this case. The idea is you should be able to tell everything important about the mock just from it's name, in this case, which class it is a mock _of_.

tests/fixtures/cat-subdiff-x-cat-subdiff.json
```diff
@@ -0,0 +1,292 @@
+{
+    "element": "shoji:view",
+    "self": "https://app.crunch.io/api/datasets/95ba1b/cube/?query=%7B%22dimensions%22%3A%5B%7B%22variable%22%3A%22https%3A%2F%2Fapp.crunch.io%2Fapi%2Fdatasets%2F95ba1b%2Fvariables%2F000000%2F%22%7D%2C%7B%22variable%22%3A%22https%3A%2F%2Fapp.crunch.io%2Fapi%2Fdatasets%2F95ba1b%2Fvariables%2F000001%2F%22%7D%5D%2C%22measures%22%3A%7B%22count%22%3A%7B%22function%22%3A%22cube_count%22%2C%22args%22%3A%5B%5D%7D%7D%2C%22weight%22%3Anull%7D&filter=%7B%7D",
+    "value": {
```
scanny, line 4: I think cat-x-cat can be used instead of this one, can't it? Of course the expectation would change, but in general there is nothing substantially different about this variant as far as I can tell.

tests/fixtures/cat-subdiff-x-mr.json
```diff
@@ -0,0 +1,426 @@
+{
+    "element": "shoji:view",
+    "self": "https://app.crunch.io/api/datasets/6c245c/cube/?query=%7B%22dimensions%22%3A%5B%7B%22variable%22%3A%22https%3A%2F%2Fapp.crunch.io%2Fapi%2Fdatasets%2F6c245c%2Fvariables%2F000001%2F%22%7D%2C%7B%22function%22%3A%22dimension%22%2C%22args%22%3A%5B%7B%22function%22%3A%22as_selected%22%2C%22args%22%3A%5B%7B%22variable%22%3A%22https%3A%2F%2Fapp.crunch.io%2Fapi%2Fdatasets%2F6c245c%2Fvariables%2Fa4260e%2F%22%7D%5D%7D%2C%7B%22value%22%3A%22subvariables%22%7D%5D%7D%2C%7B%22function%22%3A%22as_selected%22%2C%22args%22%3A%5B%7B%22variable%22%3A%22https%3A%2F%2Fapp.crunch.io%2Fapi%2Fdatasets%2F6c245c%2Fvariables%2Fa4260e%2F%22%7D%5D%7D%5D%2C%22measures%22%3A%7B%22count%22%3A%7B%22function%22%3A%22cube_count%22%2C%22args%22%3A%5B%5D%7D%7D%2C%22weight%22%3Anull%7D&filter=%7B%7D",
+    "value": {
```
scanny, line 4: CAT_X_MR may get this job done.

tests/fixtures/cat-subdiff.json
```diff
@@ -0,0 +1,167 @@
+{
+    "element": "shoji:view",
+    "self": "https://app.crunch.io/api/datasets/95ba1b/cube/?query=%7B%22dimensions%22%3A%5B%7B%22variable%22%3A%22https%3A%2F%2Fapp.crunch.io%2Fapi%2Fdatasets%2F95ba1b%2Fvariables%2F000000%2F%22%7D%5D%2C%22measures%22%3A%7B%22count%22%3A%7B%22function%22%3A%22cube_count%22%2C%22args%22%3A%5B%5D%7D%7D%2C%22weight%22%3Anull%7D&filter=%7B%7D",
+    "value": {
+        "query": {
```
scanny, line 5: Maybe just call this one `cat.json` so it can be used in non-subdiff contexts.

tests/integration/test_headers_and_subtotals.py
```diff
@@ -3810,3 +3813,457 @@ def it_provide_residual_test_stats_including_hs(self):
...
+
+
+class DescribeIntegrated_SubtotalDifferences(object):
+    """TDD driver(s) for Subtotal Difference insertions."""
+
+    def test_cat_subdiff_count(self, row_subdiff_insertions):
+        strand = Cube(CR.CAT_SUBDIFF, transforms=row_subdiff_insertions).partitions[0]
+        counts = strand.counts
+        # --- Subdiff in position 0 equals id=1 - (id=4 + id=5) ---
+        np.testing.assert_almost_equal(counts[0], counts[1] - (counts[4] + counts[5]))
```
scanny, line 3825: Just over the past couple weeks we've moved away from `np.testing.*` matchers like `.assert_almost_equal()`. Slobodan initiated the move, but I agree it's more readable and generally more in line with our overall use of `pytest`.

There are two variants that I've encountered, depending on whether the expectation is integer or float in base type.

`.counts` is `float` I believe, so this one would be:

```
assert counts[0] == pytest.approx(counts[1] - (counts[4] + counts[5]))
```

Just as I'm looking at it more closely here, I'd be inclined to make it:

```
assert counts[0] == pytest.approx(142.7) (or whatever the right answer is)
```

rather than duplicating the logic in the test.

The other case, for integer results (which have a solid equivalence comparison) is:

```
assert bases.tolist() == [1, 2, 3, 4, 5]
```

Using these two, you can cover almost all cases, however when NaNs are involved, you need:

```
assert zscores == pytest.approx([1.1, 2.2, np.nan, 4.4], nan_ok=True)
```

tests/integration/test_headers_and_subtotals.py
```diff
@@ -3810,3 +3813,457 @@ def it_provide_residual_test_stats_including_hs(self):
...
+
+    def test_cat_subdiff_prop(self, row_subdiff_insertions):
+        strand = Cube(CR.CAT_SUBDIFF, transforms=row_subdiff_insertions).partitions[0]
+        table_proportions = strand.table_proportions
+        # --- Subdiff in position 0 equals id=1 - (id=4 + id=5) ---
+        np.testing.assert_almost_equal(
+            table_proportions[0],
+            table_proportions[1] - (table_proportions[4] + table_proportions[5]),
```
scanny, line 3833: Following on to part of the prior comment, I think if you use actual values and avoid repeating the computation in the test that you can combine each set of tests that uses the same fixture. This improves performance and I expect improves the readability as well because it's a lot shorter. So it would be something like:

```python
    def it_computes_subdiff_subtotals_for_various_measures(self, row_subdiff_insertions):
        strand = Cube(CR.CAT_SUBDIFF, transforms=row_subdiff_insertions).partitions[0]

        assert strand.table_proportions[0] == pytest.approx(0.3725221)
        assert strand.row_proportions ...
        assert strand.column_proportions ...
        ...
```

scanny (changes requested):
Okay, here's a first pass to get you started. Overall I think it looks great, just a few finer points on the tests mostly. I ran out of time before getting to most of the unit tests, but I can spend more time on those on the next pass.

tests/integration/test_headers_and_subtotals.py
gergness: For my curiosity, can you describe why it's more readable? Just the fact that it uses the `==` infix operator? To me it seems a shame to lose all the niceties of np's wisdom of numeric comparisons.

tests/fixtures/cat-subdiff-x-cat-subdiff.json
gergness: Yeah, I kind of knew you'd want to use it. It was nice to have a file with real interpretations for working with it, but I can change before merging.

src/cr/cube/cubepart.py
```diff
@@ -1493,7 +1487,7 @@ def unweighted_bases(self):
 
     @lazyproperty
     def unweighted_counts(self):
-        """1D np.int64 ndarray of unweighted count for each row of stripe."""
+        """tuple of int unweighted count for each row of stripe."""
```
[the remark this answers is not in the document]
gergness: Oops, not sure how that changed, I don't think I did that intentionally, probably a rebase artifact.

tests/integration/test_headers_and_subtotals.py
gergness: This kind of opposes advice Slobodon had, where I originally had numbers, but he argued that it was hard to interpret the meaning of the test because it was just something equals a random number. Do you have advice on how to balance these two concerns?

tests/fixtures/ca-subdiff.json
```diff
@@ -0,0 +1,330 @@
+{
+    "element": "shoji:view",
+    "self": "https://app.crunch.io/api/datasets/6c245c/cube/?query=%7B%22dimensions%22%3A%5B%7B%22function%22%3A%22dimension%22%2C%22args%22%3A%5B%7B%22variable%22%3A%22https%3A%2F%2Fapp.crunch.io%2Fapi%2Fdatasets%2F6c245c%2Fvariables%2Ff054e4%2F%22%7D%2C%7B%22value%22%3A%22subvariables%22%7D%5D%7D%2C%7B%22variable%22%3A%22https%3A%2F%2Fapp.crunch.io%2Fapi%2Fdatasets%2F6c245c%2Fvariables%2Ff054e4%2F%22%7D%5D%2C%22measures%22%3A%7B%22count%22%3A%7B%22function%22%3A%22cube_count%22%2C%22args%22%3A%5B%5D%7D%7D%2C%22weight%22%3Anull%7D&filter=%7B%7D",
+    "value": {
+        "query": {
```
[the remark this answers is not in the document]
gergness: Yeah, I kind of knew this was coming. For dev purposes I like having the file with real interpretations (and that I have control of), but I can change to use existing fixtures before finalizing.

src/cr/cube/matrix/subtotals.py
scanny: Yeah, it's a good question. There are two approaches I can think of:

1. Adopt the rule that _all_ numeric measure values (but not idxs like `.inserted_column_idxs`) are float, and remove that broadly docstring-documented and otherwise implemented distinction between int and float values. I kind of think this is the right approach, but it would entail a lot of small changes, including right down to the cube-measure level where we would "cast" the ints we get for unweighted counts to floats.

2. Just remove the `dtype` setting here in subtotals and let numpy determine what type to make things and some are ints and some are floats and we don't pay too much attention. This is easy, but it feels sloppy and the kind of thing that would come to bite us later. As a general principle, we'd like to have a consistent interface, which includes consistent value types.

Let me know what you think and maybe we can discuss at stand-up.

tests/integration/test_headers_and_subtotals.py
scanny: Well, readability is at least partially subjective, but from my own perspective, I'd say that it's the fact that all assertions declare themselves right up front with an `assert ` keyword. A test naturally has three blocks, the fixturing, the call, and the assertions, and having them stand out like that makes them easier to detect while scanning (which is (or should be) much of what we call "reading" code).

@slobodan-ilic might have a different view he could share with us.

tests/fixtures/ca-subdiff.json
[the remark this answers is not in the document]
scanny: There's no problem with adding this fixture if you feel it is distinctive, just name it `ca-subvar-x-ca-cat.json` since we don't have one of those yet.

tests/integration/test_headers_and_subtotals.py
scanny: Well, all the integration-test expectations that we have contain "random" numbers, so I don't see how this is different. And duplicating the logic is not testing, it's duplicating the logic; if something went wrong in table-proportions, for example, you could get a passing test (because the logic is the same) for code that produces the wrong answer.

So I'd say using actual values for actuals is the broadly established convention, and with cause. If you wanted to make them somewhat easier to reason about, you can manipulate the fixture values so the actuals come out to some round number or something, like `8 - 2 + 3 = 9` or whatever. But a test is not to be understood in isolation, but paired with the unit (or integrated bit).

Also, I'm not sure how important it is to be able to validate actual values by inspection. Either the author got it right or you're in there modifying the test and need to re-reason about the result yourself. The actual value, once written, is just for regression purposes; to make sure the result didn't _change_ since it was validated by the author. I can't remember having ever checked whether the right values appeared in an expectation written by someone else. I think any effort in that regard is better spent in validating the final result, like Joe does.

gergness: Sounds good, thanks for explaining, I can feel lots of trade-offs here and so trying to get good guide posts.

Does this look good? 
https://github.com/Crunch-io/crunch-cube/blob/363f7dc4aded1ec32073266c373918ecbef3014f/tests/integration/test_headers_and_subtotals.py#L3842-L3902

I worry that it's past some point of too many `assert`s to be meaningful, but can't tell if that just doesn't matter because we're not expecting to have to read them one at a time.

scanny: I think this looks good. It's true there are a lot of asserts, but this is an integration test, not a unit test. Also, _Slice is a _value_ object (mostly at least) and to me, this test says: "For the case CAT_X_CAT with these transforms, `slice.values == X`, it's just that values is a collection where each member is distinctly named. In any case, I find it an entirely legitimate approach for this kind of test and there are several other examples of this approach throughout the integration tests.

One important characteristic of this approach is that it avoids a lot of little tests that all load the same fixture, but otherwise don't show that they are closely related. Of course there is the performance penalty involved in that as well, and when you multiply that small difference across a number of integration tests, the test runtime can creep up. I run the tests continuously during development such that they run on every save, and keeping the suite runtime in the single-digit seconds keeps the development tempo snappy, for me at least.

src/cr/cube/matrix/subtotals.py
gergness: Okay, I made the changes discussed (`_UnweightedCountMeasure._flat_values` now converts to ndarrays of floats). I also took out the `_dtype` methods in Subtotals in `matrix.insertions.py` because it didn't seem useful now that everything should be a float.

I have not updated docstrings. I'm a little worried about making that many changes while you, Ernesto and I all need to rebase on each other. Do you think I need to do that now?

src/cr/cube/cube.py
```diff
@@ -773,9 +773,21 @@ def _flat_values(self):
...
         """
-        return tuple(
-            np.nan if type(x) is dict else x
-            for x in self._cube_dict["result"]["measures"]["mean"]["data"]
+
+        def dict_to_nan_recursive(x):
+            if type(x) is dict:
+                return np.nan
+            elif type(x) is list:
+                return [dict_to_nan_recursive(y) for y in x]
```
gergness, line 781: @scanny - is this the right way to do this?

The problem I hit is that because of numeric arrays, this can have two different forms:

Mean of non-array: list of floats or dictionaries (indicating missing values)
Mean of array: **list of** lists of floats or dictionaries

src/cr/cube/cube.py
```diff
@@ -798,9 +810,6 @@ def _flat_values(self):
         if valid_counts:
             return np.array(valid_counts, dtype=np.float64)
 
-        if result["counts"] == [None]:
```
gergness: Also not sure about this (I had previously introduced this line to avoid having to change the behavior of `cube_part::Nub::is_empty` - see change below).

This felt cleaner to me because it makes the return type more predictable.

src/cr/cube/matrix/subtotals.py
scanny: Well, we probably should I suppose. My branch is on master now, so the bulk is already handled, and the bits that Ernesto is doing is primarily adds rather than changes, so I don't think the merge conflicts are going to be too big for him. I'm pretty sure yours will go first, because he's got a lot of changes to do on his branch and I expect it will be Monday or so before he's ready to push.

src/cr/cube/cube.py
scanny: I'd say those are distinct enough to warrant two different subclasses, where the factory chooses between `_MeanMeasure` and perhaps something like `_NumArrayMeanMeasure`. Then each `._flat_values` can do its own thing and we avoid using `if` statements as a substitute for polymorphism.

Btw, The docstring says "Return tuple ..." and the return value here is `ndarray` (which I think is preferable).

This issue came up in my review of Ernesto's branch here: https://github.com/Crunch-io/crunch-cube/pull/251#discussion_r586703412 where I explained that I thing _all_ `._flat_values` return type should be `ndarray` to avoid the unnecessary time and memory required to construct an intermediate `tuple` object that just gets immediately converted to an `ndarray` anyway.

So maybe a conflict brewing here, but that happens and Ernesto can figure it out during rebase (although it might be worth a mention to him in Slack). In any case, I think the right solution is to create the `_NumArrayMeanMeasure` or whatever name is appropriate and then adjust the factory in `_Measures` or wherever it is to choose the right subclass when constructing the measure. Also that these return `ndarray` and preferably that _all_ of the four subtypes return `ndarray` and the `_BaseMeasure.raw_cube_array` property is modified to avoid constructing an array out of an array (which it looks like you already did).

Also the docstring for `_WeightedCountMeasure._flat_values` also says `tuple`.

src/cr/cube/cube.py
scanny: Hmm, a list containing the single value `None` seems odd. Do we have JSON that looks like `"counts": [null]` or something? That would be ugly indeed, as we could choose either `null` or an empty array and surely one of those would suit better.

Anyway, I'll see what I think after looking at the Nub bit, otherwise I'd say good riddance to this line :)

src/cr/cube/cube.py
```diff
@@ -704,7 +704,7 @@ def raw_cube_array(self):
         response. Specifically, it includes values for missing elements, any
         MR_CAT dimensions, and any prunable rows and columns.
         """
-        raw_cube_array = np.array(self._flat_values).flatten().reshape(self._shape)
+        raw_cube_array = self._flat_values.flatten().reshape(self._shape)
```
scanny, line 707: Right on. Great minds think alike I suppose :)

src/cr/cube/cube.py
```diff
@@ -735,7 +735,7 @@ def valid_counts(self):
 
     @lazyproperty
     def _flat_values(self):  # pragma: no cover
-        """Return tuple of mean values as found in cube response.
+        """Return ndarray of np.float64 values as found in cube response.
```
scanny, line 738: It's my habit, and perhaps more proper, to leave out the "Return " suffix on docstrings of properties (lazy or otherwise) since they behave like attributes. So we document a property like we would an attribute, by stating what it _is_ rather that what it "returns", since returning a value is a characteristic of a function and attributes just "are" their value.

As a side benefit, it opens up first-docstring-line real estate for more descriptive words.

src/cr/cube/cube.py
```diff
@@ -784,20 +796,21 @@ class _UnweightedCountMeasure(_BaseMeasure):
...
-            # ---Under this circumstances the numeric array measures will contain the
-            # ---mean measure and a valid count measure for the unweighted counts.
-            return tuple(
-                self._cube_dict["result"]["measures"]["valid_count_unweighted"]["data"]
-            )
-        return tuple(self._cube_dict["result"]["counts"])
+        """np.ndarray of np.float64 counts before weighting or None, if unavailable.
+
+        Use floats to avoid int overflow bugs and so we can use nan.
+        """
```
scanny, line 802: One convention I've adopted, which is consistent with Python 3 type hints, is to prefix are return value that can also be None as "Optional ". So this would become "Optional 1D np.float64 ndarray of counts before weighting."

Then in the body of the docstring you can mention the circumstances under which it is None, like "This value is `None` when the cube-result does not contain a cube-counts measure."

src/cr/cube/cube.py
```diff
@@ -784,20 +796,21 @@ class _UnweightedCountMeasure(_BaseMeasure):
...
+
+        # ---If valid_count are expressed in the cube dict, returns its data.
+        # ---This condition can happen in case of numeric array cube response.
+        # ---Under this circumstances the numeric array measures will contain the
+        # ---mean measure and a valid count measure for the unweighted counts.
+        valid_counts = result["measures"].get("valid_count_unweighted", {}).get("data")
+        if valid_counts:
+            return np.array(valid_counts, dtype=np.float64)
+
+        return np.array(result["counts"], dtype=np.float64)
```
scanny, line 813: It's a very fine point, but just in case you're interested, I would generally phrase this as:

```python
return (
    np.array(valid_counts, dtype=np.float64)
    if valid counts
    else np.array(result["counts"], dtype=np.float64)
)
```

I'm not entirely sure why I like this better, except that as a reader you don't have to reason about two separate statements and maybe any unhandled "fall-through" cases or something; like it's a straight-line execution path and says "okay, we're returning a value now, there's no two ways about it, it's just a question of what that value is going to be.

Anyway do it the way you like best, just wanted to share my perspective on that.

src/cr/cube/cube.py
```diff
@@ -806,4 +819,6 @@ class _WeightedCountMeasure(_BaseMeasure):
     @lazyproperty
     def _flat_values(self):
         """tuple of numeric counts after weighting."""
-        return tuple(self._cube_dict["result"]["measures"]["count"]["data"])
+        return np.array(
+            self._cube_dict["result"]["measures"]["count"]["data"], dtype=np.float64
+        )
```
scanny, line 824: The docstring still says `tuple` but this is `1D np.float64 ndarray` now. But more importantly, I think this value needs to be optional now, like it is `None` if there is no `measures.count` field in the JSON.

Hmm, I suppose that's handled by the `_Measures.is_weighted` property that checks for this in the payload ... but that's awfully far away. Hmm, the broader structure of responsibilities is a little borked up it looks like, with like `_Measures.means` also checking the payload first.

I'd be inclined to think that `._flat_values` should be optional for _all_ measures that are sometimes not there, which I believe is all measures. Then if we want to know whether it has a measure we check whether that measure is None and we leave the payload interpretation to the actual measure class that can fully encapsulate it.

Not sure if you feel up to that refactoring just now, but I'd say at least that this `._flat_values` should be optional and maybe fix the spot in `_Measures.is_weighted` that inspects the payload for itself to rely on a value from `_WeightedCountMeasure` rather than poke around in the JSON. I suppose that means that `_BaseMeasure.raw_cube_array` would need to be optional too.

Hmm, this might require a little more noodling, and I don't want to launch you into a quagmire, but maybe you can have a think on it and we can spend a cafe minute on it or something. Either way, it can't hurt to start by making this optional.

scanny (changes requested):
@gergness a few comments I wanted to send your way before breaking for lunch, I'll continue when I get back.

src/cr/cube/cube.py
scanny: Okay, after reflection and nourishment, here's what I think:

1. All `._flat_values` are type `Optional 1D float64 ndarray` (maybe numeric array is 2D or something, I don't understand that enough to say, but ndarray anyway). The value is `None` when there is no such cube-measure in the payload or the values it needs are otherwise not to be found. This way, only the measure objects need to know where to go in the payload to get their values and all other measure-related payload inspection can be eliminated.

2. `_BaseMeasure.raw_cube_array` is also optional. Since this is the main interface method of a measure object (and `._flat_values` is private), it being `None` is how any higher-level unit figures out whether that base measure exists in the payload.

3. If we still need `_Measures.weighted_counts` to automatically substitute `_UnweightedCountMeasure` when the cube is unweighted, then we give `_Measures` a private `._weighted_counts` property which is either a `_WeightedCountMeasure` or `None` and let the interface property `.weighted_counts` perform the test and substitution, like `return self._weighted_counts if self._weighted_counts is not None else self.unweighted_counts`.

Overall, it still smells a little fishy, because I think that auto-substitution thing probably has to go, but this would at least remove part of the long-distance coupling and leakage of responsibilities.

Probably even better would be to add a `.load()` classmethod to `_BaseMeasure` and that is called for construction purposes rather than the subclass itself. That way it can return `None` if the measure cannot be created and callers don't have to test `.raw_cube_array` to see if their reference is valid or just an empty shell.

Let me know if this sounds like more than you bargained for and I can add a few-commit branch after yours merges. But it's not untypical that modules need some cleaning up before adding new features, so it wouldn't be a bad exercise to get this one tidied up a bit. This module mostly goes back to 2019 so it's a little bit crusty around the edges.

src/cr/cube/cubepart.py
```diff
@@ -491,7 +494,7 @@ def columns_scale_median(self):
                 for i in range(counts.shape[1])
             ]
         )
-
+        assert scale_median.dtype == np.float64
```
scanny, line 497: probably this line needs to come out.

src/cr/cube/cubepart.py
```diff
@@ -881,6 +878,7 @@ def rows_scale_median(self):
                 for i in range(counts.shape[0])
             ]
         )
+        assert scale_median.dtype == np.float64
```
scanny, line 881: probably this one too.

src/cr/cube/cubepart.py
```diff
@@ -1546,7 +1544,7 @@ class _Nub(CubePartition):
 
     @lazyproperty
     def is_empty(self):
-        return False if self.unweighted_count else True
+        return math.isnan(self.unweighted_count)
```
scanny, line 1547: This property needs a docstring that states its contract.

Assuming that `.is_empty` means: "contains no values", I wonder what its semantics are for a non-container object (as I understand a `Nub` to be). Also it's not clear to me the circumstances under which a `Nub` value for unweighted count can be NaN. These are questions I'd like the docstring to answer for me.

src/cr/cube/dimension.py
```diff
@@ -958,3 +958,33 @@ def label(self):
...
+        """
+        subtrahend_ids = self.subtrahend_ids
+        return np.fromiter(
+            (
+                idx
+                for idx, vector in enumerate(self._valid_elements)
+                if vector.element_id in subtrahend_ids
+            ),
+            dtype=int,
+        )
```
scanny, line 990: `subtrahend_ids` is only used once. Why wouldn't you inline `self.subtrahend_ids`?

Might also be worth a comment mentioning that the reason the `dtype=int` is important is that this value is used for indexing, an empty array takes the default type `float` and an array of type float cannot be used for indexing, so this is needed to prevent an error when the array is empty (as it often is).

src/cr/cube/matrix/assembler.py
```diff
@@ -451,16 +463,21 @@ def _assemble_matrix(self, blocks):
     def _assemble_vector(self, base_vector, subtotals, order):
         """Return 1D ndarray of `base_vector` with inserted `subtotals`, in `order`.
 
-        Each subtotal value is the result of applying np.sum to the addends extracted
-        from `base_vector` according the the `addend_idxs` property of each subtotal in
-        `subtotals`. The returned array is arranged by `order`, including possibly
-        removing hidden or pruned values.
+        Each subtotal value is the result of applying np.sum to the addends and
+        subtrahends extracted from `base_vector` according the the `addend_idxs`
```
scanny, line 467: "... according _to_ the ...

src/cr/cube/matrix/assembler.py
```diff
@@ -451,16 +463,21 @@ def _assemble_matrix(self, blocks):
...
+        and `subtrahend_idxs` property of each subtotal in `subtotals`. The returned
+        array is arranged by `order`, including possibly removing hidden or pruned
+        values.
         """
-        # TODO: This only works for "sum" subtotals, which is all that it needs so far,
+        # TODO: This works for "sum" and "diff" subtotals, which is all that it needs so far,
```
scanny, line 472: extends past the 88 char margin.

src/cr/cube/matrix/subtotals.py
```diff
@@ -73,7 +68,7 @@ def _intersections(self):
                 for row_subtotal in self._row_subtotals
                 for column_subtotal in self._column_subtotals
             ],
-            dtype=self._dtype,
+            dtype=np.float64,
```
scanny, line 71: do we still need an explicit `dtype` parameter here or can we just let the defaults take over?

src/cr/cube/matrix/subtotals.py
```diff
@@ -156,7 +151,7 @@ def _subtotal_row(self, subtotal):
 
 
 class SumSubtotals(_BaseSubtotals):
-    """Subtotal "blocks" created by np.sum() on addends, primarily counts."""
+    """Subtotal "blocks" created by np.sum() on addends and subtrahends, primarily bases."""
```
scanny, line 154: too long. can probably drop the `, primarily bases.` or stretch that to its own sentence in a docstring body. (I would just cut it.)

src/cr/cube/stripe/insertion.py
```diff
@@ -55,27 +55,43 @@ def _subtotal_values(self):
...
+    """
 
     @lazyproperty
     def _subtotal_values(self):
         """(n_row_subtotals,) ndarray of subtotal values for stripe."""
         subtotals = self._row_subtotals
 
         if len(subtotals) == 0:
-            return np.array([], dtype=self._dtype)
+            return np.array([], dtype=np.float64)
```
scanny, line 70: I think this `dtype` is safe to drop; `float` is the default type.

scanny (changes requested):
Okay, this is all the time I have for this today. I'll see if I can make it through the tests tomorrow.

src/cr/cube/cube.py
gergness: > I'd say those are distinct enough to warrant two different subclasses, where the factory chooses between _MeanMeasure and perhaps something like _NumArrayMeanMeasure. Then each ._flat_values can do its own thing and we avoid using if statements as a substitute for polymorphism.

Just saw your new set of comments, but the problem I'm hitting as I thought about this is that I don't see a good way to distinguish numeric arrays from not at the time of the factory. Did you have thoughts on that?

My first idea was to use use `_BaseMeasure._shape` but the shape of a CAT(N categories) X Numeric is the same as a Numeric Array (N subvariables).

The other areas where the code makes decisions about numeric arrays (eg `_numeric_array_dimension`) are in the `Cube` class, but I don't see a good way to pass that information along when creating the measures.

src/cr/cube/cubepart.py
gergness: The integration test that breaks is `integration/test_cube_part.py::Describe_Nub::it_is_empty` which loads https://github.com/Crunch-io/crunch-cube/blob/f9e0b2a4cf81e955e592b4584d532372c1089cce/tests/fixtures/econ-nodata-no-dims.json#L84-L86, which does indeed have `"counts":{null}`. My best guess is it's what happens when the dataset has no rows (or all rows have been filtered/excluded)?

gergness: Hm, in my testing, a query from a dataset with 0 rows, a dataset excluded down to 0 rows, and a query filtered down to 0 rows all have a 0 in the counts field, unlike this fixture which has `null`.

src/cr/cube/cube.py
gergness: Hm, but this was mislabelled because it never returns None, right?

src/cr/cube/matrix/subtotals.py
gergness: I don't think it'd error, but if empty it'd convert to a non-float, right? Is there value in ensuring that it'd be the same type as the other blocks?

src/cr/cube/cubepart.py
gergness: Fixture was introduced in https://github.com/Crunch-io/crunch-cube/pull/181

src/cr/cube/cube.py
gergness: Okay, I think I got this now

scanny (approved):
Okay, LGTM, very nice job :)

PR #256, opened 2021-03-15
Spike overlaps

slobodan-ilic:


src/cr/cube/cube.py
```diff
@@ -773,6 +793,50 @@ def _flat_values(self):
...
+        return self._all_dimensions.shape + (n_subvars,)
+
+    @lazyproperty
+    def raw_cube_array(self):
+        if self._flat_values is None:
+            return None
+        raw_cube_array = self._flat_values.reshape(self._shape)
+        # ---must be read-only to avoid hard-to-find bugs---
+        raw_cube_array.flags.writeable = False
+        return raw_cube_array
```
ernestoarbitrio, line 837: I don't see the reason of having this code fragment duplicated.
I would suggest to create a shape property in `_BaseMeasure`:
```python3
@lazyproperty
def _shape(self):
    """ ... """
    return self._all_dimensions.shape
```
and then override this shape in the `Overlap` measure and get rid of the `raw_cube_array` that is basically the same code of the `BaseMeasure`

src/cr/cube/cube.py
```diff
@@ -773,6 +793,50 @@ def _flat_values(self):
...
+            tuple(
+                np.nan if type(x) is dict else x for x in self._measure_payload["data"]
+            ),
+            dtype=np.float64,
+        ).flatten()
+
+    @lazyproperty
+    def _shape(self):
+        n_subvars = len(self._measure_payload["metadata"]["type"]["subvariables"])
+        return self._all_dimensions.shape + (n_subvars,)
```
ernestoarbitrio, line 828: please keep alphabetical order for these props

src/cr/cube/cube.py
```diff
@@ -773,6 +793,50 @@ def _flat_values(self):
...
+            return None
+        return np.array(
+            tuple(
+                np.nan if type(x) is dict else x for x in self._measure_payload["data"]
+            ),
+            dtype=np.float64,
+        ).flatten()
+
+    @lazyproperty
+    def _shape(self):
```
ernestoarbitrio, line 826: docstring

src/cr/cube/cube.py
```diff
@@ -773,6 +793,50 @@ def _flat_values(self):
...
+class _OverlapMeasure(_BaseMeasure):
+    """Statistical overlap values from a cube-response."""
+
+    @lazyproperty
+    def missing_count(self):
+        """Numeric value representing count of missing rows in response."""
+        return self._cube_dict["result"]["measures"]["overlap"].get("n_missing", 0)
+
+    @lazyproperty
+    def _measure_payload(self):
```
ernestoarbitrio, line 805: docstring

src/cr/cube/cube.py
```diff
@@ -773,6 +793,50 @@ def _flat_values(self):
         ).flatten()
 
 
+class _OverlapMeasure(_BaseMeasure):
+    """Statistical overlap values from a cube-response."""
+
+    @lazyproperty
+    def missing_count(self):
+        """Numeric value representing count of missing rows in response."""
+        return self._cube_dict["result"]["measures"]["overlap"].get("n_missing", 0)
```
ernestoarbitrio, line 802: do we need missing counts for overlaps? If yes this property is not used anywhere. So I think that the business logic at line 636 (`_Measures` -> `missing_count`) should be augmented.

src/cr/cube/stripe/cubemeasure.py
```diff
@@ -92,6 +124,15 @@ def means(self):
         return self._means
 
 
+class _CatCubeOverlaps(_BaseCubeOverlaps):
+    """Overlaps cube-measure for a non-MR stripe."""
+
+    @lazyproperty
+    def overlaps(self):
+        """1D np.float64 ndarray of cube overlaps for each stripe row."""
+        return self._means
```
ernestoarbitrio, line 133: i think this is overlaps isn't ? self._means is a copy paste typo i guess

src/cr/cube/stripe/cubemeasure.py
```diff
@@ -92,6 +124,15 @@ def means(self):
         return self._means
 
 
+class _CatCubeOverlaps(_BaseCubeOverlaps):
+    """Overlaps cube-measure for a non-MR stripe."""
```
ernestoarbitrio, line 128: Out of curiosity: do we have overlaps for non MR stripe? Considering that this is a 1D case (because of the name Stripe :D) why we need a `CatCubeOverlap` case?

ernestoarbitrio (changes requested):
left some comments

src/cr/cube/cube.py
slobodan-ilic: Done

src/cr/cube/cube.py
slobodan-ilic: we don't need it afaics

src/cr/cube/stripe/cubemeasure.py
slobodan-ilic: ditched the entire class since we don't need it

src/cr/cube/stripe/cubemeasure.py
slobodan-ilic: don't need it, removed it...

ernestoarbitrio (changes requested):
per comments

scanny (changes requested):
I'll need to take another look once you addressed these and gotten docstrings written, but this should be a good start.

src/cr/cube/cube.py
```diff
@@ -767,6 +807,15 @@ def _flat_values(self):  # pragma: no cover
         """
         raise NotImplementedError("must be implemented by each subclass")
 
+    @lazyproperty
+    def _shape(self):
+        """tuple(int) representing the shape of the measure.
```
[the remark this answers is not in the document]
slobodan-ilic: I think it deserves to have a shape of its own. I've changed the wording.

src/cr/cube/dimension.py
```diff
@@ -255,6 +262,7 @@ def _resolve_array_type(self):
             next_raw_dimension._base_type == "categorical"
             and next_raw_dimension._has_selected_category
             and next_raw_dimension._alias == self._alias
+            and next_raw_dimension._is_logical_type
```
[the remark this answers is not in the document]
slobodan-ilic: Yes. If an array has e.g. 5 categories, and one of them is selected, this is going to evaluate to true. So, if we have e.g.: `"Poor", "Good", "Satisfying", "No Data"`, and "Poor" is the default selected, this is going to be true, which is not what we want. We _only_ want it to be true when the categories are `Selected, Other, Missing`, which is defined by [-1, 0, 1] (i.e. the logical type).

src/cr/cube/matrix/assembler.py
```diff
@@ -168,6 +168,10 @@ def means(self):
         """
         return self._assemble_matrix(self._measures.means.blocks)
 
+    @lazyproperty
+    def pairwise_significance_tests(self):
```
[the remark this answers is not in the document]
slobodan-ilic: I've changed the name of this property. It's now more explicit and is documented better.

src/cr/cube/matrix/assembler.py
```diff
@@ -168,6 +168,10 @@ def means(self):
...
+        return self._measures.overlaps.pairwise_significance
```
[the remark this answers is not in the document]
slobodan-ilic: I've removed `overlaps` as a 2nd order measure completely. I think it's better now, and more explicit.

src/cr/cube/matrix/measure.py
```diff
@@ -50,6 +52,11 @@ def means(self):
         """_Means measure object for this cube-result"""
         return _Means(self._dimensions, self, self._cube_measures)
 
+    @lazyproperty
+    def overlaps(self):
```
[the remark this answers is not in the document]
slobodan-ilic: removed

src/cr/cube/matrix/measure.py
```diff
@@ -378,6 +385,160 @@ def blocks(self):
         )
 
 
+class _Overlaps(_BaseSecondOrderMeasure):
```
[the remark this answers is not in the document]
slobodan-ilic: removed

src/cr/cube/cubepart.py
```diff
@@ -618,6 +618,8 @@ def pairwise_significance_tests(self):
         significance test contains `p_vals` and `t_stats` (ndarrays that represent
         probability values and statistical scores).
         """
+        if self.dimension_types[-1] == DT.MR and self._cube.overlaps is not None:
+            return self._assembler.pairwise_significance_tests
         return tuple(
             PairwiseSignificance(self).values[column_idx]
             for column_idx in range(len(self.column_labels))
```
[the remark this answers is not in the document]
slobodan-ilic: I've created a measure class that's more in line with the latest architecture. I still haven't moved the `PairwiseSignificance` into it, but I get a broad sense of how it should be done.

src/cr/cube/matrix/cubemeasure.py
```diff
@@ -166,6 +171,72 @@ def means(self):
...
+    @lazyproperty
+    def overlaps(self):
+        """3D np.float64 ndarray of overlaps for each valid matrix cell."""
+        return self._overlaps[:, 0, :]
+
+    @lazyproperty
+    def valid_overlaps(self):
+        return self._valid_overlaps[:, 0:2, :].sum(axis=1)
+
+
```
ernestoarbitrio, line 239: why we have MrXCat Overlaps case? I don't see any reason to have overlaps with categorical variable on the column, even because overlaps are used for the pariweise significance and we do not have this features on the rows but only on the cols

src/cr/cube/matrix/assembler.py
```diff
@@ -168,6 +168,34 @@ def means(self):
...
+        Raises `ValueError if the cube-result does not include `overlaps`
+        and `valid_overlaps` cube-measures.
+        """
+        return np.array(
+            [
+                self._assemble_matrix(self._measures.get_t_stats_for(subvar_idx).blocks)
+                for subvar_idx in range(len(self.column_labels))
+            ]
+        )
+
```
ernestoarbitrio, line 198: Can you explain why we need 2 different property for tstats and pvals? They should come from the `PairwiseTtest` object isn't it. This way we compute the calc 2 times just for accessing different values. Is there a reason for doing this in separate props?

src/cr/cube/matrix/measure.py
```diff
@@ -50,6 +52,16 @@ def means(self):
         """_Means measure object for this cube-result"""
         return _Means(self._dimensions, self, self._cube_measures)
 
+    def get_t_stats_for(self, subvar_idx):
```
ernestoarbitrio, line 55: Why this `get_t_stats_for` ? It smells like a Java property 😂 . Why not `t_stats` ? Or even, as mentioned before, wrap tstats and pvals withiun a single functions that have pvals and zscore, and then just a `pairwise_t_test`  funciont where we pass the subvar_idx and returns both t-tests values? If i well remember these values are mainly used in the exporter for the color shading of the cells for rendering the significance. So we always need them both for doing stuff ... i'm inclined to think that having 2 separate function is not worthwhile.

ernestoarbitrio (changes requested):
@slobodan-ilic per comments. I just stop at the measure level cause i need to understand if something can be refactored without having t-stats and p-vals as separate measures. I'll take another look whenever you ready

src/cr/cube/matrix/cubemeasure.py
slobodan-ilic: True, we don't need it. I just added it to confirm that we're able to interpret numbers that come in from zz9. I'll delete, and also delete the tests.

src/cr/cube/matrix/assembler.py
slobodan-ilic: Yes. The reason is to not force the client to inspect the objects we return. It makes much more sense from the client perspective to use something like `pairwise_t_stats` than it does to use `pairwise_sig_tests.t_stats`. We might discuss how to optimize, but from API POW, it clearly warrants two different properties, afaics.

slobodan-ilic: The old way (where we wrap them together) is more convenient for us, but it's ugly for the clients (e.g. exporter).

src/cr/cube/matrix/measure.py
slobodan-ilic: Yeah, this is the complicated part... I need to have a measure "per subvariable". How to do that is a question... I need a matrix for each subvariable, essentially making the result of the measure 3D. But I can only apply the concept of `blocks` on a 2D. So I need to iterate. And the question is: where I do the iteration so it makes the most sense? In this approach, I've tried to iterate over subvars, and create a measure for each one of them, so that the measure itself can behave like all other measures do. And then assembler just wraps those matrices together.

I'll try to do something better... I'll see...

src/cr/cube/matrix/cubemeasure.py
slobodan-ilic: ok, ditched this

src/cr/cube/matrix/measure.py
slobodan-ilic: fixed the naming, but left them as functions. resolving conversation so that I can attend to other comments.

src/cr/cube/matrix/assembler.py
slobodan-ilic: also resolving this for now, might get to the "optimize" part later, I still have a bunch of steve's comments to address

src/cr/cube/cubepart.py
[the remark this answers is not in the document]
slobodan-ilic: I'm resolving this comment for now, as I've mostly fixed the architecture of the newly added classes, according to the new spec.

src/cr/cube/cubepart.py
```diff
@@ -618,6 +618,8 @@ def pairwise_significance_tests(self):
         significance test contains `p_vals` and `t_stats` (ndarrays that represent
         probability values and statistical scores).
```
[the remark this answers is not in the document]
slobodan-ilic: I made it separate, for `p_vals` and `t_stats`. They also return `ndarray` now, consistently. The legacy measure is also separated into two api methods, but it's still not refactored.

src/cr/cube/matrix/measure.py
```diff
@@ -378,6 +385,160 @@ def blocks(self):
         )
 
 
+class _Overlaps(_BaseSecondOrderMeasure):
+    """Provides the overlap measure for a matrix."""
+
+    @lazyproperty
+    def pairwise_significance(self):
+        return _OverlapsSignificance(self._overlaps, self._valid_overlaps).values
```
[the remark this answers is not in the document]
slobodan-ilic: This is refactored now, and docstrings added accordingly

src/cr/cube/matrix/measure.py
```diff
@@ -378,6 +385,160 @@ def blocks(self):
...
+        return self._cube_measures.cube_overlaps.valid_overlaps
+
+
+OverlapSignificance = namedtuple("OverlapSignificance", "t_stats p_vals")
+
+
+class _OverlapsSignificance(object):
+    def __init__(self, overlaps, valid_overlaps):
+        assert overlaps.shape == valid_overlaps.shape
+        assert overlaps.shape[1] == overlaps.shape[2]
```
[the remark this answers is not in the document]
slobodan-ilic: Removed. I wanted to leave them in on purpose.

src/cr/cube/matrix/measure.py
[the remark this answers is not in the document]
slobodan-ilic: After refactoring, it's accomplished by using nan blocks on each matrix

src/cr/cube/matrix/measure.py
```diff
@@ -378,6 +385,160 @@ def blocks(self):
...
+    @lazyproperty
+    def _valid_overlaps(self):
+        return self._cube_measures.cube_overlaps.valid_overlaps
+
+
+OverlapSignificance = namedtuple("OverlapSignificance", "t_stats p_vals")
+
+
+class _OverlapsSignificance(object):
+    def __init__(self, overlaps, valid_overlaps):
```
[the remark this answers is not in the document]
slobodan-ilic: removed this object. good catch

src/cr/cube/matrix/measure.py
```diff
@@ -378,6 +385,160 @@ def blocks(self):
...
+
+    @lazyproperty
+    def values(self):
+        values = []
+        # for i in range(len(self._slice.column_labels)):
+        for i in range(self._n_subvar_columns):
+            t_stats = self._t_stats[:, i, :]
+            p_vals = self._p_vals[:, i, :]
+            values.append(OverlapSignificance(t_stats, p_vals))
+        return values
```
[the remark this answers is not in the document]
slobodan-ilic: removed this completely, as well as the intermediate result class.

src/cr/cube/matrix/measure.py
```diff
@@ -378,6 +385,160 @@ def blocks(self):
...
+                    self._overlaps[row_idx], self._valid_overlaps[row_idx]
+                ).p_vals
+                for row_idx in range(self._n_rows)
+            ]
+        )
+
+
+class _PairwiseSignificanceBetweenSubvarsAAndB(object):
+    def __init__(self, overlaps, valid_overlaps, idx_a, idx_b):
+        self._overlaps = overlaps
```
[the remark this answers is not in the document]
slobodan-ilic: done. also ditched most of those classes after rfctr

src/cr/cube/matrix/measure.py
```diff
@@ -378,6 +385,160 @@ def blocks(self):
...
+    def Sa(self):
+        return self._overlaps[self._idx_a, self._idx_a]
+
+    @lazyproperty
+    def Sb(self):
+        return self._overlaps[self._idx_b, self._idx_b]
+
+    @lazyproperty
+    def Sab(self):
+        return self._overlaps[self._idx_a, self._idx_b]
```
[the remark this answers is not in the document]
slobodan-ilic: refactored

src/cr/cube/matrix/measure.py
```diff
@@ -378,6 +385,160 @@ def blocks(self):
...
+
+    @lazyproperty
+    def t_stats(self):
+        if self._idx_a == self._idx_b:
+            return 0
+
+        pa = self.Sa / self.Na
+        pb = self.Sb / self.Nb
+        pab = self.Sab / self.Nab
```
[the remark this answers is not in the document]
slobodan-ilic: refactored

src/cr/cube/stripe/measure.py
```diff
@@ -34,6 +34,11 @@ def means(self):
         """_Means measure object for this stripe."""
         return _Means(self._rows_dimension, self, self._cube_measures)
 
+    @lazyproperty
+    def overlaps(self):
+        """_Overlaps measure object for this stripe."""
+        return _Overlaps(self._rows_dimension, self, self._cube_measures)
```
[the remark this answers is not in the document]
slobodan-ilic: not a thing. removed

src/cr/cube/stripe/measure.py
```diff
@@ -192,6 +197,23 @@ def subtotal_values(self):
...
+class _Overlaps(_BaseSecondOrderMeasure):
+    """Provides the MR cube overlaps measure for a stripe.
+
+    Relies on the presence of the overlaps cube-measure in the cube-result.
+    """
+
+    # We will only ever need base values here, and not the subtotals. This is so
+    # because the only variable that can have overlaps is MR, and we don't have
+    # cr.cube insertions on the MR variables (we have zz9 insertions instead, which
+    # are going to be automatically included, since they're treated as data by cr.cube).
```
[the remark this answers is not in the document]
slobodan-ilic: yup, dead code... removed...

src/cr/cube/matrix/cubemeasure.py
```diff
@@ -166,6 +171,72 @@ def means(self):
...
+    """Overlaps cube-measure for a NOT_MR_X_MR slice."""
+
+    @lazyproperty
+    def overlaps(self):
+        """3D np.float64 ndarray of overlaps for each valid matrix cell."""
+        return self._overlaps[:, :, 0]
+
+    @lazyproperty
+    def valid_overlaps(self):
+        return self._valid_overlaps[:, :, 0:2].sum(axis=2)
```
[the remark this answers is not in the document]
slobodan-ilic: Thanks for the advice. The reason this was so unfinished is that it wasn't finished 🤷  I really didn't want you to do a review in this much detail, especially because I was uncertain about the new architecture. What I was hoping you'd do is more of a skim and general pointers.

However, after a couple of iterations, I now believe to have addressed all of your comments, and that the architecture is pretty sound.

tests/integration/test_pairwise_significance.py
```diff
@@ -324,3 +360,325 @@ def test_mr_x_mr_pairwise_t_tests(self):
...
+            slice_.pairwise_significance_t_stats[1],
+            load_python_expression("mr-x-mr-pw-tstats"),
+        )
+        np.testing.assert_array_almost_equal(
+            slice_.pairwise_significance_p_vals[1],
+            load_python_expression("mr-x-mr-pw-pvals"),
+        )
+
+
+class TestOverlapsCounts(TestCase):
```
ernestoarbitrio, line 374: General thought is that this class maybe doesn't belong in this test module, here we are testing the cubepart interface but in this class you are testing the assembler basically ... i would move this class in the right test module

tests/integration/test_pairwise_significance.py
```diff
@@ -324,3 +360,325 @@ def test_mr_x_mr_pairwise_t_tests(self):
...
+    def test_overlaps_cat_x_mr_sub_x_mr_sel(self):
+        matrix = Cube(OL.CAT_X_MR_SUB_X_MR_SEL).partitions[0]
```
ernestoarbitrio, line 376: this is not a `matrix` this is `slice` object. It seems weird that you are instanciate a partition and use it to get the assembler. and then having no assertion on the partition but using it only to get the overalps. I mean you are starting from the top-level interface (the user interface) to query the number in the assembler. I'm a little bit concerned about the test approach. Sorry

tests/integration/test_pairwise_significance.py
```diff
@@ -324,3 +360,325 @@ def test_mr_x_mr_pairwise_t_tests(self):
...
+            # [
+            #     # A, B, C
+            #     [2, 1, 1],
+            #     [1, 1, 1],
+            #     [1, 1, 1],
+            # ],  # G[2] == missing
+        ]
+
+    def test_valid_overlaps_cat_x_mr_sub_x_mr_sel(self):
+        matrix = Cube(OL.CAT_X_MR_SUB_X_MR_SEL).partitions[0]
```
ernestoarbitrio, line 403: same here

tests/integration/test_pairwise_significance.py
```diff
@@ -324,3 +360,325 @@ def test_mr_x_mr_pairwise_t_tests(self):
...
+            #     # A, B, C
+            #     [0, 0, 0],  # A
+            #     [0, 0, 0],  # B
+            #     [0, 0, 0],  # C
+            # ],
+        ]
+
+    def test_valid_overlaps_ca_sub_x_ca_cat_x_mr_sub_x_mr_sel_subvar_0(self):
+        # Test subvar X (partitiions[0])
+        matrix = Cube(OL.CA_SUB_X_CA_CAT_X_MR_SUB_X_MR_SEL).partitions[0]
```
ernestoarbitrio, line 479: and here

tests/integration/test_pairwise_significance.py
```diff
@@ -324,3 +360,325 @@ def test_mr_x_mr_pairwise_t_tests(self):
...
+            [
+                [0.0, 0.0, 0.0],
+                [0.0, 0.0, 0.0],
+                [0.0, 0.0, 0.0],
+            ],
+        ]
+
+    def test_overlaps_ca_sub_x_ca_cat_x_mr_sub_x_mr_sel_subvar_1(self):
+        # Test subvar Y (partitions[1])
+        matrix = Cube(OL.CA_SUB_X_CA_CAT_X_MR_SUB_X_MR_SEL).partitions[1]
```
ernestoarbitrio, line 505: and here

tests/integration/test_pairwise_significance.py
```diff
@@ -324,3 +360,325 @@ def test_mr_x_mr_pairwise_t_tests(self):
...
+                [0.0, 0.0, 0.0],
+                [0.0, 1.0, 1.0],
+                [0.0, 1.0, 1.0],
+            ],
+        ]
+
+
+class TestOverlapsPairwiseSignificance(TestCase):
+    def test_pairwise_significance_cat_x_mr_sub_x_mr_sel_0th_subvar(self):
+        matrix = Cube(OL.CAT_X_MR_SUB_X_MR_SEL).partitions[0]
```
ernestoarbitrio, line 587: this is a slice not a matrix

tests/integration/test_pairwise_significance.py
```diff
@@ -324,3 +360,325 @@ def test_mr_x_mr_pairwise_t_tests(self):
...
+        np.testing.assert_almost_equal(
+            matrix.pairwise_significance_p_vals[0],
+            [
+                [0.0, 0.43590578, 0.24675171],
+                [0.0, np.nan, np.nan],
+            ],
+        )
+
+    def test_pairwise_significance_cat_x_mr_sub_x_mr_sel_1st_subvar(self):
+        matrix = Cube(OL.CAT_X_MR_SUB_X_MR_SEL).partitions[0]
```
ernestoarbitrio, line 605: slice

tests/integration/test_pairwise_significance.py
```diff
@@ -324,3 +360,325 @@ def test_mr_x_mr_pairwise_t_tests(self):
...
+        np.testing.assert_almost_equal(
+            matrix.pairwise_significance_p_vals[1],
+            [
+                [0.43590578, 0.0, 0.43590578],
+                [np.nan, 0.0, np.nan],
+            ],
+        )
+
+    def test_pairwise_significance_cat_x_mr_sub_x_mr_sel_2nd_subvar(self):
+        matrix = Cube(OL.CAT_X_MR_SUB_X_MR_SEL).partitions[0]
```
ernestoarbitrio, line 623: slice

tests/integration/test_pairwise_significance.py
```diff
@@ -324,3 +360,325 @@ def test_mr_x_mr_pairwise_t_tests(self):
...
+        np.testing.assert_almost_equal(
+            matrix.pairwise_significance_p_vals[2],
+            [
+                [0.24675171, 0.43590578, 0.0],
+                [np.nan, np.nan, 0.0],
+            ],
+        )
+
+    def test_pairwise_significance_cat_x_mr_realistic_example(self):
+        matrix = Cube(OL.CAT_X_MR_REALISTIC_EXAMPLE).partitions[0]
```
ernestoarbitrio, line 641: slice

ernestoarbitrio (changes requested):
per comments ... hope this will be the last iteration ... :D

src/cr/cube/matrix/measure.py
```diff
@@ -824,3 +913,62 @@ def blocks(self):
...
+
+    def __init__(self, overlaps, valid_overlaps, idx_a, idx_b):
+        self._overlaps = overlaps
+        self._valid_overlaps = valid_overlaps
+        self._idx_a = idx_a
+        self._idx_b = idx_b
+
+    @lazyproperty
+    def p_vals(self):
+        """float64 significance p-vals for the selected subvariables."""
```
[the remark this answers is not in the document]
slobodan-ilic: I would if I knew how. The only way I can think of is referencing the formulas, but this isn't terribly useful to the api user. And as far as it's about developers, the exact same formula is written in the code. So 🤷 😄

src/cr/cube/matrix/measure.py
```diff
@@ -824,3 +913,62 @@ def blocks(self):
...
+        if self._idx_a == self._idx_b:
+            return 0.0
+
+        return 2 * (1 - t.cdf(abs(self.t_stats), df=self._N - 2))
+
+    @lazyproperty
+    def t_stats(self):
+        """float64 significance t-stats for the selected subvariables."""
```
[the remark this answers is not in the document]
slobodan-ilic: same here 😟

src/cr/cube/matrix/measure.py
```diff
@@ -824,3 +913,62 @@ def blocks(self):
...
+        )
+
+    @lazyproperty
+    def _valid_counts(self):
+        """tuple(int) of valid counts for subvars a, b, and the overlaps (a^b)."""
+        return (
+            self._valid_overlaps[self._idx_a, self._idx_a],
+            self._valid_overlaps[self._idx_b, self._idx_b],
+            self._valid_overlaps[self._idx_a, self._idx_b],
+        )
```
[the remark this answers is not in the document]
slobodan-ilic: Well, now that you point it out... They _can_ actually be float, if a weight is used. But most of the times they won't be. So the answer to _why_ (when they're indeed ints) is that they come that way from ZZ9. Because it only adds together `True` or `False` values for each row of a subvariable (to figure out how many of those `row[idx] == 1` (when it's about selected overlaps) or to figure out `row[idx] != -1` (how many are valid)). Anyways, I'll change and elaborate.

src/cr/cube/matrix/measure.py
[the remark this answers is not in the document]
ernestoarbitrio: aight

tests/integration/test_pairwise_significance.py
slobodan-ilic: I agree, but it does test the numbers rather than the structure. And I haven't really found a good place for it other than the current place. I have however added one test that tests the structure (and not numbers), and placed it in a different module.

tests/integration/test_pairwise_significance.py
slobodan-ilic: refactored and moved to another module

tests/integration/test_pairwise_significance.py
slobodan-ilic: ok, figured out how to refactor

tests/integration/test_pairwise_significance.py
slobodan-ilic: refactored

ernestoarbitrio (approved):
LGTM

PR #257, opened 2021-03-17
Stddev measure 177024189

ernestoarbitrio:


src/cr/cube/cubepart.py
```diff
@@ -942,6 +942,19 @@ def summary_pairwise_indices(self):
             self, self._alpha, self._only_larger
         ).summary_pairwise_indices
 
+    @lazyproperty
```
slobodan-ilic, line 945: `"t"` is before `"u"`

slobodan-ilic: stddev before summary

src/cr/cube/matrix/measure.py
```diff
@@ -551,7 +556,20 @@ class _Sums(_BaseSecondOrderMeasure):
     @lazyproperty
     def blocks(self):
         """2D array of the four 2D "blocks" making up this measure."""
-        return SumSubtotals.blocks(self._cube_measures.cube_sum.sums, self._dimensions)
+        return SumDiffSubtotals.blocks(
```
slobodan-ilic, line 559: Is this on purpose? Do we want to change Sum -> SumDiff? Doesn't seem related to this PR.

ernestoarbitrio: it is not related but i prefer to release it in this cr.cube version cause we don't want sum subtotals but we wanna have the subtotal differences.

src/cr/cube/matrix/cubemeasure.py
```diff
@@ -161,6 +166,89 @@ def means(self):
...
+            else _CatXMrCubeStdDev
+            if dimension_types[1] == DT.MR
+            else _CatXCatCubeStdDev
+        )
+        return CubeSumsCls(
+            dimensions, cube.stddev[cls._slice_idx_expr(cube, slice_idx)]
+        )
+
+    @lazyproperty
+    def stddev(self):
```
slobodan-ilic, line 202: Do we want to use `_base_values` instead of `stddev` for consistency?

src/cr/cube/matrix/measure.py
ernestoarbitrio: just avoid 2 releases .. that;s the only reason

src/cr/cube/matrix/cubemeasure.py
ernestoarbitrio: actually each BaseCubeMeasure have it's measure name property and not base_values. maybe i'm not following you. Can you explain your concern?

slobodan-ilic (approved):

PR #258, opened 2021-03-19
Variance derived measures should be missing when proportion is missing

gergness:
This ended up being way easier when I realized that we were just re-calculating the proportions


scanny (approved):
Nice simplification :) LGTM.

PR #259, opened 2021-03-19
share of sum measure

ernestoarbitrio:
Introduces:
- `share_sum` measure for slice and strand
- `has_scale_means` to simplify some spec business rules in `cr.exporter`


ernestoarbitrio:
> [Coverage Status](https://coveralls.io/builds/38084006)
> 
> Coverage remained the same at 100.0% when pulling **[db40c7e](https://github.com/Crunch-io/crunch-cube/commit/db40c7ec41b8b84728dabd9fe7f9a233fc795b20) on shareofsum-cube-measure-177400407** into **[d8a1ca7](https://github.com/Crunch-io/crunch-cube/commit/d8a1ca7eb26d33b825f19fb55d3e7b81793a946d) on master**.

liar

src/cr/cube/cubepart.py
```diff
@@ -534,6 +534,11 @@ def description(self):
         """str description of this slice, which it takes from its rows-dimension."""
         return self._rows_dimension.description
 
+    @lazyproperty
+    def has_scale_means(self):
```
slobodan-ilic, line 538: This sounds very much like legacy. I thought we were going to get rid of all those "old" properties that test each thing separately. What do you need this for? I guess something for the client, since it's not used anywhere in the cube. But _what_ exactly in the client? Maybe it could be done differently from the client, to avoid this (rather obsolete) property?

src/cr/cube/cubepart.py
```diff
@@ -916,6 +921,19 @@ def rows_scale_median_margin(self):
     def shape(self):
         return self.counts.shape
 
+    @lazyproperty
+    def share_sum(self):
+        """2D optional np.float64 ndarray of share of sum value for each table cell.
```
slobodan-ilic, line 926: I know that this is how we write docstrings for our api, but maybe you could add a paragraf, below everything, that describes "what the share sum is"... It would be nice to have for a reader of the api docs. Like, if I didn't know what it was, from your 1-1 explanation, I'd just stare at this 👓

src/cr/cube/cubepart.py
ernestoarbitrio: We have a TODO: here https://github.com/Crunch-io/zoom/blob/ee846e34bf4b3859fc99bf05dc37ee854c8ccdb3/exporter/src/cr/exporter/spec.py#L804 that exercise the need of this check to be done in cube. And then the need of this 2 props here

src/cr/cube/cubepart.py
```diff
@@ -1269,6 +1287,11 @@ def inserted_row_idxs(self):
         """
         return self._assembler.inserted_row_idxs
 
+    @lazyproperty
+    def has_scale_means(self):
```
slobodan-ilic, line 1291: Same as above. Seems highly redundant to introduce this to to the API, when the same thing could be checked by inspecting a different property of the same object, directly by client.

src/cr/cube/cubepart.py
```diff
@@ -1435,6 +1458,19 @@ def smoothed_dimension_dict(self):
         """dict, row dimension definition"""
         return self._rows_dimension._dimension_dict
 
+    @lazyproperty
+    def share_sum(self):
```
slobodan-ilic, line 1462: I'm starting to think if we should have a base class for both strand and slice... Maybe also start adding API into the name of those classes, because that's essentially what they are. Maybe not now, but just a thought.

src/cr/cube/cubepart.py
ernestoarbitrio: see reply above :D any suggestion is more than welcome

src/cr/cube/cubepart.py
ernestoarbitrio: Anyway I can get rid of this props and revert the exporter business rule. Let me know what's your opinion on this regard

src/cr/cube/matrix/measure.py
```diff
@@ -550,6 +555,33 @@ def _subtotal_rows(self):
         return SumSubtotals.subtotal_rows(self._base_values, self._dimensions)
 
 
+class _ShareSum(_BaseSecondOrderMeasure):
+    """Provides the share of sum measure for a matrix."""
```
slobodan-ilic, line 559: Also maybe a couple of words about _what_ these values will be? Maybe not, I'm not sure, but this seems a bit thin for a first-time reader.

src/cr/cube/stripe/measure.py
```diff
@@ -315,6 +320,24 @@ def _weighted_counts(self):
...
+
+    @lazyproperty
+    def base_values(self):
+        """1D np.float64 ndarray of share of sum for each row."""
+        sums = self._cube_measures.cube_sum.sums
+        return sums / np.sum(sums)
+
+    @lazyproperty
+    def subtotal_values(self):
+        """1D ndarray of sum subtotals for each row-subtotal."""
```
slobodan-ilic, line 337: sum or share-sum subtotals?

ernestoarbitrio: u right , copy and paste typo ... damn

tests/integration/test_cubepart.py
```diff
@@ -858,6 +865,20 @@ def it_provides_stddev_measure_for_cat_x_mr(self):
         )
         assert slice_.table_base.tolist() == [3, 3, 3]
 
+    def it_provides_share_of_sum_measure_for_mr_x_mr(self):
+        slice_ = Cube(CR.MR_X_MR_SUM).partitions[0]
+
+        assert slice_.share_sum == pytest.approx(
+            np.array(
+                [
+                    [np.nan, np.nan, np.nan],
```
slobodan-ilic, line 874: urgh, this is ugly... why do we have this ugly expectation? couldn't we find something more useful?

tests/integration/test_cubepart.py
```diff
@@ -1181,6 +1208,21 @@ def it_provides_sum_and_mean_measure_for_CAT(self):
         assert strand.means == pytest.approx([2.66666667, 3.5])
         assert strand.sums == pytest.approx([8, 7])
 
+    def it_provides_share_of_sum_measure_for_CAT(self):
+        strand = Cube(CR.CAT_SUM).partitions[0]
+
+        assert strand.sums == pytest.approx([88.0, 77.0])
+        # --- share of sum is the array of sum divided by its sum, so in this case
+        # --- [88/165, 77/165]
+        assert strand.share_sum == pytest.approx([0.5333333, 0.4666666])
```
slobodan-ilic, line 1217: I think you could use `strand.share_sum.tolist()` and then just compare to the plain list (but maybe you'd have to make longer literals for numbers). `pytest.approx` is also pretty long itself, so, see which one fits better...

slobodan-ilic (approved):

PR #260, opened 2021-03-19
Subdiff 177369190

gergness:
Bases are not considered valid for subtotal differences along the same dimension. This allowed simplifying the design of subtotal differences because the case where SumSubtotal != SumDiffSubtotal is no longer valid so we can unify them.

### DIFF
--- a/src/cr/cube/matrix/assembler.py
+++ b/src/cr/cube/matrix/assembler.py
@@ -136,7 +136,7 @@ class Assembler(object):
         # --- an MR_X slice produces a 2D columns-margin (each cell has its own N) ---
         if self._rows_dimension.dimension_type == DT.MR_SUBVAR:
             return self._assemble_matrix(
-                SumSubtotals.blocks(
+                SumDiffSubtotals.blocks(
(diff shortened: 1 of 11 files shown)


scanny (approved):
LGTM, but I'm not clear on why we need both `diff_cols_nan` and `diff_rows_nan` and why one might be `True` when the other is `False`. I don't doubt that we _do_ need both, I just can't explain to myself why.

So maybe revisit the docstring for `SumSubtotals` and use that as the place to explain it. I and others (and yourself a month from now) will need to make sense of all that before we add in something new or whatever. You're the world's foremost expert on that at the moment, but the half-life on that expertise is about 30-days and with time-zone differences etc. it makes sense for that part of the code to explain itself :)

Some questions I'd like that prose to answer:

* What does `diff_rows/cols_nan` mean? What happens when it's True? What happens when it's False?
* How does one determine whether it should be set True or False?
* Why can't `SumSubtotals` figure out for itself whether that should be True or False? (Maybe because it varies by measure?)

PR #261, opened 2021-03-22
Don't hide subtotal differences with only negative terms & allow new positive kwargs

gergness:


src/cr/cube/dimension.py
```diff
@@ -882,11 +882,24 @@ def _iter_valid_subtotal_dicts(self):
...
-            if not {"anchor", "args", "name"}.issubset(insertion_dict.keys()):
+            if not {"anchor", "name"}.issubset(insertion_dict.keys()):
+                continue
+
+            # ---use "new style" kwargs defining positive terms if available---
+            positive = insertion_dict.get("kwargs", {}).get("positive", [])
+
+            # ---but if missing, check for "old style" args defining positive terms---
+            if not positive:
+                positive = insertion_dict.get("args", [])
```
ernestoarbitrio, line 893:
```suggestion
            positive = insertion_dict.get("kwargs", {}).get(
                "positive"
            ) or insertion_dict.get("args", [])
```

ernestoarbitrio: if you like this syntax you can replace the if and the positive assignment into 1 single line

src/cr/cube/dimension.py
```diff
@@ -949,11 +962,12 @@ def addend_ids(self):
...
-        return tuple(
-            arg
-            for arg in self._subtotal_dict.get("args", [])
-            if arg in self._valid_elements.element_ids
-        )
+        # ---Prefer positive "kwargs" over "args" so we can migrate---
+        positive = self._subtotal_dict.get("kwargs", {}).get("positive", [])
+        if not positive:
+            positive = self._subtotal_dict.get("args", [])
+
```
ernestoarbitrio, line 969: same here

ernestoarbitrio (approved):
LGTM :) - but i left just couple of comments for using a compact expression for the `positive`  variable assignment. If you like you can accept my comments otherwise - stamped!

PR #265, opened 2021-03-30
enable hiding using explicit subvar id in case of array vars

ernestoarbitrio:
For array dimensions e.g. `DT.NUM_ARRAY`, `DT.MR`, `DT.CA`, handle hiding transform using explicit subvarible id VS positional idx in the element dicts.
**_Keeps the retro-compatibility with the older expressions_**

### DIFF
--- a/src/cr/cube/dimension.py
+++ b/src/cr/cube/dimension.py
@@ -591,11 +591,17 @@ class _AllElements(_BaseElements):
 
     def _element_id_from_dict(self, element_dict):
         """Returns the element identifier given its dict."""
+        selector = self._elements_transforms.get("selector")
+        default_id = element_dict["id"]
         if self._dimension_type in DT.ARRAY_TYPES:
-            # --- In case of array dimensions returns the id of the subvariable.
-            return element_dict.get("value", {}).get("id")
+            element_value = element_dict.get("value", {})
+            if selector == "alias":
+                # --- In case of specified key alias, returns the subvariable alias.
+                return element_value.get("references", {}).get("alias") or default_id
+            # --- Returns the subvariable id.
+            return element_value.get("id")
         # --- Fallback case is the positional idx.
-        return element_dict["id"]
+        return default_id
 
     @lazyproperty
     def _elements(self):
@@ -613,6 +619,11 @@ class _AllElements(_BaseElements):
             ) in self._iter_element_makings()
         )
 
+    @lazyproperty
+    def _elements_transforms(self):
+        """Element transform dict expressed in the dimension transforms expression."""
+        return self._dimension_transforms_dict.get("elements", {})
+
     def _iter_element_makings(self):
         """Generate tuple of values needed to construct each element object.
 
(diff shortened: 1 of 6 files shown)


tests/integration/test_matrix.py
```diff
@@ -596,7 +596,10 @@ def it_computes_unweighted_counts_for_cat_hs_x_cat_hs_hiddens(self):
             [341, 175, 166, 234, 234],
         ]
 
-    def it_computes_unweighted_counts_for_ca_subvar_x_ca_cat_hiddens(self):
+    @pytest.mark.parametrize(
+        "hided_rows", ({"5": {"hide": True}}, {"00004": {"hide": True}})
```
slobodan-ilic, line 600: -> `hidden`

tests/integration/test_numeric_array.py
```diff
@@ -47,6 +47,26 @@ def it_provides_means_for_num_array_grouped_by_cat(self):
...
+        slice_ = Cube(
+            NA.NUM_ARR_MEANS_GROUPED_BY_CAT, transforms=transforms
+        ).partitions[0]
+
+        assert slice_.means == pytest.approx(
+            np.array(
+                [  # -------Gender-----
+                    #     M        F
+                    [87.6666667, 52.5],  # S1: Dark Night
+                    #     --      --       S2: Fight Club HIDED
```
slobodan-ilic, line 63: -> `HIDDEN`

src/cr/cube/dimension.py
```diff
@@ -586,6 +589,14 @@ def _element_dicts(self):
             else self._type_dict["elements"]
         )
 
+    def _element_id(self, element_dict):
```
slobodan-ilic, line 592: This is a function, can we name it like a function? In the sense: verb + object... Maybe something like `_get_element_id_from`... If that's too ugly, can we at least do `_element_id_from_dict`?

src/cr/cube/dimension.py
```diff
@@ -586,6 +589,14 @@ def _element_dicts(self):
...
+        """Returns the element identifier given its dict."""
+        if self._dimension_type in DT.ARRAY_TYPES:
```
slobodan-ilic, line 594: It would be much nicer to have explicit decision making here (like based on that mentioned `"key"` argument in the transforms).

slobodan-ilic (changes requested):
Per comments. Most are addressable right away, but let's decide together about the explicit mention of the "key" argument.

src/cr/cube/dimension.py
```diff
@@ -591,11 +591,17 @@ def _element_dicts(self):
...
+        selector = self._elements_transforms.get("selector")
+        default_id = element_dict["id"]
         if self._dimension_type in DT.ARRAY_TYPES:
-            # --- In case of array dimensions returns the id of the subvariable.
-            return element_dict.get("value", {}).get("id")
+            element_value = element_dict.get("value", {})
+            if selector == "alias":
+                # --- In case of specified key alias, returns the subvariable alias.
+                return element_value.get("references", {}).get("alias") or default_id
+            # --- Returns the subvariable id.
```
slobodan-ilic, line 601: Honestly, I'd prefer an `elif` here, just to be more explicit about the fact that we're waiting for a `"subvariable_id"` rather than `"fake_selector"`. Maybe even raise on `else` with the `ValueError("Unknown selector type %s" % selector)`

ernestoarbitrio: oh I cannot do the `ValueError`. Consider that now we do not support key spec in the frontend so what we receive today is `{"S2": {"hide": True}}` without the key spec. We need a fallback case. In future this can be implemented but now is not feasible IMHO

slobodan-ilic (approved):

PR #266, opened 2021-03-30
Additional share sum measures 177548812

ernestoarbitrio:
`Slice objects`

Rename:
- share_sum -> column_share_sum

Add:
- row_share_sum
- total_share_sum

### DIFF
--- a/src/cr/cube/cubepart.py
+++ b/src/cr/cube/cubepart.py
@@ -765,6 +765,22 @@ class _Slice(CubePartition):
         return Z_975 * self.row_std_err
 
     @lazyproperty
+    def row_share_sum(self):
+        """2D optional np.float64 ndarray of row share sum value for each table cell.
+
+        Raises `ValueError` if the cube-result does not include a sum cube-measure.
+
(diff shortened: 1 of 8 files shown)


src/cr/cube/cubepart.py
```diff
@@ -1145,6 +1162,22 @@ def table_weighted_bases(self):
         """2D np.float64 ndarray of table-proportion denominator for each cell."""
         return self._assembler.table_weighted_bases
 
+    @lazyproperty
+    def total_share_sum(self):
```
slobodan-ilic, line 1166: Are we sure we want the `total_` instead of (maybe) `table_`?

ernestoarbitrio: we do, we decide it with mike and joe, .... even the companion branch in zoom will have all the measure labels prefixed with `total` instead of `table` (i have to refactor them). So maybe in a nearly future I'm gonna refactor all the table_ measure names in cr.cube either

slobodan-ilic (approved):
lgtm

PR #271, opened 2021-04-07
Hypothesis testing mean 177647978

ernestoarbitrio:
Includes:
- Covariance cube measure (_Note this measure will be used to calc the hypothesis testing for means when numeric array is on the cols, so for now is not used but tested_)
- Pairwise t test for mean difference
- Pairwise indices for mean difference

Math ref: https://www.fon.hum.uva.nl/praat/manual/Covariance__Get_significance_of_means_difference___.html

### DIFF
--- a/src/cr/cube/cubepart.py
+++ b/src/cr/cube/cubepart.py
@@ -1460,6 +1460,34 @@ class _Strand(CubePartition):
     def name(self):
         return self.rows_dimension_name
 
+    def pairwise_significance_means_p_vals(self, column_idx):
+        """Optional 1D np.float64 ndarray of p-vals for means differences.
+
+        Raises ValueError when accessed on a cube-result that does not contain a mean
+        cube-measure.
+        """
+        try:
+            return self._assembler.pairwise_significance_means_p_vals(column_idx)
+        except ValueError:
+            raise ValueError(
+                "`.pairwise_significance_means_p_vals` is undefined for a cube-result "
+                "without a mean measure"
+            )
+
+    def pairwise_significance_means_t_stats(self, column_idx):
+        """Optional 1D np.float64 ndarray of t-stats for means differences.
+
+        Raises ValueError when accessed on a cube-result that does not contain a mean
+        cube-measure.
+        """
+        try:
+            return self._assembler.pairwise_significance_means_t_stats(column_idx)
+        except ValueError:
+            raise ValueError(
+                "`.pairwise_significance_means_t_stats` is undefined for a cube-result "
+                "without a mean measure"
+            )
+
     @lazyproperty
     def population_counts(self):
         """1D np.float64 ndarray of population count for each row of strand.
--- a/src/cr/cube/cubepart.py
+++ b/src/cr/cube/cubepart.py
@@ -676,6 +676,28 @@ class _Slice(CubePartition):
         # If overlaps are not defined, default to the old-way of calculation
         return PairwiseSignificance(self).values[column_idx].t_stats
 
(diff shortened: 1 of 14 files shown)


src/cr/cube/cubepart.py
```diff
@@ -1394,6 +1460,34 @@ def min_base_size_mask(self):
     def name(self):
         return self.rows_dimension_name
 
+    def pairwise_significance_means_p_vals(self, column_idx):
```
slobodan-ilic, line 1463: I can't see from diff, is this on `_Stripe`? If so, what does the `column_idx` mean (i.e. do we even need it on a stipe)?

src/cr/cube/matrix/measure.py
```diff
@@ -56,18 +56,36 @@ def means(self):
...
 
     def pairwise_t_stats_for_subvar(self, subvar_idx):
-        """_PairwiseSigTStats measure object for this cube-result and selected subvar."""
+        """_PairwiseSigTStats measure object for this cube-result and selected subvar"""
         return _PairwiseSigTStats(
             self._dimensions, self, self._cube_measures, subvar_idx
         )
 
+    def pairwise_significance_means_p_vals(self, column_idx):
+        """_PairwiseMeansSigPVals measure object for this cube-result."""
```
slobodan-ilic, line 78: Docstring should explain what `column_idx` is. Also, we're using `column_idx` in some places, and `subvar_idx` in others. Is this the `subvar_idx` or not? It's not clear on the first read (without going back to the description of the PR).

src/cr/cube/stripe/assembler.py
```diff
@@ -55,6 +55,24 @@ def means(self):
         """
         return self._assemble_vector(self._measures.means.blocks)
 
+    def pairwise_significance_means_p_vals(self, column_idx):
+        """1D optional np.float64 ndarray of mean difference p_vals for column idx.
```
slobodan-ilic, line 59: Why would we ever want this in 1D? I don't really get this part.

tests/integration/test_pairwise_significance.py
```diff
@@ -491,3 +491,151 @@ def test_pairwise_significance_mr_x_mr(self):
...
+            [(), (0,), (0,), (0,)],
+            [(1, 2, 3), (2,), (), (2,)],
+        ]
+        assert slice_.pairwise_means_indices_alt.tolist() == [
+            [(), (), (0, 1), (0, 1)],
+            [(), (0,), (0, 3), (0,)],
+            [(1, 2, 3), (2,), (), (2,)],
+        ]
+
+    def but_it_raises_an_error_when_mean_diff_significance_is_not_available(self):
```
slobodan-ilic, line 562: The name of the test should follow either TDD (i.e. starts with `test_`) or BDD, starting with (`it_`, `but_`, `and_`). At least while we're in the same file, but we should also extend this to the entire library eventually. Imagine trying to find all tests with `:vimgrep /test_/ cube/tests/**/*.py` and not succeeding :)

src/cr/cube/stripe/assembler.py
ernestoarbitrio: Yes u right this is not needed ... I'll delete it. Sorry my mistake

src/cr/cube/cubepart.py
ernestoarbitrio: this has to be deleted, and btw it exists in the strand ... but nevermind this is not useful ... i;ll get rid of it

slobodan-ilic (approved):

PR #272, opened 2021-04-12
fix: overlaps

slobodan-ilic:
* rfctr architecture wrt/ row_id
* add test case from alpha

### DIFF
--- a/src/cr/cube/matrix/measure.py
+++ b/src/cr/cube/matrix/measure.py
@@ -521,8 +521,9 @@ class _PairwiseSigTStats(_BaseSecondOrderMeasure):
             [
                 [
                     _PairwiseSignificaneBetweenSubvariablesHelper(
-                        self._cube_measures.cube_overlaps.overlaps[row_idx],
-                        self._cube_measures.cube_overlaps.valid_overlaps[row_idx],
+                        self._cube_measures.cube_overlaps.overlaps,
+                        self._cube_measures.cube_overlaps.valid_overlaps,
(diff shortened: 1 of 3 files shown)


src/cr/cube/matrix/measure.py
```diff
@@ -1239,8 +1242,9 @@ def t_stats(self):
         Sa, Sb, Sab = self._selected_counts
         Na, Nb, Nab = self._valid_counts
         pa, pb, pab = Sa / Na, Sb / Nb, Sab / Nab
+        pct_a, pct_b = self._proportions
```
ernestoarbitrio, line 1245: i think that `pct` prefix should be changed in something that recall the proportions cause pct can be interpreted as percentages that is basically prop * 100.

src/cr/cube/matrix/measure.py
```diff
@@ -1253,6 +1257,14 @@ def _df(self):
         Na, Nb, Nab = self._valid_counts
         return Na + Nb - Nab
 
+    @lazyproperty
+    def _proportions(self):
+        """tuple(float64) of column proportions"""
```
ernestoarbitrio, line 1262: I would expand the docstring of this property, TBH I'd like to see what `pa` and `pb` are  also a description on these `pa/Sa` and `pb/Sb` divisions

ernestoarbitrio (changes requested):
per comments, but looks good afterwards

ernestoarbitrio (approved):
LGTM

PR #274, opened 2021-04-14
pairwise indices with overlaps

slobodan-ilic:


src/cr/cube/cubepart.py
```diff
@@ -614,6 +614,15 @@ def pairwise_indices(self):
...
         """
+        # If overlaps are defined, calculate significance based on them
+        if (
+            self.dimension_types[-1] == DT.MR
+            and self._cube.overlaps is not None
+            and self._cube.valid_overlaps is not None
+        ):
+            return self._assembler.pairwise_indices(self._alpha, self._only_larger)
+
+        # If overlaps are not defined, default to the legacy way of calculating sig
```
ernestoarbitrio, line 625: I think that this change is needed also in the `pairwise_indices_alt` to enable the 2 level (aBCdeF) pairwise sig letters

tests/integration/test_pairwise_significance.py
```diff
@@ -571,6 +571,19 @@ def test_pairwise_cat_x_mr_gender_x_all_pets_owned(self):
...
+        assert slice_.column_percentages.tolist() == [
+            [75.0, 20.0, 60.0],
+            [25.0, 80.0, 40.0],
+        ]
+
+        assert slice_.pairwise_indices.tolist() == [
+            [(), (0, 2), ()],
+            [(1,), (), (1,)],
+        ]
+
```
ernestoarbitrio, line 586: I'd add a case with the 2 level of significance

ernestoarbitrio (changes requested):
per comments

src/cr/cube/cubepart.py
slobodan-ilic: done

tests/integration/test_pairwise_significance.py
slobodan-ilic: done

tests/integration/test_pairwise_significance.py
```diff
@@ -571,6 +571,27 @@ def test_pairwise_cat_x_mr_gender_x_all_pets_owned(self):
...
+        assert slice_.pairwise_indices_alt.tolist() == [
+            [(), (0, 2), ()],
+            [(1,), (), (1,)],
+        ]
```
ernestoarbitrio, line 593: could you use a sig level alt like 0.001 to see if the indices are different? Just to be sure that that method is doing something different

ernestoarbitrio (changes requested):
1 minor comment than LGTM

tests/integration/test_pairwise_significance.py
slobodan-ilic: done

ernestoarbitrio (approved):

PR #280, opened 2021-04-19
Weighted proportions for overlaps 177801573

slobodan-ilic:
Per [ticket](https://www.pivotaltracker.com/story/show/177801573)


tests/integration/test_pairwise_significance.py
```diff
@@ -512,25 +512,29 @@ def test_pairwise_significance_mr_x_mr(self):
     def test_pairwise_cat_x_mr_gender_x_all_pets_owned(self):
```
ernestoarbitrio, line 512: I would use the suffix `_weighted` where it's needed in the test name. This way we understand that we are testing the cube with a weight

src/cr/cube/matrix/measure.py
```diff
@@ -565,6 +566,7 @@ def p_vals(self):
             [
                 [
                     _PairwiseSignificaneBetweenSubvariablesHelper(
+                        self._cube_measures.weighted_cube_counts.weighted_counts,
```
ernestoarbitrio, line 569: what if `weighted_counts`  is `None`? This is an optional cube measure and if it doesn't exists its value is None. Do we have a test that exercises this case?

ernestoarbitrio (changes requested):
Per comments

ernestoarbitrio (approved):

PR #284, opened 2021-04-20
Pop size cat dates 177831399

slobodan-ilic:


src/cr/cube/cubepart.py
```diff
@@ -775,9 +775,14 @@ def pairwise_significance_tests(self):
...
     @lazyproperty
     def population_counts(self):
-        return (
-            self.table_proportions * self._population * self._cube.population_fraction
+        proportions = (
+            self.row_proportions
+            if self._dimensions[0].is_cat_date
+            else self.column_proportions
+            if self._dimensions[1].is_cat_date
+            else self.table_proportions
```
ernestoarbitrio, line 783: I don't know if checking the dimension type this way is the best approach.
1. at the `_Slice` level we already have `_rows_dimension` and `_columns_dimension` so it would be more explicit instead of `self._dimensions[0] or [1]`
2. I wonder if the cat date type can be a real dimension type and so do the check like:
```
if self.rows_dimension_type == DT.CAT_DATE:
```
maybe in here https://github.com/Crunch-io/crunch-cube/blob/a1b615593281a9cdcdb8c87e708bccaabe139422/src/cr/cube/dimension.py#L275

src/cr/cube/cubepart.py
```diff
@@ -775,9 +775,14 @@ def pairwise_significance_tests(self):
 
     @lazyproperty
     def population_counts(self):
```
ernestoarbitrio, line 777: Even if the porperty doesn't have a doctring, considering that you touched it you should provide a meaningful docstring also because it introduces new business logic

src/cr/cube/dimension.py
```diff
@@ -416,6 +416,16 @@ def hidden_idxs(self):
...
+    @lazyproperty
+    def is_cat_date(self):
+        """True if the dimension is a categorical date, False otherwise.
+
+        A dimension is considered a categorical date if it:
+        1. is a categorical dimension in the first place and
+        2. if it has a special entry "date" in the dimension dictionary.
+        """
+        return any(el.has_date for el in self.all_elements)
+
```
ernestoarbitrio, line 428: per previous comment maybe this logic can be moved al the dimension type inspection level

ernestoarbitrio (changes requested):
per comments

src/cr/cube/cubepart.py
malecki: what about MR × date? is the date as columns still dimension 1 if there's a selections dimension?

slobodan-ilic: no, @malecki in that case it's hidden, because `_dimensions` only has the visible ones. But I agree with the first @ernestoarbitrio 's comment, I'll make it more explilcit. About the special type, though, @ernestoarbitrio I thought about that too, but I'm afraid what other things it might break, since it's definitely a CAT dimension, and needs to behave like one. And since we're doing it with enums, and not classes, I can't simply inherit from the base class and have mostly the same behavior... Anyways, I'll check...

src/cr/cube/cubepart.py
slobodan-ilic: hehe, this is a revenge I see :)

src/cr/cube/dimension.py
slobodan-ilic: yeah, if we can easily create a new type, sure. But I'm afraid of that new type, that we'll have to sprinkle a bunch of 
```
if dimension_type in (DT.CAT, DT.CAT_DATE):
   . . .
```
where we wouldn't really care in the first place...

src/cr/cube/cubepart.py
ernestoarbitrio: ok i see

src/cr/cube/cubepart.py
ernestoarbitrio: ahahaha no it is not :D it's steve's lessons

src/cr/cube/cubepart.py
slobodan-ilic: Ok, I ended up refactoring it according to your recommendation. It seems to be working fine with exporter too...

src/cr/cube/cubepart.py
slobodan-ilic: ❤️

slobodan-ilic: Addressed

src/cr/cube/dimension.py
slobodan-ilic: ok, refactored...

src/cr/cube/dimension.py
```diff
@@ -241,6 +241,15 @@ def _is_array_cat(self):
         """
         return "subreferences" in self._dimension_dict["references"]
 
+    @lazyproperty
+    def _is_cat_date(self):
```
ernestoarbitrio, line 245: docstring

src/cr/cube/cubepart.py
```diff
@@ -775,9 +775,26 @@ def pairwise_significance_tests(self):
...
+
+        If any of the dimensions (rows or columns) is a categorical date, the
+        appropriate percentages are used, to calculate the population counts, so as to
+        be the total amount among categorical dates.
+
+        Otherwise, table percents are used to calculate population counts.
+        """
+        proportions = (
+            self.row_proportions
+            if self._rows_dimension.dimension_type == DT.CAT_DATE
```
ernestoarbitrio, line 792: at slice level we have `rows_dimension_type` property ... so you can use that one

ernestoarbitrio (changes requested):
per comments

ernestoarbitrio (approved):
LGTM but i'd like to be sure that the `population_counts_moe` for strand is correct (0) in case of CAT_DATE dimension.

PR #286, opened 2021-04-22
spike: plug FREQUENCY sort-by-value gap

scanny:
Wire up _Strand sort-by-value for "univariate-measure" keyword case.

This should fix the existing alpha-Sentry error on sort-by-value for FREQUENCY analyses (aka. 1D card, _Strand).

I'll continue work on another branch to add detailed unit tests for these and add the remaining measures and so forth.


ernestoarbitrio (approved):
LGTM

PR #288, opened 2021-04-23
[#177690013]: Refactor hypothesis testing

ernestoarbitrio:
Refactor the pairwise significance computation. This needs for exporter to avoid showing color shading (pval) for subtotal of differences


tests/unit/matrix/test_subtotals.py
```diff
@@ -200,6 +201,102 @@ def it_can_compute_a_subtotal_row_to_help(self, request):
...
+    ):
+        subtotal_.addend_idxs = addend_idxs
+        subtotal_.subtrahend_idxs = subtrahend_idxs
+        property_mock(request, PairwiseSigTestSubtotals, "_nrows", return_value=3)
+        subtotals = PairwiseSigTestSubtotals(
+            None, None, None, None, 0, diff_cols_nan, None
+        )
+        column = np.array([1, 2])
+
+        assert subtotals._subtotal_column(subtotal_, column) == pytest.approx(
```
slobodan-ilic, line 252: Maybe put the actual call on a separate line, before the assertion, so it's more easily detectable.

tests/unit/matrix/test_subtotals.py
```diff
@@ -200,6 +201,102 @@ def it_can_compute_a_subtotal_row_to_help(self, request):
...
+    ):
+        subtotal_.addend_idxs = addend_idxs
+        subtotal_.subtrahend_idxs = subtrahend_idxs
+        property_mock(request, PairwiseSigTestSubtotals, "_ncols", return_value=3)
+        subtotals = PairwiseSigTestSubtotals(
+            None, None, None, None, 0, None, diff_rows_nan
+        )
+        row = np.array([1, 2])
+
+        assert subtotals._subtotal_row(subtotal_, row) == pytest.approx(
```
slobodan-ilic, line 281: same here

tests/unit/matrix/test_subtotals.py
```diff
@@ -200,6 +201,102 @@ def it_can_compute_a_subtotal_row_to_help(self, request):
...
+
+    def it_can_compute_a_intersection_cell_value_to_help(self):
+        assert np.isnan(
+            PairwiseSigTestSubtotals(None, None, None, None, 0)._intersection(
+                None, None
+            )
+        )
+
+    @pytest.mark.parametrize(
+        ("addend_idxs", "subtrahend_idxs", "diff_cols_nan", "expected_value"),
```
slobodan-ilic, line 228: I don't really understand the `diff_cols_nan` and what it's supposed to mean... But this is just at the first glance, I might be able to understand it more after thorough review.

slobodan-ilic: Also, expected value for what? I'm not clear on how the `_subtotal_column` is supposed to work. Its API is a bit weird, where you have to pass it the subtotal and a column... I hope I'll understand it better by the time I finish with the review.

tests/unit/matrix/test_measure.py
```diff
@@ -115,6 +119,40 @@ def it_provides_access_to_the_cube_measures_to_help(
...
+            request,
+            "cr.cube.matrix.measure.%s" % MeasureCls.__name__,
+            return_value=measure_,
+        )
+        _cube_measures_prop_.return_value = cube_measures_
+        measures = SecondOrderMeasures(None, dimensions_, None)
+        measure = getattr(measures, measure_prop_name)
+        col_idx = 0
+
+        assert measure(col_idx) is measure_
```
slobodan-ilic, line 151: This is a bit convoluted. I had a hard time figuring out what is what and what you're trying to assert. I'd prefer if instead of `measure` you'd use something more similar to `MeasureCls`, or something more appropriate. The main issue is that I don't understand what the `measure(col_idx)` call is supposed to be doing, because I don't know what `measure` is supposed to be. Also, while you're at it, please separate out that call on its own line, and do the assertions afterwards.

tests/integration/test_pairwise_significance.py
```diff
@@ -219,6 +194,193 @@ def test_cat_x_cat_hs_pairwise_t_tests(self):
...
+        )
+        assert slice_.pairwise_significance_p_vals(1)[:, 0] == pytest.approx(
+            np.full(6, np.nan), nan_ok=True
+        )
+
+        # Testing select a base col (1)
+        assert slice_.pairwise_significance_t_stats(6) == pytest.approx(
+            np.array(
+                [
+                    [np.nan, np.nan, np.nan, np.nan, np.nan, np.nan, np.nan],
```
slobodan-ilic, line 329: The instant suspicion is: "will it work on other positions than `top` and `bottom`?". I'd like to see a "cross" (like a sniper target), rather than the "angle ruler" :)

src/cr/cube/stripe/insertion.py
```diff
@@ -61,6 +61,15 @@ class SumSubtotals(_BaseSubtotals):
     bases are additive even across subtrahends of subtotals.
```
slobodan-ilic, line 61: Can you please also document in this docstring what the `diffs_nan` is used for? I can't figure it out by inspecting the changes. I can guess that its the case where you "don't want to add the specific measure values for the differences", but even so, I'd like to know when that applies and why.

src/cr/cube/matrix/assembler.py
```diff
@@ -118,16 +118,13 @@ def column_weighted_bases(self):
     def columns_base(self):
         """1D/2D np.float64 ndarray of unweighted-N for each slice column/cell."""
         # --- an MR_X slice produces a 2D column-base (each cell has its own N) ---
-        rows_dim_type = self._rows_dimension.dimension_type
         columns_base = self._measures.columns_base
-        if rows_dim_type in (DT.MR_SUBVAR, DT.NUM_ARRAY):
-            return self._assemble_matrix(
-                SumSubtotals.blocks(columns_base, self._dimensions, diff_cols_nan=True)
-            )
+        if self._rows_dimension.dimension_type in (DT.MR_SUBVAR, DT.NUM_ARRAY):
```
slobodan-ilic, line 122: This remark is not related to your PR, but what would happen if you were to add `DT.CA_SUBVAR` here? How many tests would break, if any? Because I think it might be the right thing to do (although I don't immediately know what the `_assemble_matrix` would do in such case). The main point being - columns base are not really "a thing" over a set of subvariables (rows dimension).

src/cr/cube/matrix/measure.py
```diff
@@ -85,17 +87,29 @@ def pairwise_means_indices(self, alpha, only_larger):
         )
 
     def pairwise_p_vals_for_subvar(self, subvar_idx):
-        """_PairwiseSigPVals measure object for this cube-result and selected subvar"""
-        return _PairwiseSigPVals(
+        """_PairwiseSubvarSigPVals measure object for this cube-result"""
+        return _PairwiseSubvarSigPVals(
```
slobodan-ilic, line 91: Maybe consider a different name for this class, something like: `_PairwiseSigPValsForSubvar`. It's more explicit, and corresponds to the method name better (at least in my mind). The current one just sounds as a bunch of different words stacked together 😸

src/cr/cube/matrix/measure.py
```diff
@@ -85,17 +87,29 @@ def pairwise_means_indices(self, alpha, only_larger):
...
             self._dimensions, self, self._cube_measures, subvar_idx
         )
 
     def pairwise_t_stats_for_subvar(self, subvar_idx):
-        """_PairwiseSigTStats measure object for this cube-result and selected subvar"""
-        return _PairwiseSigTStats(
+        """_PairwiseSubvarSigTStats measure object for this cube-result"""
```
slobodan-ilic, line 96: same here

src/cr/cube/matrix/subtotals.py
```diff
@@ -593,3 +593,154 @@ def _subtotal_row(self, subtotal):
...
+        dimensions,
+        column_proportions,
+        columns_base,
+        selected_column_idx,
+        diff_cols_nan=True,
+        diff_rows_nan=True,
+    ):
+        super(PairwiseSigTestSubtotals, self).__init__(base_values, dimensions)
+        self._column_proportions = column_proportions
+        self._columns_base = columns_base
```
slobodan-ilic, line 613: Can't you just pass `second_order_measures` here, instead of each measure separately? I'm talking about the 2 ones right above this comment. It'd be less messy if you could do that.

tests/unit/matrix/test_subtotals.py
ernestoarbitrio: Basically when the `diff_cols_nan` is True all columns and rows subtotals will be nan. This means that for `PairwiseSigTestSubtotals` if the param is True all the cols or row have to be full of np.nan that are dashes in the exporter. So for each column i need to pass the subtotal and the column itself cause we have the cases that we have to return the col itself of nan if diff_cols_nan is true and if the subtotal has subtraends

src/cr/cube/matrix/assembler.py
ernestoarbitrio: yes, i do agree and I think this is related to what you discovered recently on cube. 🤷🏽‍♂️

src/cr/cube/matrix/subtotals.py
ernestoarbitrio: good catch :D

slobodan-ilic (approved):

PR #290, opened 2021-04-26
[#177898553]: fix weighted issue

ernestoarbitrio:


src/cr/cube/cube.py
```diff
@@ -492,6 +492,8 @@ def weighted_counts(self):
         weighted_counts = (
             self._measures.weighted_valid_counts
             if self._measures.weighted_valid_counts is not None
+            else self._measures.unweighted_valid_counts
+            if self._measures.unweighted_valid_counts is not None
             else self._measures.weighted_counts
```
slobodan-ilic, line 497: If it somehow gets to this part, it will still blow up.

PR #292, opened 2021-04-26
Don't show scale median on row/column of subotal difference

gergness:
Can't calculate the scale median on the row of a row subtotal difference or the column of a column subtotal difference.

### DIFF
--- a/src/cr/cube/cubepart.py
+++ b/src/cr/cube/cubepart.py
@@ -499,7 +499,9 @@ class _Slice(CubePartition):
 
         not_a_nan_index = ~np.isnan(self._rows_dimension_numeric_values)
         numeric_values = self._rows_dimension_numeric_values[not_a_nan_index]
-        counts = np.nan_to_num(self.counts[not_a_nan_index, :]).astype("int64")
+        counts = np.nan_to_num(
+            self._assembler.column_comparable_counts[not_a_nan_index, :]
+        ).astype("int64")
         scale_median = np.array(
             [
                 self._median(np.repeat(numeric_values, counts[:, i]))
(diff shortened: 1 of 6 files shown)


scanny (approved):
In general, we should be driving all non-trivial measure computation down, out of _Slice and _Strand and into second-order measure objects.

This work is incomplete, for example `.rows/columns_scale_median` still does computation on assembled measures. This is because we needed to draw a scope line somewhere on the large measure-consolidation effort and we intend to move these incrementally. The work on pushing scaled-count measures (scale-mean, scale-median, etc.) down is more advanced in `_Stripe`, which may provide something of a model.

Anyway, there's an opportunity here to advance that work by moving at least part of this down there, but I'll understand if we don't want to take this moment to undertake that. But just so you know what the architectural vision is, the `cr.cube.cubepart` bits of this would be one-liners like `return self._assembler.columns_scale_median` and there would be possibly a single `cr.cube.matrix.measure` object, perhaps `_ScaledCounts` that would encapsulate all the bits about knowing the numeric values and scaling counts for the purposes of computing these marginals. It would probably make sense to move all the `scale-x` related measures at once since they share a knowledge of numeric-values and some depend on others in that family.

This looks okay as a patch, so if we're in a hurry I'm fine with this, but in general I think we should be advancing measure consolidation at every opportunity.

gergness:
Hm, consolidating the scale measures is harder than I originally thought. It's different from all existing measures for two reasons:

1) The behavior for subtotals is different from existing measures. The "row scale median" of a subtotal (not a subdiff) is defined, but it's not simply the addition of the components, it must be calculated from the counts of the components. (eg `f(x +y) != f(x) + f(y)` in this case, but I believe all other measures defined so far are communitive in this way).
2) Maybe not important, but I think they would be the first second order measures that are 1D.

PR #293, opened 2021-04-27
[#177923295]: Pop fraction for CAT DATE

slobodan-ilic:
* Drive by TDD, create a unit test that expects correct behavior for the
  population fraction, based on the `is_cat_date` filter stats setting
* Implement correct functionality

### DIFF
--- a/src/cr/cube/cube.py
+++ b/src/cr/cube/cube.py
@@ -734,16 +734,18 @@ class _Measures(object):
 
         # Try and get the new-style complete-cases filtered counts
         filter_stats = self._cube_dict["result"].get("filter_stats", {})
-        weighted = filter_stats.get("filtered_complete", {}).get("weighted")
-        if weighted:
+        weighted_filtered_complete = filter_stats.get("filtered_complete", {}).get(
+            "weighted"
+        )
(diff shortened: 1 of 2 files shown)


src/cr/cube/cube.py
```diff
@@ -733,18 +733,19 @@ def population_fraction(self):
...
-            .get("filtered_complete", {})
-            .get("weighted")
-        )
-        if filter_stats:
-            # If new format is present in response json, use that for pop fraction
-            numerator = filter_stats["selected"]
-            denominator = numerator + filter_stats["other"]
+        filter_stats = self._cube_dict["result"].get("filter_stats", {})
+        weighted = filter_stats.get("filtered_complete", {}).get("weighted")
+        if weighted:
```
ernestoarbitrio, line 738: i don't like this check on is weighted. I mean the name is odd to me. It seems that we're checking for something weighted but actually we are doing a check on if filter_stats new format are present do something otherwise do other. So i would choose another name for that var, maybe `weighted_filtered_complete` ?

ernestoarbitrio (changes requested):
per comments

ernestoarbitrio (approved):

PR #294, opened 2021-04-28
Set weighted N to - for subdifference columns

gergness:
exporter used the `*_margin` both as the "Weighted N" in the summary row and as the numerator for calculating the proportions in the "All" column.

In the first case, we want to set to nan for subtotal difference columns because all other measures are set to nan.

But in the second case, we want to keep the number.

Fix by calculating the `*_margin_proportion` in cr.cube and using that in exporter. I can also changing by having two kinds of "`margin`s", but I thought it'd be hard to understand their difference (especially since the difference between base and margin is already a little fuzzy)

### DIFF
--- a/tests/unit/matrix/test_assembler.py
+++ b/tests/unit/matrix/test_assembler.py
@@ -34,6 +34,7 @@ from cr.cube.matrix.measure import (
 )
 
 from ...unitutil import (
+    ANY,
     class_mock,
     instance_mock,
     method_mock,
@@ -233,7 +234,7 @@ class DescribeAssembler(object):
         columns_margin = assembler.columns_margin
 
         _assemble_vector_.assert_called_once_with(
-            assembler, [1, 2, 3], [3, 5], [0, -2, 1, 2, -1]
+            assembler, [1, 2, 3], [3, 5], [0, -2, 1, 2, -1], diffs_nan=True
         )
         assert columns_margin.tolist() == [1, 3, 2, 3, 5]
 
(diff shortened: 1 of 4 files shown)


tests/unit/matrix/test_assembler.py
```diff
@@ -230,7 +231,69 @@ def but_it_provides_a_2D_columns_margin_for_an_MR_X_cube_result(
...
+        cube_result_matrix_.table_margin = 6.0
+        cube_result_matrix_.columns_margin = np.array([1, 2, 3])
+        _column_subtotals_prop_.return_value = [3, 5]
+        _column_order_prop_.return_value = [0, -2, 1, 2, -1]
+        _assemble_vector_.return_value = [1, 3, 2, 3, 5]
+        assembler = Assembler(None, None, None)
+
+        columns_margin = assembler.columns_margin_proportion
+
+        _assemble_vector_.assert_called_once_with(
```
gergness, line 262: I had trouble with this, and makes me think there may be a better way to write this test (or the code underlying). `cube_result_matrix_.columns_margin` must be an `np.ndarray` instead of a list because it gets divided and `[1, 2, 3] / 6.0` gives error about not division not being defined on list & int. But then with the `ndarray` the `assert_called_once_with` struggles to compare.

scanny: This is exactly how I would solve it, use `ANY` for the offending args and isolate those args for separate testing with `method_x_.call_args.args[i]`.

`.assert_called_once_with()` is a convenience matcher, just frequently saving us from doing this with every single call-arg, but when it won't work like in this (pretty rare) case, we need to do that work for ourselves. Looks fine to me, although of course a little more work to read, but whatever right?, sometimes code takes some work to interpret, our job is just to avoid _unnecessary_ work for the reader.

scanny (approved):
LGTM. A comment or two for informational purposes.

PR #295, opened 2021-04-28
[#177917018]: move the whole logic of valid counts in the actual weighted and unweighted counts

ernestoarbitrio:
`valid_counts` (weighted or unweighted) basically are equal to `counts` and `unweighted_counts` when the payload has these measures inside. This because it's up to cube to get the proper counts in it's measures.

This PR removes all the valid counts logic

### DIFF
--- a/src/cr/cube/matrix/measure.py
+++ b/src/cr/cube/matrix/measure.py
@@ -152,25 +152,13 @@ class SecondOrderMeasures(object):
         """_TotalShareSum measure object for this cube-result"""
         return _TotalShareSum(self._dimensions, self, self._cube_measures)
 
-    @lazyproperty
-    def unweighted_counts(self):
+    def unweighted_counts(self, diffs_nan):
         """_UnweightedCounts measure object for this cube-result."""
-        return _UnweightedCounts(self._dimensions, self, self._cube_measures)
+        return _UnweightedCounts(self._dimensions, self, self._cube_measures, diffs_nan)
 
-    @lazyproperty
-    def unweighted_valid_counts(self):
-        """_UnweightedValidCounts measure object for this cube-result."""
-        return _UnweightedValidCounts(self._dimensions, self, self._cube_measures)
-
-    @lazyproperty
-    def weighted_counts(self):
+    def weighted_counts(self, diffs_nan):
         """_WeightedCounts measure object for this cube-result."""
-        return _WeightedCounts(self._dimensions, self, self._cube_measures)
-
-    @lazyproperty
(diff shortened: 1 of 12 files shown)


src/cr/cube/matrix/measure.py
```diff
@@ -152,25 +152,13 @@ def total_share_sum(self):
         """_TotalShareSum measure object for this cube-result"""
         return _TotalShareSum(self._dimensions, self, self._cube_measures)
 
-    @lazyproperty
-    def unweighted_counts(self):
+    def unweighted_counts(self, diffs_nan):
```
slobodan-ilic, line 155: What is `diffs_nan`? Please improve the docstring.

src/cr/cube/matrix/measure.py
```diff
@@ -152,25 +152,13 @@ def total_share_sum(self):
...
         """_UnweightedCounts measure object for this cube-result."""
-        return _UnweightedCounts(self._dimensions, self, self._cube_measures)
-
-    @lazyproperty
-    def unweighted_valid_counts(self):
-        """_UnweightedValidCounts measure object for this cube-result."""
-        return _UnweightedValidCounts(self._dimensions, self, self._cube_measures)
+        return _UnweightedCounts(self._dimensions, self, self._cube_measures, diffs_nan)
```
slobodan-ilic, line 157: here too

src/cr/cube/matrix/measure.py
```diff
@@ -152,25 +152,13 @@ def total_share_sum(self):
...
 
-    @lazyproperty
-    def weighted_counts(self):
+    def weighted_counts(self, diffs_nan):
```
slobodan-ilic, line 159: here too

src/cr/cube/matrix/measure.py
```diff
@@ -1228,6 +1216,14 @@ def blocks(self):
 class _UnweightedCounts(_BaseSecondOrderMeasure):
     """Provides the unweighted-counts measure for a matrix."""
 
+    def __init__(
+        self, dimensions, second_order_measures, cube_measures, diff_nans=False
```
slobodan-ilic, line 1220: Why is there a default value here, for diff_nans, but not in the other class?

slobodan-ilic (approved):
Seems good, just fix the docstring

src/cr/cube/matrix/measure.py
ernestoarbitrio: this is wrong ... my bad

PR #296, opened 2021-04-30
Bug pairwise mean indices 177973334

ernestoarbitrio:
Avoid ravel (as @slobodan-ilic original idea) in pairwise mean indices, to respect the original shape in case of array of empty tuples


src/cr/cube/matrix/assembler.py
```diff
@@ -194,11 +194,16 @@ def pairwise_means_indices(self, alpha, only_larger):
 
         return np.array(
             [
-                tuple(col_ord_map[idx] for idx in idxs) if idxs is not None else None
-                for idxs in assembled_matrix.ravel()
+                [
+                    tuple(col_ord_map[idx] for idx in idxs)
```
slobodan-ilic, line 198: If we changed the name of `col_ord_map` to just `ord_map`, so that it's a bit shorter, the display of the list comprehension would be much more pretty.

slobodan-ilic (approved):

PR #297, opened 2021-05-03
[#177998734]: zscore measure consolidation

ernestoarbitrio:


tests/unit/matrix/test_cubemeasure.py
```diff
@@ -1193,6 +1193,26 @@ def it_computes_the_column_bases(self, request, columns_margin, expected_value):
...
+        np.testing.assert_almost_equal(
+            residuals,
+            np.array(
+                [
+                    [-0.9521905, -0.3555207, -0.2275962, -0.1660169],
+                    [0.0, 0.0, 0.0, 0.0],
+                    [0.2762585, 0.1951985, 0.1485686, 0.1184429],
+                ]
+            ),
+        )
```
scanny, line 1214: This can be:

```python
assert pytest.approx(residuals) == [
    [-0.9521905, -0.3555207, -0.2275962, -0.1660169],
    [0.0, 0.0, 0.0, 0.0],
    [0.2762585, 0.1951985, 0.1485686, 0.1184429],
]
```

scanny (changes requested):
This is definitely on the right track. Just one change of substance, also check your unit coverage and fill in any gaps, maybe also pick up a couple unit-test coverage gaps to start working that toward zero.

tests/unit/matrix/test_cubemeasure.py
```diff
@@ -1316,10 +1345,22 @@ def it_knows_its_weighted_counts(self, raw_weighted_counts):
             )
         )
 
+    def it_knows_its_zscores(self, raw_weighted_counts):
+        weighted_cube_counts = _CatXMrWeightedCubeCounts(None, raw_weighted_counts)
+        assert weighted_cube_counts.zscores.tolist() == [
+            [-1.7549928774784245, -1.6818357317441646, -1.754992877478425],
+            [1.7549928774784245, 1.6818357317441637, 1.754992877478425],
+        ]
```
scanny, line 1353: This will not pass reliably because it tests floats for equality. Use `pytest.approx()` on the left side and then you can drop the `.tolist()` as well. The `.tolist()` pattern only applies for integer results.

tests/unit/matrix/test_cubemeasure.py
```diff
@@ -1316,10 +1345,22 @@ def it_knows_its_weighted_counts(self, raw_weighted_counts):
...
+
+    def it_does_not_calculate_zscores_when_matrix_is_defective(self):
+        raw_weighted_counts = np.array([[[1, 1], [2, 2]]])
+        weighted_cube_counts = _CatXMrWeightedCubeCounts(None, raw_weighted_counts)
+        assert weighted_cube_counts.zscores.tolist() == [[0, 0]]
```
scanny, line 1358: Why are these zeroes instead of NaNs if this array is defective? A 2 x 2 array of integer values is not defective. Also, these are weighted counts and so are floats; better to use 1.1, 2.2 etc. rather than integers. It's just as easy to get it right and more closely approximates the runtime values.

tests/unit/matrix/test_cubemeasure.py
```diff
@@ -1316,10 +1345,22 @@ def it_knows_its_weighted_counts(self, raw_weighted_counts):
...
+
     # fixtures -------------------------------------------------------
 
     @pytest.fixture
-    def raw_weighted_counts(self, request):
+    def raw_weighted_counts(self):
```
scanny, line 1363: 👍

tests/unit/matrix/test_cubemeasure.py
```diff
@@ -1798,35 +1825,6 @@ def it_knows_its_weighted_counts(self):
...
-                np.array([[1, 2], [0, 0]]),
-                np.array([[np.nan, np.nan], [np.nan, np.nan]]),
-            ),
-            (
-                np.array([[0, 2], [0, 4]]),
-                np.array([[np.nan, np.nan], [np.nan, np.nan]]),
-            ),
-        ),
-    )
-    def but_its_zscores_are_NaNs_for_a_deficient_matrix(
```
scanny, line 1823: This is a better test title than `it_does_not_calculate_zscores_when_matrix_is_defective`. It does in fact return a value and this title spells out what that is and why.

tests/unit/matrix/test_measure.py
```diff
@@ -1290,4 +1337,49 @@ def it_computes_its_blocks_to_help(self, request, dimensions_):
...
     @pytest.fixture
     def dimensions_(self, request):
-        return (instance_mock(request, Dimension), instance_mock(request, Dimension))
+        return instance_mock(request, Dimension), instance_mock(request, Dimension)
+
+
+class Describe_Zscores(object):
+    """Unit test suite for `cr.cube.matrix.measure._Zscores` object."""
+
+    def it_computes_zscore_subtotals_blocks_to_help(self, request, dimensions_):
```
scanny, line 1346: `ZscoreSubtotals.blocks` is an interface method (not a helper) and so the `_to_help` suffix is not appropriate for this test (or the next one).

tests/unit/matrix/test_measure.py
```diff
@@ -1290,4 +1337,49 @@ def it_computes_its_blocks_to_help(self, request, dimensions_):
...
+        weighted_cube_counts_ = instance_mock(request, _BaseWeightedCubeCounts)
+        _weighted_cube_counts = property_mock(
+            request, _BaseSecondOrderMeasure, "_weighted_cube_counts"
+        )
+        _weighted_cube_counts.return_value = weighted_cube_counts_
```
scanny, line 1351: 1. This property mock should be on `_Zscores`, not its base class.
2. The return value can be specified in the `property_mock()` call as a keyword parameter `return_value=weighted_cube_counts_.
3. The name should be `_weighted_cube_counts_prop_`, however no name will be needed when you move the return value into the mock call and it can just start out as `property_mock(...`.

tests/unit/matrix/test_measure.py
```diff
@@ -1290,4 +1337,49 @@ def it_computes_its_blocks_to_help(self, request, dimensions_):
...
+        zscores = _Zscores(dimensions_, None, None)
+
+        blocks = zscores.blocks
+
+        ZscoreSubtotals_.blocks.assert_called_once_with(
+            weighted_cube_counts_, dimensions_
+        )
+        assert blocks == [[[1], [2]], [[3], [4]]]
+
+    def it_computes_nan_subtotals_blocks_to_help(self, request, dimensions_):
```
scanny, line 1363: Maybe `but_the_subtotal_blocks_are_NaNs_when_an_MR_dimension_is_present()`.

tests/unit/matrix/test_measure.py
```diff
@@ -1290,4 +1337,49 @@ def it_computes_its_blocks_to_help(self, request, dimensions_):
...
+        weighted_cube_counts_ = instance_mock(
+            request, _BaseWeightedCubeCounts, zscores=[1, 2]
```
scanny, line 1365: maybe `zscores=[[1, 2], [3, 4]]` just so the reader doesn't have to puzzle over why these would be 1D.

tests/unit/matrix/test_measure.py
```diff
@@ -1290,4 +1337,49 @@ def it_computes_its_blocks_to_help(self, request, dimensions_):
...
+        )
+        _weighted_cube_counts = property_mock(
+            request, _BaseSecondOrderMeasure, "_weighted_cube_counts"
+        )
+        _weighted_cube_counts.return_value = weighted_cube_counts_
```
scanny, line 1370: same as above

tests/unit/matrix/test_subtotals.py
```diff
@@ -519,17 +519,27 @@ def _table_margin_prop_(self, request):
...
+            request,
+            ZscoreSubtotals,
+            "_intersections",
+            return_value=np.array([[1, 2], [3, 4]]),
+        )
+
+        intersections = ZscoreSubtotals.intersections(_base_weighted_counts_, None)
+
+        _init_.assert_called_once_with(ANY, _base_weighted_counts_, None, True, True)
+        assert intersections.tolist() == [[1, 2], [3, 4]]
```
scanny, line 539: Why does `ZscoreSubtotals` need a `.intersections()` interface method? I'm not finding any callers.

It does however have a `.blocks()` interface method and I don't see a test for that here, maybe just convert this one into that and remove the intersections classmethod.

tests/unit/matrix/test_subtotals.py
```diff
@@ -580,98 +588,75 @@ def it_can_compute_a_subtotal_intersection_with_subtrahends(
...
-        subtotals = ZscoreSubtotals(None, None, cube_result_matrix_)
-        assert subtotals._table_margin == 42
-
     # --- fixture components -----------------------------------------
 
     @pytest.fixture
-    def _base_counts_prop_(self, request):
-        return property_mock(request, ZscoreSubtotals, "_base_counts")
+    def _base_weighted_counts_(self, request):
+        return instance_mock(request, _BaseWeightedCubeCounts)
```
scanny, line 654: just `weighted_counts_` here I think. `_BaseWeightedCubeCounts` is the right class to use, but the name need not reflect that because in operation this value can (and will) be any of its subclasses.

Btw, an instance mock will in general never start with an underscore. That's just for method and property mocks on helper methods (that themselves start with an underscore).

tests/unit/matrix/test_subtotals.py
```diff
@@ -580,98 +588,75 @@ def it_can_compute_a_subtotal_intersection_with_subtrahends(
             subtrahend_idxs=np.array(col_subtrahends),
         )
 
-        subtotals = ZscoreSubtotals(None, None, None)
+        subtotals = ZscoreSubtotals(_base_weighted_counts_, None)
```
scanny, line 591: just `weighted_counts_`, see below comment on fixture. same throughout this test class.

tests/unit/matrix/test_subtotals.py
```diff
@@ -580,98 +588,75 @@ def it_can_compute_a_subtotal_intersection_with_subtrahends(
...
-        subtotals = ZscoreSubtotals(None, None, cube_result_matrix_)
+        _base_weighted_counts_.weighted_counts = np.arange(12).reshape(3, 4)
+        _base_weighted_counts_.columns_margin = np.array([12, 15, 18, 21])
+        _base_weighted_counts_.table_margin = 66
+        subtotals = ZscoreSubtotals(_base_weighted_counts_, None)
 
         np.testing.assert_almost_equal(
             subtotals._subtotal_row(subtotal_),
             np.array([-0.7044435, -0.2161134, 0.2033553, 0.5833346]),
         )
```
scanny, line 636: `assert subtotals._subtotal_row(subtotal_) == pytest.approx([-0.704435, ...])`

tests/unit/matrix/test_subtotals.py
```diff
@@ -580,98 +588,75 @@ def it_can_compute_a_subtotal_intersection_with_subtrahends(
...
     ):
         subtotal_.subtrahend_idxs = np.array([0])
         property_mock(request, ZscoreSubtotals, "_ncols", return_value=4)
-        subtotals = ZscoreSubtotals(None, None, None)
+        subtotals = ZscoreSubtotals(_base_weighted_counts_, None)
 
         np.testing.assert_almost_equal(
             subtotals._subtotal_row(subtotal_),
             np.array([np.nan, np.nan, np.nan, np.nan]),
         )
```
scanny, line 648: assert ... == pytest.approx(..., nanok=True)

scanny (approved):
A lot of cleanup items. Learn aggressively from these so I don't have to type so much next time :)

I think its fine after these so approving in advance.

tests/unit/matrix/test_cubemeasure.py
ernestoarbitrio: it is defective for a catxmr matrix
```
        if not np.all(counts.shape) or np.linalg.matrix_rank(counts) < 2:
            return np.zeros(counts.shape)
```

PR #299, opened 2021-05-04
[#178016400]: mv counts as lazyproperties in measures

ernestoarbitrio:


tests/unit/matrix/test_measure.py
```diff
@@ -1259,14 +1231,19 @@ def it_computes_its_blocks_to_help(self, request, dimensions_):
...
 
 class Describe_WeightedCounts(object):
     """Unit test suite for `cr.cube.matrix.measure._WeightedCounts` object."""
 
-    def it_computes_its_blocks_to_help(self, request, dimensions_):
+    def it_computes_its_blocks_to_help(self, request, dimensions_, cube_measures_):
         # --- these need to be in list form because the assert-called-with mechanism
         # --- uses equality, which doesn't work on numpy arrays. Normally this would be
         # --- the array itself.
+        cube_measures_.weighted_cube_valid_counts = None
```
scanny, line 1246: same here.

scanny (changes requested):
Ok, largely quite good. Definitely the right approach. Still some big open questions in my mind on how many other properties the valid-counts cubemeasure objects need to support. Some work to be done on unit-test coverage and a few small tweaks.

scanny (approved):
Okay, this is good enough to move us forward, but we really need to resolve the question of whether valid-counts is in every way interchangeable with counts except that they come from a different source. It would be a really bad idea to have two cube-measure objects that had identical code and just got their `.counts` parameter from a different place.

One other option is for `_WeightedCounts` to ask the `cube` object for valid counts and use them instead of counts when applicable, as long as they had exactly the same shape as `cube.counts`. That could be handled simply in `_BaseWeightedCubeCounts.factory()`.

The one big question in my mind is whether `.valid_counts` also includes missing counts. Those are required to compute `unconditional_row_margin` for `.column_index` purposes.

We should discuss this with Mike at standup tomorrow before cutting a release. I've rebased my branch onto this one and its tests are passing again now, so whatever release we make should include both of those branches (like mine, which includes yours). I'll click my branch for you to review in a minute here.

scanny (approved):
LGTM :)

PR #300, opened 2021-05-04
Measure consolidation

gergness:
I think scale medians are looking good. Also have scale mean and scale mean standard deviation. I'm thinking it makes sense to merge these in together as an intermediate point. I can't yet do scale mean standard error because it depends on the margin, so this is all of the easy scale count measures.

What's needed so that we can sort by these marginal measures by value? Do we need a `_SortRowsByMarginHelper`?

### DIFF
--- a/src/cr/cube/cubepart.py
+++ b/src/cr/cube/cubepart.py
@@ -465,11 +465,7 @@ class _Slice(CubePartition):
 
         This value is `None` if no row element has been assigned a numeric value.
         """
-        return (
-            np.sqrt(self._columns_scale_mean_variance)
-            if self._rows_have_numeric_value
-            else None
-        )
+        return self._assembler.columns_scale_mean_stddev
 
     @lazyproperty
     def columns_scale_mean_stderr(self):
@@ -1025,11 +1021,7 @@ class _Slice(CubePartition):
 
         This value is `None` if no column elements have an assigned numeric value.
         """
-        return (
-            np.sqrt(self._rows_scale_mean_variance)
-            if self._columns_have_numeric_value
-            else None
-        )
+        return self._assembler.rows_scale_mean_stddev
 
     @lazyproperty
     def rows_scale_mean_stderr(self):
--- a/src/cr/cube/cubepart.py
+++ b/src/cr/cube/cubepart.py
@@ -393,16 +393,7 @@ class _Slice(CubePartition):
 
         This value is `None` if no row element has an assigned numeric value.
         """
-        if not self._rows_have_numeric_value:
-            return None
-
-        inner = np.nansum(
-            self._rows_dimension_numeric_values[:, None] * self.column_proportions,
-            axis=0,
-        )
-        not_a_nan_mask = ~np.isnan(self._rows_dimension_numeric_values)
-        denominator = np.sum(self.column_proportions[not_a_nan_mask, :], axis=0)
-        return inner / denominator
+        return self._assembler.columns_scale_mean
 
     @lazyproperty
     def columns_scale_mean_margin(self):
@@ -993,14 +984,7 @@ class _Slice(CubePartition):
 
         This value is `None` if no column element has an assigned numeric value.
         """
-        if not self._columns_have_numeric_value:
-            return None
-
-        column_numeric_values = self._columns_dimension_numeric_values
-        inner = np.nansum(column_numeric_values * self.row_proportions, axis=1)
-        not_a_nan_index = ~np.isnan(column_numeric_values)
-        denominator = np.sum(self.row_proportions[:, not_a_nan_index], axis=1)
-        return inner / denominator
+        return self._assembler.rows_scale_mean
 
     @lazyproperty
     def rows_scale_mean_margin(self):
--- a/src/cr/cube/cubepart.py
+++ b/src/cr/cube/cubepart.py
@@ -1087,21 +1087,7 @@ class _Slice(CubePartition):
 
         This value is `None` if no column element has an assigned numeric value.
         """
-        if not self._columns_have_numeric_value:
-            return None
-
-        not_a_nan_index = ~np.isnan(self._columns_dimension_numeric_values)
-        numeric_values = self._columns_dimension_numeric_values[not_a_nan_index]
-        counts = np.nan_to_num(
-            self._assembler.row_comparable_counts[:, not_a_nan_index]
-        ).astype("int64")
-        scale_median = np.array(
-            [
-                self._median(np.repeat(numeric_values, counts[i, :]))
-                for i in range(counts.shape[0])
-            ]
-        )
-        return scale_median
+        return self._assembler.rows_scale_median
 
     @lazyproperty
     def rows_scale_median_margin(self):
--- a/src/cr/cube/cubepart.py
+++ b/src/cr/cube/cubepart.py
@@ -1372,25 +1372,6 @@ class _Slice(CubePartition):
         return not np.all(np.isnan(self._rows_dimension_numeric_values))
 
     @lazyproperty
-    def _rows_scale_mean_variance(self):
-        """Optional 1D np.float64 ndarray of scale-mean variance for each row."""
-        if not self._columns_have_numeric_value:
-            return None
-
-        # --- Note: the variance for scale is defined as sum((Yi−Y~)2/(N)), where Y~ is
-        # --- the mean of the data.
-        not_a_nan_index = ~np.isnan(self._columns_dimension_numeric_values)
-        col_dim_numeric = self._columns_dimension_numeric_values[not_a_nan_index]
-
-        numerator = self.counts[:, not_a_nan_index] * pow(
-            np.broadcast_to(col_dim_numeric, self.counts[:, not_a_nan_index].shape)
-            - self.rows_scale_mean.reshape(-1, 1),
-            2,
-        )
-        denominator = np.sum(self.counts[:, not_a_nan_index], axis=1)
-        return np.nansum(numerator, axis=1) / denominator
-
-    @lazyproperty
     def _table_proportion_variance(self):
         """2D ndarray of np.float64 table-proportion variance for each matrix cell."""
         p = self.table_proportions
--- a/src/cr/cube/cubepart.py
+++ b/src/cr/cube/cubepart.py
@@ -506,21 +506,7 @@ class _Slice(CubePartition):
 
         This value is `None` if no row element has been assigned a numeric value.
         """
-        if not self._rows_have_numeric_value:
-            return None
-
-        not_a_nan_index = ~np.isnan(self._rows_dimension_numeric_values)
-        numeric_values = self._rows_dimension_numeric_values[not_a_nan_index]
-        counts = np.nan_to_num(
-            self._assembler.column_comparable_counts[not_a_nan_index, :]
-        ).astype("int64")
-        scale_median = np.array(
-            [
-                self._median(np.repeat(numeric_values, counts[:, i]))
-                for i in range(counts.shape[1])
-            ]
-        )
-        return scale_median
+        return self._assembler.columns_scale_median
 
     @lazyproperty
     def columns_scale_median_margin(self):
--- a/src/cr/cube/cubepart.py
+++ b/src/cr/cube/cubepart.py
@@ -1371,9 +1371,6 @@ class _Slice(CubePartition):
             )
         )
 
-    def _median(self, values):
-        return np.median(values) if values.size != 0 else np.nan
-
     @lazyproperty
     def _row_variance(self):
         """2D np.float64 ndarray of row-percentage variance for each cell."""
--- a/src/cr/cube/dimension.py
+++ b/src/cr/cube/dimension.py
@@ -873,13 +873,13 @@ class _OrderSpec(object):
 
     @lazyproperty
     def insertion_id(self):
-        """int insertion-id in the "insertion_id" field of the transform dict.
+        """int insertion-id in the "id" field of the transform dict.
 
-        Raises KeyError if this transform dict does not contain an "insertion_id" field.
-        Note that not all order types use the "insertion_id": field but this field is
-        required in all that do.
+        Raises KeyError if this transform dict does not contain an "id" field. Note that
+        not all order types use the "id": field but this field is required in all that
+        do.
         """
-        return self._order_dict["insertion_id"]
+        return self._order_dict["id"]
 
     @lazyproperty
     def measure(self):
@@ -1073,7 +1073,7 @@ class _Subtotal(object):
     @lazyproperty
     def insertion_id(self):
         """int unique identifier of this subtotal within this dimension's insertions."""
-        return self._subtotal_dict.get("insertion_id", self._fallback_insertion_id)
+        return self._subtotal_dict.get("id", self._fallback_insertion_id)
 
     @lazyproperty
     def label(self):
--- a/src/cr/cube/dimension.py
+++ b/src/cr/cube/dimension.py
@@ -880,6 +880,16 @@ class _OrderSpec(object):
         return self._order_dict["insertion_id"]
 
     @lazyproperty
+    def marginal(self):
+        """Member of enums.MARGINAL corresponding to "marginal": field in order transform.
+
+        Raises KeyError if the order dict has no "marginal": field and ValueError if the
+        value in that field is not a recognized marginal keyword. Note that not all order
+        types use the "marginal": field.
+        """
+        raise NotImplementedError()
+
+    @lazyproperty
     def measure(self):
         """Member of enums.MEASURE corresponding to "measure": field in order transform.
 
--- a/src/cr/cube/dimension.py
+++ b/src/cr/cube/dimension.py
@@ -873,13 +873,11 @@ class _OrderSpec(object):
 
     @lazyproperty
     def insertion_id(self):
-        """int insertion-id in the "id" field of the transform dict.
+        """int insertion-id in the "insertion_id" field of the transform dict.
 
-        Raises KeyError if this transform dict does not contain an "id" field. Note that
-        not all order types use the "id": field but this field is required in all that
-        do.
+        Raises KeyError if this transform dict does not contain an "insertion_id" field.
         """
-        return self._order_dict["id"]
+        return self._order_dict["insertion_id"]
 
     @lazyproperty
     def measure(self):
--- a/src/cr/cube/dimension.py
+++ b/src/cr/cube/dimension.py
@@ -11,7 +11,12 @@ else:
 
 import numpy as np
 
-from cr.cube.enums import COLLATION_METHOD as CM, DIMENSION_TYPE as DT, MEASURE
+from cr.cube.enums import (
+    COLLATION_METHOD as CM,
+    DIMENSION_TYPE as DT,
+    MARGINAL,
+    MEASURE,
+)
 from cr.cube.util import lazyproperty
 
 
@@ -887,7 +892,16 @@ class _OrderSpec(object):
         value in that field is not a recognized marginal keyword. Note that not all order
         types use the "marginal": field.
         """
-        raise NotImplementedError()
+        return MARGINAL(self.marginal_keyname)
+
+    @lazyproperty
+    def marginal_keyname(self):
+        """str value of "marginal": field in order transform.
+
+        Raises KeyError if the order dict has no "marginal": field. Note that not all
+        order types use the "measure": field, but it is a required field in all that do.
+        """
+        return self._order_dict["marginal"]
 
     @lazyproperty
     def measure(self):
--- a/src/cr/cube/matrix/assembler.py
+++ b/src/cr/cube/matrix/assembler.py
@@ -419,6 +419,11 @@ class Assembler(object):
         )
 
     @lazyproperty
+    def rows_scale_median(self):
+        """Optional 1D Row scale median marginal for this cube-result"""
+        return self._assemble_marginal(self._measures.rows_scale_median)
+
+    @lazyproperty
     def sums(self):
         """2D optional np.float64 ndarray of sum for each cell.
 
@@ -605,6 +610,22 @@ class Assembler(object):
         value is.
         """
         return self._assemble_matrix(self._measures.zscores.blocks)
+
+    def _assemble_marginal(self, marginal):
+        """Optional 1D ndarray created from a marginal.
+
+        The assembled marginal is the shape of either a row or column (determined by
+        `marginal.direction`), and with the ordering that's applied to those
+        dimensions.
+
+        It is None when the marginal is not defined (`marginal._is_defined`).
+        """
+        if not marginal.is_defined:
+            return None
+
+        order = self._column_order if marginal.direction == "row" else self._row_order
+
+        return np.hstack(marginal.blocks)[order]
 
     def _assemble_matrix(self, blocks):
         """Return 2D ndarray matrix assembled from `blocks`.
--- a/src/cr/cube/matrix/assembler.py
+++ b/src/cr/cube/matrix/assembler.py
@@ -1098,7 +1098,11 @@ class _SortRowsByMarginalHelper(_RowOrderHelper):
 
     @lazyproperty
     def _element_values(self):
-        raise ValueError("_element_values not implemented yet ")
+        return self._marginal.blocks[0]
+
+    @lazyproperty
+    def _marginal(self):
+        raise ValueError("Not yet implemented")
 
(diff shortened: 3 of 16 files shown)


src/cr/cube/cubepart.py
```diff
@@ -1090,18 +1090,7 @@ def rows_scale_median(self):
         if not self._columns_have_numeric_value:
             return None
```
scanny, line 1091: The `._columns_have_numeric_value` property should probably move down, at least as far as `Assembler` but I think probably down to the `ScaledCounts` class because that's who would naturally need to know what the numeric values were and be able to easily test whether there were any.

src/cr/cube/matrix/assembler.py
```diff
@@ -413,9 +425,23 @@ def rows_margin_proportion(self):
...
+    @lazyproperty
+    def rows_scale_median(self):
+        """Optional 1D Row scale median margin for this cube-result"""
+        if not self._measures.scaled_counts.column_has_numeric_values:
+            return None
+
+        return self._assemble_vector(
+            self._measures.scaled_counts.rows_scale_median,
             self._row_order,
         )
```
scanny, line 446: Okay, I think maybe you're right about where to do this test, that is here in `Assembler` rather than further down. So the interaction model would be like this:

```python
>>> scaled_counts = self._measures.scaled_counts
>>> scaled_counts.columns_have_numeric_values
False
>>> scaled_counts.rows_scale_median
ValueError("`.rows_scale_median` is undefined on a matrix with no column element numeric values")
```

Btw, I think it should be `.columns_have_...` instead of `.column_has_...`.

So here in assembler we check before accessing and thereby avoid an exception. In sort by value we catch the exception and return payload order instead. That seems right to me.

scanny: One other thing on this, the return value needs to say something like `.rows_scale_median_blocks` rather than just `.rows_scale_median`. The latter sounds like either an object or an assembled vector or whatever; I think for readability's sake we need to make it explicit this is a collection (pair in this case). "blocks" comes up a lot elsewhere so should be a familiar term to a reader.

src/cr/cube/matrix/assembler.py
```diff
@@ -640,28 +666,13 @@ def _assemble_matrix(self, blocks):
...
-        vector_subtotals = np.array(
-            [
-                np.nan
-                if diffs_nan and len(subtotal.subtrahend_idxs) > 0
-                else np.sum(base_vector[subtotal.addend_idxs])
-                - np.sum(base_vector[subtotal.subtrahend_idxs])
-                for subtotal in subtotals
-            ]
-        )
-        return np.hstack([base_vector, vector_subtotals])[order]
```
scanny, line 664: Good to see this in red ink :)

src/cr/cube/matrix/measure.py
```diff
@@ -8,7 +8,11 @@
 from scipy.stats import t
 
 from cr.cube.matrix.cubemeasure import CubeMeasures
-from cr.cube.matrix.subtotals import SumSubtotals, NanSubtotals, NoneSubtotals
+from cr.cube.matrix.subtotals import (
+    SumSubtotals,
+    NanSubtotals,
+    NoneSubtotals,
+)
```
scanny, line 15: This will fail Black on test. If you remove the last comma (after `NoneSubtotals`) it should reformat back to a single line. That one-way change is a gotcha in Black on Python 3. It puts it in if its missing, but doesn't take it out if it will fit on one line again once you make a deletion.

src/cr/cube/matrix/measure.py
```diff
@@ -982,6 +991,63 @@ def _subtotal_rows(self):
         )
 
 
+class _ScaledCounts(_BaseSecondOrderMeasure):
+    """Provider of scaled count based marginals"""
```
scanny, line 995: Well, I have to say I am starting to come around to the idea of having marginals each having a distinct class, like measures do. These could share a common base class if they needed to share computations, like retrieving numeric values or something. That way you can have some diversity in the interface, like you could ask for the base-values separately from the subtotals if you needed to, which some other measures might and perhaps already do.

Also it would make calling them prettier, like:

```python
assemble_column_vector(measures.rows_margin.blocks)
```

or

```python
rows_scale_median = self._measures.rows_scale_median
return (
    None
    if rows_scale_median.is_not_computable_or_whatever
    else assemble_column_vector(rows_scale_median.blocks)
)
```

Not sure this whether this is the right moment, but as I'm seeing it shape up I'm starting to feel the benefit that would bring.

Maybe `.is_not_computable_or_whatever` could be `.has_blocks` or maybe `.is_defined` and that would be applicable to other margins as well without having to know the details of why it might undefined in this particular case.

scanny (commented):
Okay, here are my initial reactions, should be some food for thought :)

gergness:
@scanny I have a backup of the old way locally, but I just force pushed another spike.

I ended up liking making the marginal class know it's direction and then adding a `assemble_marginal` function.

src/cr/cube/matrix/measure.py
```diff
@@ -1320,6 +1325,118 @@ def blocks(self):
...
+        )
+
+    @lazyproperty
+    def direction(self):
+        """String, either "row" or "column", the directionality the marginal."""
+        raise NotImplementedError(  # pragma: no cover
+            "%s must implement `.direction`" % type(self).__name__
+        )
+
+    @lazyproperty
```
gergness, line 1356: It occurs to me that this (and another one below) don't really need to be lazyproperties, they could just be properties. Should I change?

src/cr/cube/matrix/measure.py
```diff
@@ -1320,6 +1325,118 @@ def blocks(self):
...
+    """
+
+    @lazyproperty
+    def blocks(self):
+        """List of the 2 1D ndarray "blocks" of the rows-scale-median.
+
+        These are the base-values and the subtotals.
+        """
+        if not self.is_defined:
+            raise ValueError("No numeric values are defined on the columns dimension.")
```
scanny, line 1417: I'm thinking we should probably stick to speaking in caller-space for an exception like this one, meaning something like:

"rows-scale-median is undefined when no numeric values are defined on columns dimension".

In this case, no caller on `_Slice` is going to see this today, but still, might as well spell out as much as possible, like "this exception was raised because it is undefined for this value (cube-result)" and "the reason this cube-result is insufficient is that it has no numeric values on the columns dimension". This gives the caller all they need to know to fix the problem.

scanny (commented):
Okay, this is shaping right up. I think you're pretty much there on the median and the rest should fall into place with the pattern you've established.

Just a couple fine points in comments.

src/cr/cube/matrix/assembler.py
```diff
@@ -624,6 +629,22 @@ def zscores(self):
...
+        dimensions.
+
+        It is None when the marginal is not defined (`marginal._is_defined`).
+        """
+        if not marginal.is_defined:
+            return None
+
+        order = self._column_order if marginal.direction == "row" else self._row_order
+
+        return np.hstack(marginal.blocks)[order]
```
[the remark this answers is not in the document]
scanny: Okay, one thing that occurred to me after reflection. If we do it this `._assemble_marginal()` way, we will still need `._assemble_vector()` or its equivalent because the row and column label vectors are not a marginal and won't need anything outside of `Assembler` to determine their values (they come straight from `Dimension`). So I suppose I'm back to favoring `._assemble_column_vector` and `._assemble_row_vector`, although there could be other options. Why don't you have a noodle on it and we can discuss on the next pass or whatever.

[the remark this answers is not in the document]
gergness: I added `_assemble_marginal` and haven't refactored the ones that currently use `_assemble_vector`. But right now, the labels actually already have their own function (so we currently have 3).

- `_assemble_vector` - Used by `rows/columns_margin`, & `rows/columns_margin_proportion`
- `_dimension_labels` - Used by `row/column_labels` (and does the assembling)
- `_assemble_marginal` - My new method

In later refactoring, I think there should be two `_assemble_marginal` and `_assemble_vector`, and internally `_assemble_marginal` should call `_assemble_vector`, it just uses the marginal's information to pass the right arguments.

src/cr/cube/matrix/measure.py
```diff
@@ -1326,6 +1336,178 @@ def blocks(self):
...
+        These are the base-values and the subtotals.
+        """
+        raise NotImplementedError(  # pragma: no cover
+            "%s must implement `.blocks`" % type(self).__name__
+        )
+
+    @lazyproperty
+    def orientation(self):
+        """String, either "row" or "column", the orientation the marginal."""
+        return self._orientation
```
scanny, line 1369: Let's make this an Enum. I'm thinking `cr.cube.enums.ORIENTATION`. I bet we end up using this more than we initially expected.

src/cr/cube/matrix/measure.py
```diff
@@ -1326,6 +1336,178 @@ def blocks(self):
...
+
+    @lazyproperty
+    def is_defined(self):
+        """True if any have numeric values"""
+        return not np.all(np.isnan(self._opposing_numeric_values))
+
+    @lazyproperty
+    def _sorted_counts(self):
+        axis = 1 if self.orientation == "rows" else 0
```
scanny, line 1451: This cluster all needs docstrings.

scanny (commented):
Okay, a few small comments. This looks solid to me, I'd say it's ready for replication to the other scaled-count marginals.

src/cr/cube/matrix/measure.py
```diff
@@ -1326,6 +1356,351 @@ def blocks(self):
         return ZscoreSubtotals.blocks(self._weighted_cube_counts, self._dimensions)
 
 
+# === MARGINALS ===
+
+
+class _BaseSecondOrderMarginal(object):
+    """Base class for all second-order measure objects."""
```
scanny, line 1363: I'm not sure there is such thing as a "first-order" marginal, is there? Wouldn't `_BaseMarginal` say all that needs to be said?

Docstring is unedited copy from `_BaseSecondOrderMeasure`.

src/cr/cube/matrix/measure.py
```diff
@@ -1326,6 +1356,351 @@ def blocks(self):
...
+
+    def __init__(self, dimensions, second_order_measures, cube_measures, orientation):
+        self._dimensions = dimensions
+        self._second_order_measures = second_order_measures
+        self._cube_measures = cube_measures
+        self._orientation = orientation
```
scanny, line 1369: After living with it for a while I found `second_order_measures` to be unnecessarily long-winded and in `_Strand` I changed it to just `measures` with good results. Maybe since this is new we do the same here and then we can get around to changing it elsewhere sometime later.

tests/unit/matrix/test_measure.py
```diff
@@ -1371,3 +1413,491 @@ def but_the_subtotal_blocks_are_NaNs_when_an_MR_dimension_is_present(
...
+    @pytest.mark.parametrize(
+        "orientation, array, expected",
+        (
+            (OR.ROWS, np.array([[0, 1, 2], [3, 4, 5]]), [3, 12]),
+            (OR.COLUMNS, np.array([[0, 1, 2], [3, 4, 5]]), [3, 5, 7]),
+            (OR.ROWS, np.array([]).reshape(0, 6), []),
+            (OR.COLUMNS, np.array([]).reshape(6, 0), []),
+        ),
+    )
+    def it_can_apply_along_orientation(self, orientation, array, expected):
```
scanny, line 1444: A `_to_help` suffix wouldn't be out of place here, to indicate this is testing a helper and not an interface method.

scanny (approved):
Okay, this looks real good to me. I don't think I need to look at it again. We can do a quick cafe for any questions. Plenty of comments but they're all fine points. Approving in advance.

src/cr/cube/dimension.py
```diff
@@ -873,13 +873,13 @@ def element_ids(self):
...
     def insertion_id(self):
-        """int insertion-id in the "insertion_id" field of the transform dict.
+        """int insertion-id in the "id" field of the transform dict.
 
-        Raises KeyError if this transform dict does not contain an "insertion_id" field.
-        Note that not all order types use the "insertion_id": field but this field is
-        required in all that do.
+        Raises KeyError if this transform dict does not contain an "id" field. Note that
+        not all order types use the "id": field but this field is required in all that
+        do.
```
scanny, line 880: The `Note that not all ...` sentence seems like a cut-and-paste error from when I created that in the sort-by-value PR. I'm thinking that doesn't really fit in this context, do you? Maybe we should cut that sentence.

tests/unit/test_dimension.py
```diff
@@ -1208,14 +1208,14 @@ def it_knows_the_ordered_element_ids_for_an_explicit_sort(
         assert _OrderSpec(None, None).element_ids == expected_value
 
     def it_knows_the_sort_vector_insertion_id(self, _order_dict_prop_):
-        _order_dict_prop_.return_value = {"insertion_id": 42}
+        _order_dict_prop_.return_value = {"id": 42}
```
scanny, line 1211: Ah, no, I think I see now. So the _insertion_ gets `"id":`, but the _order_ still gets `"insertion_id":`. So the id of the insertion itself appears in `transforms.rows_dimension.insertions[i].id` but the _reference_ to it in the order dict appears at `order.insertion_id`.

tests/unit/test_dimension.py
```diff
@@ -1496,7 +1496,7 @@ def it_provides_access_to_the_addend_element_ids(
         assert subtotal.addend_ids == expected_value
 
     def it_knows_its_insertion_id_when_it_has_been_assigned_one(self):
-        assert _Subtotal({"insertion_id": 42}, None, None).insertion_id == 42
+        assert _Subtotal({"id": 42}, None, None).insertion_id == 42
```
scanny, line 1499: so this one stays. The two in `Describe_OrderSpec` go back to `"insertion_id":`.

src/cr/cube/cubepart.py
```diff
@@ -1399,9 +1347,6 @@ def _dimensions(self):
             )
         )
 
-    def _median(self, values):
-        return np.median(values) if values.size != 0 else np.nan
-
```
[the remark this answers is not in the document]
gergness: all except `_rows_scale_mean_variance` still are being used (`_columns_scale_mean_variance` is used for pairwise significance for now)

src/cr/cube/matrix/assembler.py
```diff
@@ -1096,4 +1096,20 @@ class _SortRowsByMarginalHelper(_RowOrderHelper):
     a matrix dimension has an opposing dimension.
     """
 
-    pass
+    @lazyproperty
+    def _element_values(self):
+        raise ValueError("_element_values not implemented yet ")
```
scanny: I usually use `NotImplementedError` (by itself) for fields I haven't gotten to yet. The naming seems apt :) I generally don't bother with a message since I can tell what's not implemented by the error message location and I'm going to implement it in a second anyway.

I do however make it a practice to write the docstring as soon as I add the method. When I (used to) wait I often ended up not remembering it. It's also a key opportunity to make sure I have a clear idea what this method is going to do for me (its contract). That's good timing because if I change my mind about what (or how much) it should do (which happens more than one might expect) then I can change the caller without having to revert code I've already written a test for a committed.

tests/unit/matrix/test_assembler.py
```diff
@@ -1561,11 +1562,26 @@ def it_provides_the_order(self, request):
...
+        _SortRowsByMarginalHelper(None, None)._order
 
         SortByValueCollator_.display_order.assert_called_once_with(
             _rows_dimension_(),
             _element_values_(),
             _subtotal_values_(),
             _empty_row_idxs_(),
         )
+
+    def it_provides_the_element_values_to_help(self, request, _marginal_prop_):
```
scanny, line 1579: `request` is not necessary unless it is used (easy to forget on copy-paste, I do it frequently :).

tests/unit/matrix/test_assembler.py
```diff
@@ -1561,11 +1562,26 @@ def it_provides_the_order(self, request):
...
+        assert _SortRowsByMarginalHelper(None, None)._subtotal_values == "b"
+
+    # fixture components ---------------------------------------------
+
+    @pytest.fixture
+    def _marginal_prop_(self, request):
+        marginal_ = instance_mock(request, _BaseMarginal, blocks=["a", "b"])
+        return property_mock(
+            request, _SortRowsByMarginalHelper, "_marginal", return_value=marginal_
+        )
```
scanny, line 1620: I find it wiser to not make pytest fixtures anything more than a single, simple mock (no specified return or attribute values). I used to do things like this, but discovered after a while that it made the fixtures less flexible and also makes it so you have to scroll down to the fixtures to understand the test. That latter reason is probably the more compelling one.

So here you would have both `_marginal_prop_` and `marginal_` as fixtures, inject them both in the test, and then have `_marginal_prop_.return_value = marginal_` in the test fixturing block. Also, the `marginal_.blocks = ["a", "b"]` would be in the test fixturing block and you can make full sense of the test without ever scrolling down (which is a real pain when test suites get bigger).

So a test like this is easy to make sense of. Really the name should tell you everything you need to know about a mock:

```python
def it_provides_the_element_values_to_help(self _marginal_prop_, marginal_):
    _marginal_prop_.marginal_ = marginal_
    marginal_.blocks = ["a", "b"]

    assert _SortRowsByMarginalHelper(None, None)._element_values == "a"
```

src/cr/cube/dimension.py
```diff
@@ -879,6 +879,16 @@ def insertion_id(self):
...
 
+    @lazyproperty
+    def marginal(self):
+        """Member of enums.MARGINAL corresponding to "marginal": field in order transform.
+
+        Raises KeyError if the order dict has no "marginal": field and ValueError if the
+        value in that field is not a recognized marginal keyword. Note that not all order
+        types use the "marginal": field.
+        """
+        raise NotImplementedError()
```
scanny: You don't need the parens on an exception if you don't supply a message.

src/cr/cube/matrix/assembler.py
```diff
@@ -1102,7 +1103,21 @@ def _element_values(self):
...
     def _marginal(self):
-        raise ValueError("Not yet implemented")
+        """Marginal object providing values for sort."""
+        propname_by_marginal = {
+            MARGINAL.SCALE_MEAN: "rows_scale_mean",
+            MARGINAL.SCALE_MEAN_STDDEV: "rows_scale_mean_stddev",
+            MARGINAL.SCALE_MEDIAN: "rows_scale_median",
+        }
+        marginal = self._order_spec.marginal
+        marginal_propname = propname_by_marginal.get(marginal)
```
scanny, line 1113: It's a fine point, but this can be condensed to:

```python
    marginal = self._order_spec.marginal
    marginal_propname =  {
        MARGINAL.SCALE_MEAN: "rows_scale_mean",
        MARGINAL.SCALE_MEAN_STDDEV: "rows_scale_mean_stddev",
        MARGINAL.SCALE_MEDIAN: "rows_scale_median",
    }.get(marginal)
```

That's one less temporary variable to think of a name for and the reader to keep track of.

tests/unit/matrix/test_assembler.py
```diff
@@ -1574,6 +1579,34 @@ def it_provides_the_order(self, request):
...
+        second_order_measures_ = instance_mock(request, SecondOrderMeasures)
+        setattr(second_order_measures_, marginal_prop_name, "foo")
+        _order_spec_ = instance_mock(request, _OrderSpec, marginal=marginal)
+        dimensions_ = [instance_mock(request, Dimension, order_spec=_order_spec_)]
+        order_helper = _SortRowsByMarginalHelper(dimensions_, second_order_measures_)
+
+        assert order_helper._marginal == "foo"
+
+    def but_it_raises_on_unknown_marginal(self, request):
+        second_order_measures_ = instance_mock(request, SecondOrderMeasures)
```
scanny, line 1600: If you use the same mock twice (or more) I would make it a fixture. The inlines are only for cases where there's just the one test that's ever going to use it. This particular case is one of the reasons I went from `second_order_measures` to the shorter `measures`. The longer version works okay in the code but parameter-list real estate is more precious in the unit tests it seems :)

It looks like there are at least four mocks here that could profitably be moved to the `fixture components` section below.

tests/unit/matrix/test_assembler.py
```diff
@@ -1574,6 +1579,34 @@ def it_provides_the_order(self, request):
...
+        _order_spec_ = instance_mock(request, _OrderSpec, marginal="bar")
+        dimensions_ = [instance_mock(request, Dimension, order_spec=_order_spec_)]
```
scanny, line 1602: This is an isolation gap. You're exercising `_RowOrderHelper._order_spec` unnecessarily. What you want is `_order_spec_prop_` and `_order_spec_prop_.return_value = order_spec_`. Dimensions don't come into it for this unit.

tests/unit/matrix/test_assembler.py
```diff
@@ -1574,6 +1579,34 @@ def it_provides_the_order(self, request):
...
+        order_helper = _SortRowsByMarginalHelper(dimensions_, second_order_measures_)
+
+        with pytest.raises(NotImplementedError) as e:
+            order_helper._marginal
+
+        assert str(e.value) == ("sort-by-value for marginal 'bar' is not yet supported")
```
scanny, line 1608: Looks like Black left on parens when the string got short enough to fit on one line. Not sure why it does that.

tests/unit/test_dimension.py
```diff
@@ -1217,6 +1222,26 @@ def but_it_raises_when_insertion_id_not_present(self, _order_dict_prop_):
...
 
+    def it_knows_the_marginal_specified_as_the_sort_basis(self, _order_dict_prop_):
+        _order_dict_prop_.return_value = {"marginal": "scale_mean"}
+        assert _OrderSpec(None, None).marginal == MARGINAL.SCALE_MEAN
+
+    def but_it_raises_when_marginal_field_is_not_present(self, _order_dict_prop_):
+        _order_dict_prop_.return_value = {}
+        with pytest.raises(KeyError) as e:
+            _OrderSpec(None, None).marginal
+        assert str(e.value) == "'marginal'"
```
scanny, line 1233: If we're going to bother to test an exception the exception message should be more descriptive. I'm thinking this is a case that's unlikely to arise once we're done with migration because all valid values will be implemented. I would just leave this test out.

tests/unit/test_dimension.py
```diff
@@ -1217,6 +1222,26 @@ def but_it_raises_when_insertion_id_not_present(self, _order_dict_prop_):
...
+
+    def and_it_raises_when_the_marginal_keyword_is_not_recognized(
+        self, _order_dict_prop_
+    ):
+        _order_dict_prop_.return_value = {"marginal": "foobar"}
+        with pytest.raises(ValueError) as e:
+            _OrderSpec(None, None).marginal
+        # --- `.endswith()` to accommodate a Python 2.7 vs. 3.0 error message difference
+        # --- in Enum module
+        assert str(e.value).endswith(" is not a valid MARGINAL")
```
scanny, line 1243: Probably this one too because once migration is done all the enum keywords will be defined. The exception should and will still arise if there's some development-time error, but I don't think we need to verify that it does. We're not the ones raising the exception. Come to think of it, I think that's the rule-of-thumb; only test exceptions that you yourself raise in the code.

src/cr/cube/matrix/assembler.py
```diff
@@ -1032,3 +1087,48 @@ def _subtotal_values(self):
...
+class _SortRowsByMarginalHelper(_RowOrderHelper):
+    """Orders elements by the values of an opposing marginal vector.
+
+    This would be like "order rows in descending order by value of rows scale mean.
+    column. An opposing-marginal ordering is only available on a matrix, because only
+    a matrix dimension has an opposing dimension.
+    """
+
+    @lazyproperty
+    def _element_values(self):
```
scanny, line 1101: docstring

src/cr/cube/matrix/assembler.py
```diff
@@ -1032,3 +1087,48 @@ def _subtotal_values(self):
...
+        """tuple of int element-idx specifying ordering of dimension elements."""
+        return SortByValueCollator.display_order(
+            self._rows_dimension,
+            self._element_values,
+            self._subtotal_values,
+            self._empty_row_idxs,
+        )
+
+    @lazyproperty
+    def _subtotal_values(self):
```
scanny, line 1133: docstring

scanny (approved):
Okay, LGTM with these changes :)

PR #301, opened 2021-05-04
[#177992894]: spike - consolidate col index

gergness:
WIP


scanny (commented):
I think my main initial question here is whether we need a whole new family of cubemeasure objects for this or whether these could just be additional properties on `_*WeightedCubeCounts`. Basically my reasoning would be that these objects derive from the same cube-measure (cube-counts) and so aren't really fundamentally different.

gergness:
got a new ticket number so will be re-making this PR

PR #303, opened 2021-05-07
[#178045298]: Fix pop counts

slobodan-ilic:
* Add test cases for cat array and pop counts
* Fix pop counts to not be divided between subvariables

### DIFF
--- a/tests/integration/test_cubepart.py
+++ b/tests/integration/test_cubepart.py
@@ -1013,7 +1013,7 @@ class Describe_Slice(object):
             abs=10e-2,
         )
 
-    def it_uses_correct_proportion_for_cat_array_subvariables(self):
+    def it_uses_rows_proportions_for_ca_subvar_x_ca_cat_population_counts(self):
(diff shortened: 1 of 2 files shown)


tests/integration/test_cubepart.py
```diff
@@ -1013,6 +1013,47 @@ def it_uses_column_proportions_for_pop_counts_when_column_dim_is_cat_date(self):
...
+                    [0.0, 33.33333333, 16.66666667, 50.0],
+                ]
+            )
+        )
+        # ---Assert population is equal to total for each subvariable
+        assert slice_.population_counts.sum(axis=1) == pytest.approx(
+            np.array([100, 100, 100])
+        )
+
+    def it_uses_correct_proportion_for_pets_array(self):
```
ernestoarbitrio, line 1032: i would use the name of the test with the same structure of the fixture. So I would say:
`_for_mr_x_ca_cat_x_ca_subvar` and not `pets_array`

ernestoarbitrio (approved):

PR #305, opened 2021-05-19
Margin consolidation 178165164

gergness:
Consolidate rows_margin and columns_margin while being minimally destructive. I still think we could save ourselves a lot of confusion by more refactoring about what uses bases vs margins, but that ended up being a really big task.

It does require removing some tests from exporter, these were ones that calculated the percent moe on cat arrays. The denominator of percent moe is the margin across the subvariable dimension, so they are not valid.

https://github.com/Crunch-io/zoom/pull/3910


src/cr/cube/matrix/measure.py
```diff
@@ -1856,6 +1902,29 @@ def _stddev_func(self):
...
+
+    @lazyproperty
+    def is_defined(self):
+        """True if counts are defined."""
+        try:
+            self._counts
+            return True
+        except ValueError:
+            return False
+
```
ernestoarbitrio, line 1926: i don't see any usage of this property, what's used for?

ernestoarbitrio (changes requested):
Per comments. I can take another look when you address these comments

src/cr/cube/matrix/assembler.py
```diff
@@ -151,22 +151,13 @@ def columns_dimension_numeric_values(self):
     def columns_margin(self):
         """1D/2D np.float64 ndarray of weighted-N for each column of this slice."""
         # --- an MR_X slice produces a 2D columns-margin (each cell has its own N) ---
+        # --- This is really just another way to call the columns_weighted_bases ---
+        # --- TODO: Should this error for 2D margins? ---
```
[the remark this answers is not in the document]
gergness: We are considering making `columns_margin` always be 1D, and error on arrays. Then having a `columns_margin_range` that would be used for array variables. This would require more changes to exporter though

src/cr/cube/matrix/cubemeasure.py
```diff
@@ -914,14 +918,31 @@ def factory(cls, cube, dimensions, slice_idx):
...
         return np.broadcast_to(self.columns_margin, self.weighted_counts.shape)
 
+    @lazyproperty
+    def column_comparable_counts(self):
+        """2D np.float64 ndarray of weighted-count for cells in columns that share a base
+
+        Cells in a column don't share a base when they go across array subvariables.
+        Raises when counts are not comparable.
+        """
+        if self._dimensions_are_array[1]:
```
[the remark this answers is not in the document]
gergness: oh nice. The negative indexing still feels weird to me (maybe just because it's different between python and R), so it felt better to add  `_row/column_is_array` property.

src/cr/cube/matrix/measure.py
gergness: It's used in `_assemble_marginal`, to make the marginals optional. I'm not sure, maybe it should just propagate the error instead?

gergness: I can see further down the line making a `.rows_margin_available` property exposed in `_Slice` though

src/cr/cube/matrix/cubemeasure.py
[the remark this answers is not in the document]
gergness: oh nevermind, I was confused about the difference between the dimensions on `_Cube` and on `_Slice`. So my code is like you suggested.

src/cr/cube/matrix/cubemeasure.py
```diff
@@ -1042,6 +1036,24 @@ def _array_type_std_res(self, counts, total, rowsum, colsum):
...
+
+    @lazyproperty
+    def _row_is_array(self):
+        """Bool indicating whether the row is an array variable type
+
+        `_dimensions` is already apparent dimensions, so the rows is the 2nd to last
+        dimension. (order is rows, column if 2D or tabs, rows, columns if 3D)
+        """
+        return self._dimensions[-2].dimension_type in DT.ARRAY_TYPES
+
```
ernestoarbitrio: I think you don't need 2 properties considering that these guards are used only 1 time, so you can use directly
`self._dimensions[-1].dimension_type in DT.ARRAY_TYPES` as condition in the row and col comparable counts.

src/cr/cube/matrix/assembler.py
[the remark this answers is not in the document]
ernestoarbitrio: Can you write this more meaningful explanation as the TODO ? If we read at this comment in 2 months we propably have forgotten the rationale :D

src/cr/cube/matrix/measure.py
ernestoarbitrio: we don't have any integration test that steps in the except branch. This guard seems a little bit fishy to me. Also, when the self._counts raise a `ValueError` ? 
The _counts are defined as:
```
    @lazyproperty
    def _counts(self):
        """List of 2 ndarrays count blocks, in their original payload order

        The counts come from the count measure blocks that are comparable in the correct
        orientation, and include the base values and the subtotals for the orientation.
        """
        if self.orientation == MO.ROWS:
            # --- Use *column* comparable counts (going across rows) ---
            counts = self._second_order_measures.column_comparable_counts.blocks
            # --- Get base values & *row* subtotals
            return [counts[0][0], counts[1][0]]
        else:
            # --- Use *row* comparable counts (going across columns) ---
            counts = self._second_order_measures.row_comparable_counts.blocks
            # --- Get base values & *column* subtotals
            return [counts[0][0], counts[0][1]]
 ```
 That seems they will be always present, so I can't buy the `is_defined` written this way. We need to understand when, in this case, the `is_defined` is False.

Comment (ernestoarbitrio) [reply src/cr/cube/matrix/measure.py:1925]: I see you wrote a unit test for this property but we don't have any integration test that steps in the except branch, and I think an integration tests that exercise this conditional branch is mandatory.

Also, when the self._counts raise a `ValueError` ? 
The _counts are defined as:
```
    @lazyproperty
    def _counts(self):
        """List of 2 ndarrays count blocks, in their original payload order

        The counts come from the count measure blocks that are comparable in the correct
        orientation, and include the base values and the subtotals for the orientation.
        """
        if self.orientation == MO.ROWS:
            # --- Use *column* comparable counts (going across rows) ---
            counts = self._second_order_measures.column_comparable_counts.blocks
            # --- Get base values & *row* subtotals
            return [counts[0][0], counts[1][0]]
        else:
            # --- Use *row* comparable counts (going across columns) ---
            counts = self._second_order_measures.row_comparable_counts.blocks
            # --- Get base values & *column* subtotals
            return [counts[0][0], counts[0][1]]
 ```
 
 That seems they will be always present, so I can't buy the `is_defined` written this way. We need to understand when, in this case, the `is_defined` is False.
 
 We need a real(ish) fixture that exercise the condition when `is_defined` is False, I'm imagining a guard like:
 ` self._counts[0].size > 0` or something similar, but i can't figure out when the base counts are not present.

src/cr/cube/matrix/assembler.py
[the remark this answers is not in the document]
ernestoarbitrio: Can you write this more meaningful explanation as the TODO ? If we read at this comment in 2 months we propably have forgotten the rationale :D

ernestoarbitrio (changes requested):
Per comments, the major concern is the missing tests for the `is_defined` in the SummationCounts class

src/cr/cube/matrix/measure.py
```diff
@@ -384,12 +384,25 @@ class _ColumnComparableCounts(_BaseSecondOrderMeasure):
...
         return SumSubtotals.blocks(
-            self._weighted_cube_counts.column_comparable_counts,
+            self._weighted_cube_counts.weighted_counts,
             self._dimensions,
             diff_rows_nan=True,
         )
 
+    @lazyproperty
+    def is_defined(self):
+        """Bool indicating whether counts row comparable counts are defined
```
ernestoarbitrio, line 400: i think this docstring is wrong

src/cr/cube/matrix/measure.py
```diff
@@ -985,12 +998,23 @@ class _RowComparableCounts(_BaseSecondOrderMeasure):
...
         return SumSubtotals.blocks(
-            self._weighted_cube_counts.row_comparable_counts,
+            self._weighted_cube_counts.weighted_counts,
             self._dimensions,
             diff_cols_nan=True,
         )
 
+    @lazyproperty
+    def is_defined(self):
+        """Bool indicating whether counts row comparable counts are defined
```
ernestoarbitrio, line 1012: i think the `counts` before row is a typo, isn't it?

src/cr/cube/matrix/measure.py
```diff
@@ -1642,6 +1666,17 @@ def _counts(self):
...
+    @lazyproperty
+    def _counts_are_defined(self):
+        """Bool indicating whether counts are defined for this orientation
+
+        Counts are undefined across subvariable dimensions.
+        """
+        if self.orientation == MO.ROWS:
+            return self._second_order_measures.column_comparable_counts.is_defined
+        else:
+            return self._second_order_measures.row_comparable_counts.is_defined
```
ernestoarbitrio, line 1678: the else statement can be removed:
```python
if self.orientation == MO.ROWS:
    return self._second_order_measures.column_comparable_counts.is_defined
return self._second_order_measures.row_comparable_counts.is_defined
 ```

ernestoarbitrio (changes requested):

ernestoarbitrio (approved):
LGTM! :D

PR #308, opened 2021-05-21
Consolidate table_proportion & (spike of) population_counts

gergness:


src/cr/cube/cubepart.py
```diff
@@ -707,23 +707,18 @@ def population_counts(self):
...
+        affect sort order.
         """
-        proportions = (
-            self.row_proportions
-            if self.rows_dimension_type in (DT.CAT_DATE, DT.CA_SUBVAR)
-            else self.column_proportions
-            if self.columns_dimension_type in (DT.CAT_DATE, DT.CA_SUBVAR)
-            else self.table_proportions
+        return (
+            self._assembler.cell_population_fractions
```
scanny, line 718: The naming choice `.cell_population_fractions` is a little unfortunate here. In what way is it related to `_cube.population_fraction` which shares the same root name? I think the answer is "it's really a different concept", but I'll read on.

----

Okay, you'll see below, I recommend `.population_proportions` for this name. It's still the familiar proportions concept, one of row, column, or table, just that which one is appropriate for computing population depends on a few things.

src/cr/cube/matrix/assembler.py
```diff
@@ -342,6 +342,17 @@ def pairwise_significance_means_t_stats(self, column_idx):
...
+    @lazyproperty
+    def cell_population_fractions(self):
+        proportions = (
+            self.row_proportions
+            if self._rows_dimension.dimension_type in (DT.CAT_DATE, DT.CA_SUBVAR)
+            else self.column_proportions
+            if self._columns_dimension.dimension_type in (DT.CAT_DATE, DT.CA_SUBVAR)
+            else self.table_proportions
+        )
+        return proportions
```
scanny, line 354: Okay, needs a docstring but I see what you're doing here. Maybe `.population_proportions`. I would interpret this as the fraction of the overall population represented by the base for each cell.

We've historically used the prefix `cell_` to mean the same concept as `table_` means here, like `cell_percents` probably still lives somewhere in exporter somewhere and meaning `table_` (or perhaps some places `total_`) percents. So `cell_` is probably better avoided as potentially confusing, also all proportions measures have a value for each cell, so no need to be explicit about that.

I wouldn't involve a new `proportions` name here, `return (self.row_ ... if ... else ...)` states it clearly and simply to me. The docstring can fill in the idea that the proportions basis varies and why.

scanny: Btw, this property is out of position would belong much higher up in alphabetical order. Interestingly, if you change the name to `.population_proportions` then it's in exactly the right position.

src/cr/cube/matrix/measure.py
```diff
@@ -318,6 +323,27 @@ def _weighted_cube_counts(self):
...
+
+    @lazyproperty
+    def blocks(self):
+        return (
+            self._second_order_measures.row_proportions.blocks
+            if self._dimensions[0].dimension_type in (DT.CAT_DATE, DT.CA_SUBVAR)
+            else self._second_order_measures.column_proportions.blocks
+            if self._dimensions[1].dimension_type in (DT.CAT_DATE, DT.CA_SUBVAR)
+            else self._second_order_measures.table_proportions.blocks
+        )
```
scanny, line 344: You'll need to choose between putting this logic here or up in assembler. I think the choice comes down to whether we need it for sort-by-value, which I expect Mike will want. In that case, the assembler bit would just become `self._assemble_matrix(self._measures.population_proportions.blocks)` and this would stay.

You would of course need an entry in `SecondOrderMeasures` for it then.

Docstring.

scanny (commented):
Okay, looks good, a couple of directional comments, pretty sure those will make sense to you.

ernestoarbitrio (approved):
LGTM

PR #314, opened 2021-05-28
[178331859]: avoid sentry error for sort-by-value 

ernestoarbitrio:


gergness (commented):
I can see that this handles the case when the sort-by-value measure name is not recognized, but I'm not sure it covers all the times when sort-by-value would fail. (I realize now that this maybe what you were asking me yesterday, sorry!)

Does it cover these cases?
- What about when you try to sort by `scale_mean`, but there are no numeric values (and so `rows_scale_mean` doesn't exist)?
- What about when you try to sort by `weighted_counts`, using `element_id` = 10, but there is no opposing element with id=10.
- I think it's validated, so this won't come up in practice, but it'd be good not to error for other weird things in the transform too.

gergness (changes requested):
Here are some motivating integration tests. I think the first 2 definitely need to be handled, I'm less sure that the third one matters (because the transforms are validated by zoom). I also don't think this is exhaustive, but hopefully gives you a sense of the kinds of things we want to recover from.

```
    def it_can_recover_from_sort_by_value_failure_invalid_id(self):
        """Falls back to payload order when opposing element id does not exist."""
        transforms = {
            "rows_dimension": {
                "order": {
                    "type": "opposing_element",
                    "element_id": 500,  # --- There is no element with id=500
                    "measure": "col_percent",
                    "direction": "ascending",
                }
            }
        }
        slice_ = _Slice(Cube(CR.CAT_4_X_CAT_4), 0, transforms, None, 0)

        # --- measure element id = 500 does not exist
        assert 500 not in slice_._dimensions[0].element_ids
        # --- measures should be in payload order because id was unavailable
        assert slice_.counts.tolist() == [
            [14.0, 14.0, 13.0, 16.0], [22.0, 14.0, 19.0, 19.0], 
            [14.0, 16.0, 19.0, 18.0], 
            [17.0, 12.0, 28.0, 11.0],
        ]

    def it_can_recover_from_sort_by_value_failure_no_measure(self):
        """Falls back to payload order when measure is unavailable."""
        transforms = {
            "rows_dimension": {
                "order": {
                    "type": "marginal",
                    "marginal": "scale_mean",  # --- scale_mean is unavailable
                    "direction": "ascending",
                    "fixed": {"top": [2], "bottom": [1]},
                }
            }
        }
        slice_ = _Slice(Cube(CR.CAT_4_X_CAT_4), 0, transforms, None, 0)

        # --- scale means don't exist because there are no numeric_values
        assert slice_.rows_scale_mean is None
        # --- but other measures should be in payload order
        assert slice_.counts.tolist() == [
            [14.0, 14.0, 13.0, 16.0], [22.0, 14.0, 19.0, 19.0], 
            [14.0, 16.0, 19.0, 18.0], 
            [17.0, 12.0, 28.0, 11.0],
        ]

    def it_can_recover_from_sort_by_value_invalid_sort_type(self):
        """Falls back to payload order when measure is unavailable."""
        transforms = {
            "rows_dimension": {
                "order": {
                    "type": "foo",  # --- not a valid sort order dict
                }
            }
        }
        slice_ = _Slice(Cube(CR.CAT_4_X_CAT_4), 0, transforms, None, 0)

        # --- measure should be in payload order
        assert slice_.counts.tolist() == [
            [14.0, 14.0, 13.0, 16.0], [22.0, 14.0, 19.0, 19.0], 
            [14.0, 16.0, 19.0, 18.0], 
            [17.0, 12.0, 28.0, 11.0],
        ]

```

gergness (approved):

gergness:
Oh sorry I reviews wrong pr, I’ll look at this in a bit

gergness (approved):
LGTM (and I actually read it this time)

PR #315, opened 2021-05-28
More margin sorting

gergness:
Consolidate margin proportion and rows base and allow them to be sorted.


tests/unit/matrix/test_cubemeasure.py
```diff
@@ -2235,6 +2235,26 @@ def it_knows_its_columns_pruning_base(self):
...
+                [  # -- row 1 ------------
+                    [7, 8, 9],  # -- selected --
+                    [3, 2, 1],  # -- not --
+                    # --------------------
+                ],
+            ]
+        )
+        cube = _MrXCatMatrix(None, None, unweighted_counts, None)
+
+        assert cube.rows_base.tolist() == [12, 48]
```
ernestoarbitrio, line 2254: i don't understand this test cause we're using the `_BaseResultMatrix` class that If i well remember is something we should get rid of

ernestoarbitrio (changes requested):
per comments

src/cr/cube/matrix/cubemeasure.py
```diff
@@ -1765,7 +1765,7 @@ def rows_base(self):
 
         Only selected values contribute to the rows-base.
         """
-        return np.sum(self._unweighted_counts[:, 0, :], axis=1)
+        return np.sum(self._unweighted_counts[:, 0, :], axis=1) * 2
```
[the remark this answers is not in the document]
gergness: Oh shoot, that should not have been included. I was confused why there was no coverage (I didn't realize this was in the legacy class) and so was testing out if anything broke.

tests/unit/matrix/test_cubemeasure.py
gergness: Yeah, this tests was needed to get coverage back to 100%, I had thought leaving in the legacy code for now and then taking it out in one fell swoop (and maybe with tests) made sense. Do you think I should just take it out? Or put a "no coverage" directive?

ernestoarbitrio: if this is needed for the 100% coverage means that the `rows_base` measure is not fully consolidated with the new architecture. I would understand what's is missed here.

ernestoarbitrio (approved):
LGTM! Good to see all those properties deleted from the legacy class :D

PR #316, opened 2021-05-30
Weighted bases shouldn't add across subvariable dimensions (cat array and num array)

gergness:
We should never be adding counts across the subvariables of arrays (not MRs, Cat arrays nor Numeric arrays), but curently we treat cat array subvariables and numeric array subvariables the same as true categorical variables (CAT). This cleans up this mistake for weighted bases.

Some notes (and examples of what we don't want to be doing) here:
https://www.notion.so/crunch/Examples-of-cr-cube-summing-across-subvariables-3e957442d7bd4688931130f2da664d79

I believe that the unweighted base code could also be cleaned up by a similar effort.


ernestoarbitrio (changes requested):
General Comment:
Basically if a measure property depends on the dimensionality we should declare that property in each subclass. This means that even if the `table_base` and all the other properties you touched are in the `_BaseWeightedCounts` class, now they depends on the row and col dimension, this means that we should define those properties in each subclass and leave the decision logic within the `factory` method.

src/cr/cube/matrix/cubemeasure.py
```diff
@@ -1030,34 +1028,71 @@ def baseline(self):
...
+        """2D ndarray of np.float64 column-proportion denominator for each cell."""
+        # --- Never add across subvariables. If rows dimension is an array, then the
+        # --- column_bases is the same as the counts.
+        return (
+            self.weighted_counts
+            if self._row_dimension_is_array
+            else np.broadcast_to(
+                np.sum(self._weighted_counts, axis=0), self.weighted_counts.shape
+            )
+        )
```
ernestoarbitrio, line 1041:
```suggestion
       if self._row_dimension_is_array:
           return self.weighted_counts
       return np.broadcast_to(
                np.sum(self._weighted_counts, axis=0), self.weighted_counts.shape
            )
```

src/cr/cube/matrix/cubemeasure.py
```diff
@@ -1030,34 +1028,71 @@ def baseline(self):
...
-        return np.sum(self._weighted_counts, axis=0)
+    def column_bases(self):
+        """2D ndarray of np.float64 column-proportion denominator for each cell."""
+        # --- Never add across subvariables. If rows dimension is an array, then the
+        # --- column_bases is the same as the counts.
+        return (
+            self.weighted_counts
+            if self._row_dimension_is_array
+            else np.broadcast_to(
+                np.sum(self._weighted_counts, axis=0), self.weighted_counts.shape
```
ernestoarbitrio, line 1039: why sum on the private `_weighted_counts` and the shape is from the public `weighted_counts`?

src/cr/cube/matrix/cubemeasure.py
```diff
@@ -1030,34 +1028,71 @@ def baseline(self):
...
+        # --- Never add across subvariables. If columns dimension is an array, then the
+        # --- rows_bases is the same as the counts.
+        return (
+            self.weighted_counts
+            if self._column_dimension_is_array
+            else np.broadcast_to(
+                np.sum(self._weighted_counts, axis=1)[:, None],
+                self._weighted_counts.shape,
+            )
+        )
```
ernestoarbitrio, line 1055: same as before

src/cr/cube/matrix/cubemeasure.py
```diff
@@ -1105,27 +1140,21 @@ def baseline(self):
...
+        """2D np.float64 ndarray of column-proportion denominator for each cell."""
+        # --- Never add across subvariables. If rows dimension is an array, then the
+        # --- column_bases is the same as the counts.
+        if self._row_dimension_is_array:
+            return self.weighted_counts
+        # --- only selected counts contribute to the columns bases, which is summed
         # --- across the rows (axis 0).
-        return np.sum(self._weighted_counts[:, :, 0], axis=0)
+        return np.broadcast_to(
+            np.sum(self._weighted_counts[:, :, 0], axis=0), self.weighted_counts.shape
```
ernestoarbitrio, line 1152: why sum on the private _weighted_counts and the shape is from the public weighted_counts?

src/cr/cube/matrix/cubemeasure.py
```diff
@@ -1204,41 +1241,47 @@ def baseline(self):
...
+        # --- row_bases is the same as the counts.
+        if self._column_dimension_is_array:
+            return self.weighted_counts
 
-    @lazyproperty
-    def rows_margin(self):
-        """1D np.float64 ndarray of weighted-N for each matrix row."""
-        return np.sum(self.weighted_counts, axis=1)
+        return np.broadcast_to(
+            np.sum(self.weighted_counts, axis=1)[:, None], self.weighted_counts.shape
```
ernestoarbitrio, line 1258: here you used the same public property 🤷🏽‍♂️

tests/integration/test_cubepart.py
```diff
@@ -1311,6 +1311,52 @@ def it_uses_columns_proportions_for_ca_cat_x_ca_subvar_population_counts(self):
             np.array([100, 100, 100, 100, 100, 100, 100, 100, 100, 100])
         )
 
+    def it_has_bases_that_dont_sum_across_ca_subvars(self):
+        slice_ = Cube(CR.ca_cat_x_ca_subvar).partitions[0]
```
ernestoarbitrio, line 1315: Tipically the fixture name is all CAPITALIZED, please change the name in `CA_CAT_X_CA_SUBVAR` to be compliant with all the tests

tests/integration/test_cubepart.py
```diff
@@ -1311,6 +1311,52 @@ def it_uses_columns_proportions_for_ca_cat_x_ca_subvar_population_counts(self):
...
+                    {
+                        "anchor": "top",
+                        "args": [0, 2],
+                        "function": "subtotal",
+                        "name": "A",
+                    }
+                ],
+            },
+        }
+        slice_ = Cube(CR.ca_cat_x_ca_subvar, transforms=transforms).partitions[0]
```
ernestoarbitrio, line 1344: same

tests/unit/matrix/test_cubemeasure.py
```diff
@@ -1379,37 +1370,59 @@ def it_knows_its_baseline(self, request, counts_with_missings):
...
+            np.array([[9.9, 7.7, 5.5], [9.9, 7.7, 5.5]])
         )
 
-    def it_knows_its_row_bases(self, request, raw_weighted_counts):
-        property_mock(
-            request,
-            _CatXCatWeightedCubeCounts,
-            "rows_margin",
-            return_value=np.array([8.8, 9.9]),
+    def it_knows_its_column_bases_array_x(
```
ernestoarbitrio, line 1385: can you explain me the meaning of the _x? and also why the name of the tests after this one have the _x_ in the names?

ernestoarbitrio: oh i get it, but TBH they're very complicate to interpret. Can we find another way to name these tests?

Usually we use dimensionality but now that you are generalizing the array I can understand the reason of the names but for example when we have a 1D we can just use `it_knows_row_bases_for_array_types` or something similar. Sorry for the noise :D

tests/unit/matrix/test_cubemeasure.py
```diff
@@ -1642,14 +1784,30 @@ def it_knows_its_table_bases(self, request):
             [7.6, 7.6, 7.6],
         ]
 
-    def it_knows_its_table_margin(self, raw_weighted_counts):
+    def it_knows_its_table_margin_x_nonarray(
```
ernestoarbitrio, line 1787: for example here you are testing the case of MR X CAT so it can be `it_knows_table_margin_for_mr_x_cat`

ernestoarbitrio (changes requested):
per comments

src/cr/cube/matrix/cubemeasure.py
gergness: The shape of the public weighted counts is definitely what we want in the output. This code used to be separated out into the row/column margin and then broadcast to be the shape of the counts, but now that it's closer together I agree it's confusing. 

I've changed them use the public `weighted_counts` for both parts in these cases. The math is the same.

tests/unit/matrix/test_cubemeasure.py
gergness: We could add `any`? Do you think `it_knows_its_column_bases_array_x_any` is clearer?

ernestoarbitrio: i would add a for if it's right in english `it_knows_its_column_bases_for_array_x_any`

ernestoarbitrio (approved):
LGTM after last comments

gergness:
Here are the changes required for the companion branch:
https://github.com/Crunch-io/zoom/compare/margin-companion-178322893...base-companion-178317604?expand=1#diff-6e29902d5d6f0933374b2b1c91fa27c8228a90accf6f3267bbc3c3a601e3ab90

ernestoarbitrio (approved):

PR #317, opened 2021-06-02
More measure consolidation to move away from the legacy matrix classes

gergness:
The old `TableStdErrSubtotals` used the legacy matrix class and I wanted to refactor it. Because of the improvements to the base measures, not as much of the special logic was needed.

Does require a few test changes to zoom though, they are here: https://github.com/Crunch-io/zoom/compare/base-companion-178317604...tbl-std-err-companion-178317604


tests/integration/test_headers_and_subtotals.py
```diff
@@ -2621,115 +2621,112 @@ def it_calculates_residuals_for_cat_x_items_x_cats_with_col_insertion(self):
...
-                    0.0002579,
-                    0.00052932,
-                    0.00047783,
-                    0.00082904,
-                    0.00105215,
-                    0.00090766,
-                    0.00107047,
-                    0.00174012,
-                ],
+        assert slice_.table_std_err.tolist() == [
```
ernestoarbitrio, line 2624: why thid expectation has changed? and more important, why this has a different shape than the old one?

src/cr/cube/matrix/assembler.py
```diff
@@ -640,13 +642,7 @@ def table_margin_unpruned(self):
...
     def table_stderrs(self):
         """2D np.float64 ndarray of std-error of table-percent for each matrix cell."""
-        return self._assemble_matrix(
-            TableStdErrSubtotals.blocks(
-                self._cube_result_matrix.table_stderrs,
-                self._dimensions,
-                self._cube_result_matrix,
-            )
-        )
+        return self._assemble_matrix(self._measures.table_stderrs.blocks)
```
ernestoarbitrio, line 645: i would add a `_` in the measure name `table_std_errs`

ernestoarbitrio (commented):
overall looks good. The only thin that i can't understand is the change of the expectation in the integration test

tests/integration/test_headers_and_subtotals.py
gergness: I don't think it has a new shape? I count it as 8 X 11 in both.

It changed because in the legacy matrix the bases for cat arrays sums across the cat array subvariables but in the new new ones they don't. Therefore, the `table_proportions` are different, and thus the `table_variance` changes.

ernestoarbitrio: sorry my bad :(

PR #318, opened 2021-06-03
Consolidate table_margin

gergness:
I'm not sure it's totally worth it, but I kind of like having the versions of the table_margin with fixed dimensionality. Let me know what you think.

### DIFF
--- a/src/cr/cube/matrix/assembler.py
+++ b/src/cr/cube/matrix/assembler.py
@@ -149,11 +149,11 @@ class Assembler(object):
         # --- TODO: Should column_margin only be defined when it's 1D? This would
         # --- require changes to exporter to use the bases to give a
         # --- "column_margin_range"
-        if not self._measures.columns_weighted_counts.is_defined:
+        if not self._measures.columns_weighted_base.is_defined:
             return self.column_weighted_bases
 
         # --- otherwise columns-base is a vector ---
-        return self._assemble_marginal(self._measures.columns_weighted_counts)
+        return self._assemble_marginal(self._measures.columns_weighted_base)
 
     @lazyproperty
     def columns_margin_proportion(self):
@@ -161,7 +161,7 @@ class Assembler(object):
         # --- an MR_X slice produces a 2D columns-margin (each cell has its own N) ---
         # --- TODO: Should colums_margin_proportion only be defined when it's 1D? This
         # --- requires changes to exporter to use the bases to give a "rows_margin_range"
-        if not self._measures.columns_table_proportions.is_defined:
+        if not self._measures.columns_table_proportion.is_defined:
             return self._assemble_matrix(
                 SumSubtotals.blocks(
                     self.columns_margin / self.table_weighted_bases,
@@ -170,7 +170,7 @@ class Assembler(object):
             )
 
         # --- otherwise columns-base is a vector ---
-        return self._assemble_marginal(self._measures.columns_table_proportions)
+        return self._assemble_marginal(self._measures.columns_table_proportion)
 
     @lazyproperty
     def columns_scale_mean(self):
@@ -412,7 +412,7 @@ class Assembler(object):
             return self.row_unweighted_bases
 
         # --- otherwise rows-base is a vector ---
-        return self._assemble_marginal(self._measures.rows_unweighted_counts)
+        return self._assemble_marginal(self._measures.rows_unweighted_base)
 
     @lazyproperty
     def rows_dimension_fills(self):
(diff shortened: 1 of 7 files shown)


src/cr/cube/matrix/assembler.py
```diff
@@ -574,36 +574,43 @@ def table_margin(self):
...
+        CAT_X_CAT matrix produces a scalar value for this property.
+        """
+        # TODO: see TODO in `.table_base_unpruned`
+        if self._measures.table_margin.is_defined:
+            return self._measures.table_margin.value
+        if self._measures.columns_table_margin.is_defined:
+            return self._measures.columns_table_margin.blocks[0]
+        if self._measures.rows_table_margin.is_defined:
+            return self._measures.rows_table_margin.blocks[0]
+        return self._measures.table_weighted_bases.blocks[0][0]
```
ernestoarbitrio, line 613: I'm not seeing anything that take care of order ... does this property need to be underneath the sort-by-value logic?

src/cr/cube/matrix/measure.py
```diff
@@ -2239,6 +2283,61 @@ def is_defined(self):
...
+
+        These are the base-values and the subtotals.
+        """
+        return [
+            self._apply_along_orientation(np.sum, counts) for counts in self._counts
+        ]
+
+    @lazyproperty
+    def is_defined(self):
+        """True if comparable counts are defined.
```
ernestoarbitrio, line 2301: why comparable counts?

ernestoarbitrio (commented):
per comments, I need to understand something more on this PR sorry

src/cr/cube/matrix/assembler.py
```diff
@@ -149,19 +149,19 @@ def columns_margin(self):
         # --- TODO: Should column_margin only be defined when it's 1D? This would
         # --- require changes to exporter to use the bases to give a
         # --- "column_margin_range"
-        if not self._measures.columns_margin.is_defined:
+        if not self._measures.columns_weighted_counts.is_defined:
```
scanny, line 152: @gergness The naming convention is that marginals are singular and (body) measures are plural. The `rows_` and `columns_` prefix helps differentiate, but I think this name needs to be singular.

In this case, I think the name you want is `columns_weighted_base`, meaning "these are the base for each column". The fact that they are computed from counts does not directly describe their role.

gergness: I now see that the scale measures can't treat the margin as a true 1D strand, which is what you were hinting at. So it's true we can't simplify by treating them as true strands.

However, this isn't really why I think of them as counts.
> The fact that they are computed from counts does not directly describe their role.

It's true that counts and bases have different roles, but in the situations that a 1D version of these "bases for a margins" exists, we know there is no array, so the counts and bases are equal. To me, the base's main role is that it's used as the denominator for a proportion calculation.

A part of my consolidation though has been to use the 2D `row_weighted_bases` as the denominator for `row_proportions`, so we don't actually use them as a denominator very often. Instead, the 1D form is actually used as the numerator of the `margin_table_proportions`. 

All of that to say, I think it's a little weird to use something called a "base" as a numerator, but I can see how the "count" name is also confusing, especially if we want to keep the name singular to indicate it's on a margin.

src/cr/cube/matrix/assembler.py
gergness: Looking closer, I think the `table_margin_unpruned` is intentionally unordered. If it were ordered than it would have to be pruned.

gergness:
Closing this and will move it to another pivotal ticket

src/cr/cube/matrix/measure.py
```diff
@@ -1899,24 +1919,22 @@ def _opposing_numeric_values(self):
         )
 
 
-class _MarginProportion(_BaseMarginal):
-    """The 'proportion-margin' (aka rows-margin-proportion/columns-margin-proportion)"""
+class _MarginTableProportion(_BaseMarginal):
+    """The 'margin-table-proportion', proportions of the margin (as if it were a stripe)
+
+    Defined as the margin-weighted-counts divided by the margin-weighted-base.
+    """
```
scanny, line 1926: Hmm, okay, so this bit about the "table" being in the name has to do with the stripe-like idea? We probably need to give that some further thought or rethinking. Happy to chat about that if you like.

scanny (commented):
Okay, some meaty comments I think. I didn't get all the way through the matrix-measure changes because I think those are still going to shift somewhat with other changes like ranges and whatnot. Anyway I can take a closer look at those on the next pass. Happy to cafe for a chat if you want me to say more on any of these.

PR #319, opened 2021-06-04
[#171500470]: Add __repr__ to CubePart

slobodan-ilic:


src/cr/cube/cubepart.py
```diff
@@ -248,6 +249,29 @@ def __init__(self, cube, slice_idx, transforms, population, mask_size):
         self._population = population
         self._mask_size = mask_size
 
+    def __repr__(self):
```
ernestoarbitrio, line 252: Can this property be generalized in the superclass `CubePartition` ? I think so

src/cr/cube/cubepart.py
```diff
@@ -1341,6 +1365,27 @@ def __init__(self, cube, transforms, population, ca_as_0th, slice_idx, mask_size
...
+                self.name,
+                self.dimension_types[0].name,
+            )
+            headers = ["", self.name]
+            table = [
+                [row_label, row] for row_label, row in zip(self.row_labels, self.counts)
+            ]
+            return title + "\n" + tabulate(table, headers)
+        except Exception:
+            return super(_Slice, self).__repr__()
```
ernestoarbitrio, line 1387:
```suggestion
            return super(_Strand, self).__repr__()
```

ernestoarbitrio (changes requested):
Per comments + i think you should cover the lines underneath the Exception

src/cr/cube/cubepart.py
slobodan-ilic: It operates differently on both subclasses, and I believe the behavior is different enough that it warrants extracting the method into each subclass. For strand, e.g., you don't need to handle row labels, while for the slice you do.

ernestoarbitrio (approved):

PR #320, opened 2021-06-07
Consolidate hypothesis testing 178390751

ernestoarbitrio:
Notion page: https://www.notion.so/crunch/Sort-by-value-Transform-for-Dashboards-and-Export-PRD-759062891aa2471e8f9ca9c766ee6e72#440ab7f0a4234e8eb811678762ffb17a

_Goal: support sort-by-value for pairwise significance measures._

All pairwise significance measures have been moved from `methods` to `lazyproperties`. 

The new way to specify the pairwise sig column baseline is to use a new object in the transforms:

```python
        transforms = {
            "columns_dimension": {
                "pairwise_significance": {
                    "column_significance": {"element_id": 0}  # < - normal column
                }
            }
        }

        transforms = {
            "columns_dimension": {
                "pairwise_significance": {
                    "column_significance": {
                        "insertion_id": 1 or "str_id"  # < - insertion column
                    }
                }
            }
        }
```


src/cr/cube/dimension.py
```diff
@@ -954,6 +959,41 @@ def _order_dict(self):
...
 
+class _PairwiseSignificanceSpec(object):
+    """Value object providing access to details of pairwise significance information."""
+
+    def __init__(self, dimension, dimension_transforms_dict):
+        self._dimension = dimension
+        self._dimension_transforms_dict = dimension_transforms_dict
+
+    @lazyproperty
+    def column_idx(self):
```
scanny, line 970: I'd be thinking `.reference_column_idx`. By itself, `.column_idx` could mean a lot of things.

src/cr/cube/dimension.py
```diff
@@ -954,6 +959,41 @@ def _order_dict(self):
         return self._dimension_transforms_dict.get("order") or {}
 
 
+class _PairwiseSignificanceSpec(object):
+    """Value object providing access to details of pairwise significance information."""
```
scanny, line 963: Okay, I'm still getting used to the idea of this being a separate value object rather than just being attributes directly on `Dimension`, but I think it's probably the right choice.

In any case, why wouldn't you move `Cube.alpha_values`, `.only_larger`, etc. down here? You could still get them temporarily from a dimension (which `_Slice` has access to) if they're necessary during the transition and then you wouldn't need the `CubePartition._alpha` and `._alpha_alt`, etc. properties at all, you could just inline the access to these in the pairwise-sig implementations like that of `_Slice.summary_pairwise_indices` that were going away anyway.

src/cr/cube/dimension.py
```diff
@@ -954,6 +959,41 @@ def _order_dict(self):
...
+    @lazyproperty
+    def insertion_idx(self):
+        """int index of insertion column as reference for the pairwise sig tests."""
+        if self._insertion_id:
+            return self._dimension.insertion_ids.index(self._insertion_id)
+        return None
+
+    @lazyproperty
+    def _insertion_id(self):
+        """int insertion-id in the "insertion_id" field of the transform dict."""
```
scanny, line 983: Clarify this docstring. It's also a restatement of where to get it. Make it say what it _means_ instead. I think it identifies the reference column when the reference column is an insertion rather than a "base" column.

src/cr/cube/matrix/assembler.py
```diff
@@ -233,47 +233,64 @@ def means(self):
         """
         return self._assemble_matrix(self._measures.means.blocks)
 
-    def pairwise_indices(self, alpha, only_larger):
+    @lazyproperty
+    def pairwise_indices(self):
         """2D optional ndarray of tuple of int column-idxs pairwise-t threshold.
 
         Raises `ValueError if the cube-result does not include `overlaps` cube-measures.
         """
```
scanny, line 241: What does `"2D optional ndarray"` mean? I think you must mean `"Optional 2D ndarray"`, but that would imply the value can be either `2D ndarray` or `None`. This was true of the prior implementation but no longer appears to be.

Also, I'm not clear why it would (unconditionally) `"Raise ValueError if the cube-result does not include overlaps cube-measures`". Aren't overlaps only required when the columns-dimension is MR?

While you're in there, make sure the back-ticks match, a closing one is missing after `ValueError` and also I don't know why it would be "cube-measures", overlaps is only one cube-measure, isn't it?

src/cr/cube/matrix/assembler.py
```diff
@@ -233,47 +233,64 @@ def means(self):
...
-                    self.pairwise_significance_t_stats(col),
-                    alpha,
-                    only_larger,
-                )
-                for col in range(len(self._column_order))
-            ]
-        ).T
+    @lazyproperty
+    def pairwise_indices_alt(self):
+        """2D optional ndarray of tuple of int column-idxs pairwise-t alt. threshold.
```
scanny, line 253: "Optional 2D ndarray", not 2D optional ndarray.

The " pairwise-t alt. threshold" suffix doesn't make any sense. Just end the summary sentence after "column_idxs." and add a continuation paragraph where you spell out the rest.

Also this one makes the same overgeneralization of when `ValueError` might be raised.

While we're talking about that, is that exception still possible and if so who is catching that exception? I'm not sure we want to propagate that to users, do we?

src/cr/cube/matrix/assembler.py
```diff
@@ -233,47 +233,64 @@ def means(self):
...
+            p_vals, t_stats, self._alpha_alt, self._only_larger
+        )
 
-    def pairwise_means_indices(self, alpha, only_larger):
+    @lazyproperty
+    def pairwise_means_indices(self):
         """2D optional ndarray of tuple of int column-idxs means pairwise-t threshold.
 
         Raises `ValueError if the cube-result does not include `means` cube-measures.
         """
```
scanny, line 273: same docstring problems.

src/cr/cube/matrix/assembler.py
```diff
@@ -233,47 +233,64 @@ def means(self):
...
+        p_vals, t_stats = self._measures.pairwise_means_indices.values
+        return self._assemble_pw_indices(
+            p_vals, t_stats, self._alpha, self._only_larger
+        )
+
+    @lazyproperty
+    def pairwise_means_indices_alt(self):
+        """2D opt. ndarray of tuple of int column-idxs means pairwise-t alt threshold
+
+        Raises `ValueError if the cube-result does not include `means` cube-measures.
```
scanny, line 283: and again.

src/cr/cube/matrix/assembler.py
```diff
@@ -233,47 +233,64 @@ def means(self):
...
+        return self._assemble_pw_indices(
+            p_vals, t_stats, self._alpha_alt, self._only_larger
+        )
 
-    def pairwise_significance_p_vals(self, column_idx):
+    @lazyproperty
+    def pairwise_significance_p_vals(self):
         """2D optional np.float64 ndarray of overlaps-p_vals matrices for subvar idx.
 
         For cubes where the last dimension is categorical, column idxs represent
```
scanny, line 296: and here.

src/cr/cube/matrix/assembler.py
```diff
@@ -286,17 +303,22 @@ def pairwise_significance_p_vals(self, column_idx):
...
             )
-        base_column_idx = self._column_order[column_idx]
-        return self._assemble_matrix(
-            self._measures.pairwise_p_vals(base_column_idx).blocks
-        )
+        return self._assemble_matrix(self._measures.pairwise_p_vals.blocks)
+
+    @lazyproperty
+    def pairwise_significance_p_vals_scale_means(self):
+        """1D optional np.float64 ndarray of p_vals for scale means measure"""
```
scanny, line 315: and here.

src/cr/cube/matrix/assembler.py
```diff
@@ -286,17 +303,22 @@ def pairwise_significance_p_vals(self, column_idx):
...
+        return np.block(self._measures.pairwise_p_vals_scale_means.blocks)[
+            self._column_order
+        ]
```
scanny, line 318: Isn't this a marginal? Why wouldn't we use the `._assemble_marginal()` method for this (or whatever it's called) instead of reimplementing it?

src/cr/cube/matrix/assembler.py
```diff
@@ -286,17 +303,22 @@ def pairwise_significance_p_vals(self, column_idx):
...
 
-    def pairwise_significance_t_stats(self, column_idx):
+    @lazyproperty
+    def pairwise_significance_t_stats(self):
         """2D optional np.float64 ndarray of overlaps-t_stats matrices for subvar idx.
 
         For cubes where the last dimension is categorical, column idxs represent
```
scanny, line 324: revisit docstring

src/cr/cube/matrix/assembler.py
```diff
@@ -309,34 +331,38 @@ def pairwise_significance_t_stats(self, column_idx):
...
-        )
+        return self._assemble_matrix(self._measures.pairwise_t_stats.blocks)
 
-    def pairwise_significance_means_p_vals(self, column_idx):
+    @lazyproperty
+    def pairwise_significance_t_stats_scale_means(self):
+        """1D optional np.float64 ndarray of t-stats for scale means measure"""
+        return np.block(self._measures.pairwise_t_stats_scale_means.blocks)[
+            self._column_order
+        ]
```
scanny, line 346: same as above on reimplementing assemble_marginal and fix docstring.

scanny (changes requested):
Okay, I'm out of time for today, but here are some to get you started.

src/cr/cube/cube.py
```diff
@@ -328,6 +370,15 @@ def inflate(self):
             self._mask_size,
         )
 
+    @lazyproperty
+    def has_overlaps(self):
+        """True if overlaps are defined and the last dimension is MR, False otherwise"""
```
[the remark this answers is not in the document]
ernestoarbitrio: @slobodan-ilic ⬆️  please can you help with this comment? I just moved this property in `cr.cube` cause in my opinion is the best place for it, but TBH it's something you did and I cannot totally get through the steve's comment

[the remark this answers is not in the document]
slobodan-ilic: The proper comment here should probably read something like:
```
"""True if sig testing should be based on `cube_overlaps` and `cube_valid_overlaps` measures.

In cases where the columns dimension is of the Multiple response type, the users usually want
to adjust for the "overlaps" between tested subvariables, which are neither fully independent, nor
fully dependent. The measures indicating this are `cube_overlaps` and `valid_cube_overlaps`,
and are calculated as part of the ZZ9 cube response.

If such measures are present in the payload, and the column variable is indeed an MR, indicate
that we want to calculate the adjusted significance testing (return `True`). If they're not present,
or the column dimension of interest is not an MR, we should default to the unadjusted
significance testing (return `False`).
"""
```
Adjust accordingly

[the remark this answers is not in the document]
ernestoarbitrio: Thanks

src/cr/cube/cubepart.py
```diff
@@ -146,7 +146,8 @@ def variable_name(self):
     @lazyproperty
     def _alpha(self):
         """float confidence-interval threshold for pairwise-t (sig) tests."""
-        return self._alpha_values[0]
+        # TODO: remove me when `measures/pairwise_significance` has been refactored
+        return self._cube.alpha_values[0]
```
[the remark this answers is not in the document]
ernestoarbitrio: I'm inclined to have them in cube cause IMHO they are general transforms and not dimension related. This properties will be here for a very short time cause after the marginal refactoring those will go away as the TODO says

src/cr/cube/dimension.py
ernestoarbitrio: gothca ... thanks nice idea

src/cr/cube/matrix/assembler.py
ernestoarbitrio: i think that marginals has some logic like orientation and is_defined property. We don't have this for pairwise summaries

src/cr/cube/matrix/assembler.py
```diff
@@ -692,6 +771,88 @@ def _assemble_matrix(self, blocks):
...
+                )
+                for i in self._column_order
+            ]
+        ).T
+
+    def _assemble_pw_scale_mean_indices(self, p_vals, t_stats, alpha, only_larger):
+        """List of tuples indicating the significance of scale means values
+
+        Considering this output: [(3,), (2, 3), (3,), ()] and scale_means values
+        [26, 30, 21, 11], each element contains a tuple of other element
```
scanny, line 835: This all doesn't seem right to me. Measures should not be computed in assembler. That's what the measure is for. Otherwise the measure isn't available to other measures for secondary computations.

The idea would be to compute the _payload order_ indices and then just translate those payload column-order indices to sorted column-order indices in assembler. The rest of assembling the matrix should work fine just like it does for any other measure and you don't need to reimplement it here.

Basically you construct a mapping of payload order to sorted order, which is very much like if not exactly the `._column_order` value. Then you visit each tuple in the indices array and exchange each integer in each tuple with the _sorted-order_ index instead of the payload-order index is started with. Then you're done. Everything else happens the same way as any other measure.

scanny (changes requested):
Okay, I didn't get far when I discovered a basic error in how this is organized. Indices need to be computed in the measure and not in assembler. Then assembler's role is just to adjust the indices from payload-order to sorted-order. Everything else (like `._assemble_matrix` happens the same way as any other measure.

src/cr/cube/matrix/assembler.py
ernestoarbitrio: ok I'll try to map the indices

src/cr/cube/cubepart.py
```diff
@@ -695,60 +610,96 @@ def pairwise_means_indices_alt(self):
...
-    def pairwise_significance_means_p_vals(self, column_idx):
-        """Optional 2D ndarray of means significance p-vals matrices for column idx."""
-        # Significance of means difference is available only is cube contains means.
+    @lazyproperty
+    def pairwise_significance_t_stats_scale_means(self):
+        """1D ndarray of t-stats for each pair of compared groups of scale means.
+
+        The t-stats are calculated considering the row dimension numeric values.
+        """
+        return self._assembler.pairwise_t_stats_scale_means
```
scanny, line 669: If this is 1D, I assume it's a marginal. State in the docstring whether it appears as a summary row or summary column (which implies its length is ncols or nrows respectively). Also add an intuitive explanation for what it is, maybe something like ("Summarizes the ...") so we can make sense of the name. Also indicate what a cell value looks like. Is that a tuple of values like `(0.035, 0.130, ..)` one for each significant value? Do whatever it takes to make it so a user can figure out if this is something they're looking for or not. This docstring is the public documentation for this property.

scanny (approved):
Okay, LGTM after some docstring improvement :)

src/cr/cube/dimension.py
```diff
@@ -1300,7 +1300,7 @@ def reference_column_idx(self):
         """Optional int index of base column as reference for the pairwise sig tests."""
```
gergness, line 1300: docstring is out of date now, right? 
"Optional int index" -> "Optional element id" (which is either an int/str if you want to have the type)

(I think Mike's arguing in slack that this should be called the "code", but this would be the only place we use that terminology)

gergness (commented):
I had originally wanted to go through and understand the whole thing, which I haven't had time to do yet. (Not sure what you were asking for, but since Steve already approved, is it good?)

One comment about a docstring change

scanny (commented):
Will not review. Just commenting to clear pending review request.

PR #321, opened 2021-06-08
Sort by "derived" insertions

gergness:
Sort by value had not been updated for the new MR insertions.

### DIFF
--- a/src/cr/cube/dimension.py
+++ b/src/cr/cube/dimension.py
@@ -366,6 +366,23 @@ class Dimension(object):
         )
 
     @lazyproperty
+    def derived_elements(self):
+        """_Elements object providing access to derived elements.
+
+        Derived elements are insertions computed by zz9.
+        """
+        return self.all_elements.derived_elements
+
+    @lazyproperty
+    def derived_element_ids(self):
+        """tuple of insertion-id for each derived element in this dimension.
+
+        Insertion-ids appear in the order that the derived elements are defined. Can
+        be int or string.
+        """
+        return tuple(s.element_id for s in self.derived_elements)
+
+    @lazyproperty
     def description(self):
         """str description of this dimension."""
         # ---First authority in cascade is analysis-specific dimension transform. None
@@ -594,6 +611,11 @@ class _AllElements(_BaseElements):
         self._dimension_type = dimension_type
 
     @lazyproperty
+    def derived_elements(self):
+        """_DerivedElements object containing only derived elements."""
+        return _DerivedElements(self._elements, self._dimension_transforms_dict)
+
+    @lazyproperty
     def valid_elements(self):
         """_ValidElements object containing only non-missing elements."""
         return _ValidElements(self._elements, self._dimension_transforms_dict)
@@ -704,6 +726,25 @@ class _AllElements(_BaseElements):
         }
         shimmed_transforms.update(self._dimension_transforms_dict.get("elements", {}))
         return shimmed_transforms
+
+
+class _DerivedElements(_BaseElements):
+    """Sequence of derived element objects for a dimension.
+
+    *derived_elements* is a sequence of _Elements containing the elements
+    of a dimension that are derived (aka zz9 computed insertions on array variables).
+    This object is only intended to be constructed by _AllElements.derived_elements and
+    there should be no reason to construct it directly.
+    """
+
+    def __init__(self, all_elements, dimension_transforms_dict):
+        self._all_elements = all_elements
+        self._dimension_transforms_dict = dimension_transforms_dict
+
+    @lazyproperty
+    def _elements(self):
+        """tuple containing actual sequence of element objects."""
+        return tuple(element for element in self._all_elements if element.derived)
 
 
 class _ValidElements(_BaseElements):
--- a/src/cr/cube/matrix/assembler.py
+++ b/src/cr/cube/matrix/assembler.py
@@ -833,13 +833,21 @@ class _BaseOrderHelper(object):
         insertion, hiding, pruning, and ordering transforms specified in the
         rows-dimension.
         """
-        order_spec = dimensions[0].order_spec
-        HelperCls = {
-            CM.LABEL: _SortRowsByLabelHelper,
-            CM.MARGINAL: _SortRowsByMarginalHelper,
-            CM.OPPOSING_ELEMENT: _SortRowsByBaseColumnHelper,
-            CM.OPPOSING_INSERTION: _SortRowsByInsertedColumnHelper,
-        }.get(order_spec.collation_method, _RowOrderHelper)
+        collation_method = dimensions[0].order_spec.collation_method
+        dim_type = dimensions[1].dimension_type
+        HelperCls = (
+            _SortRowsByBaseColumnHelper
+            if collation_method == CM.OPPOSING_ELEMENT
+            else _SortRowsByDerivedColumnHelper
+            if collation_method == CM.OPPOSING_INSERTION and dim_type in DT.ARRAY_TYPES
+            else _SortRowsByInsertedColumnHelper
+            if collation_method == CM.OPPOSING_INSERTION
+            else _SortRowsByLabelHelper
+            if collation_method == CM.LABEL
+            else _SortRowsByMarginalHelper
+            if collation_method == CM.MARGINAL
+            else _RowOrderHelper
+        )
 
         return HelperCls(dimensions, second_order_measures)._display_order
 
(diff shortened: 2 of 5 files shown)


src/cr/cube/dimension.py
```diff
@@ -650,6 +672,25 @@ def _iter_element_makings(self):
...
+
+    *derived_elements* is a sequence of _Elements containing the elements
+    of a dimension that are derived (aka zz9 computed insertions on array variables).
+    This object is only intended to be constructed by _AllElements.derived_elements and
+    there should be no reason to construct it directly.
+    """
+
+    def __init__(self, all_elements, dimension_transforms_dict):
+        self._all_elements = all_elements
+        self._dimension_transforms_dict = dimension_transforms_dict
```
slobodan-ilic, line 686: Why is this needed in this class?

tests/integration/test_cubepart.py
```diff
@@ -902,6 +902,26 @@ def it_can_sort_by_column_percent(self):
...
+            }
+        }
+        slice_ = Cube(MRI.CAT_X_MR, transforms=transforms).partitions[0]
+        assert slice_.counts.tolist() == [
+            [0.0, 0.0, 0.0, 0.0],
+            [0.0, 0.0, 0.0, 0.0],
+            [0.6, 0.0, 0.6, 0.0],
+            [0.9, 0.9, 0.0, 0.0],
+            [1.5, 1.5, 1.5, 0.0],
+        ]
```
slobodan-ilic, line 923: It's tricky to see what is what here. Could you please annotate (like with comments) what is what, and what we should expect? Also, I always find it very useful to have an expectation for the case with "no transforms" alongside the one "with transforms", preferably in the same test, which makes it much easier to compare against.

src/cr/cube/matrix/assembler.py
```diff
@@ -1127,6 +1134,26 @@ def _subtotal_values(self):
...
+    being that these insertions must be calculated by zz9.
+    """
+
+    @lazyproperty
+    def _column_idx(self):
+        """int index of column whose values the sort is based on."""
+        # --- MR insertions are called derived elements by cube
+        derived_element_ids = self._columns_dimension.derived_element_ids
+        insertion_id = self._order_spec.insertion_id
+        return derived_element_ids.index(insertion_id)
```
slobodan-ilic, line 1154: I'm not sure I get this concept 100%. I'd love a ☕  session together, so that you can take me through this, at your convenience.

slobodan-ilic (changes requested):
Mostly questions, but one change required in test code.

src/cr/cube/dimension.py
gergness: Hm, on closer look I'm not sure why. I was copying from `_ValidElements` which also has it, but I don't see anything different. Will investigate

tests/integration/test_cubepart.py
gergness: Ha, I really like comparisons too, but in an earlier PR (https://github.com/Crunch-io/crunch-cube/pull/300#discussion_r632144756) Steve said he didn't like those comparisons. Do you have advice on making it more appetizing for Steve? Otherwise, are comments the happy medium?

slobodan-ilic: I'll bring it up in standup, or at the meeting afterwards. Maybe a comment after each line, representing the original order, or something. I really feel I need this as a reader of the tests.

src/cr/cube/dimension.py
scanny: @gergness just a note on `_ValidElements` and possibly applicable to `_DerivedElements`. The reason for a "subset" subclass is essentially to make the collection and special properties and methods defined in `_BaseElements` available on these subset classes. The collection and special properties/methods are:

* indexed access: `elements[idx]` or `elements[n:m]`
* iteration: `for e in elements`
* len: `len(elements)`
* .element_ids: `elements.element_ids -> (1, 4, 5)`
* .element_idxs: `elements.element_idxs -> (0, 3, 4)`
* .get_by_id(element_id): `elements.get_by_id(1) -> elements[0]`

All of these methods are computation on the `._elements` instance variable, so changing that to be a subset provides all these methods for "free" in a subclass. A subclass is also free to define new methods specific to its subset if appropriate.

You could possibly return a tuple of Element objects instead if you knew for sure you didn't need more than the sequence methods (indexed access, iteration, and len) that could just as well be provided by `tuple()`. I'd be inclined to think subclassing is a better bet though so you don't need to compute element-ids or payload-order element-idxs yourself; those seem likely to be needed by someone somewhere along the line. Maybe lookups too if you need to find a derived element for sorting purposes.

tests/integration/test_cubepart.py
scanny: I don't have any problem at all with adding the untransformed order as a comment like:

```python
# --- payload order is ---
# [1.5, 1.5, 1.5, 0.0],
# [0.6, 0.0, 0.6, 0.0],
# [0.0, 0.0, 0.0, 0.0],
# [0.9, 0.9, 0.0, 0.0],
# [0.0, 0.0, 0.0, 0.0],
```

It's just that it muddles the test for me if you're actually _testing_ that the untransformed order is what you expect. Either that needs to be tested or it doesn't (perhaps because it's already tested well enough in other tests, which we would expect to be 95%+ the case). If it _doesn't_ need to be tested, then you have redundancy in the tests which consumes test runtime and increases the cognitive load of the tests.

More importantly for me, you need two fixturings, which breaks the one fixture per test pattern which makes tests quick to understand (because they look like all the other tests) and compact (because there's only one fixture and one expectation.

Also, you can hide a test failure. If the first (redundant) test fails, it prevents the second (real) test from running. If they are separate, the "real" test will be run in any case.

I'm sure there are other reasons, but in any case, I don't have a problem using comments containing baseline (untransformed in this case) results to clarify what's happening. I just don't like the idea of using "actual" tests just for this documentation purpose.

scanny: One other argument, just while I'm thinking about it. Say you had a scenario where col-percents were not appearing properly when in payload order, but were showing up fine when sorted by a derived subvariable or whatever. Then that test would fail saying: `it_sorts_column_percents_in_derived_subvar_order FAILED` whereas the actual problem is more like `it_displays_column_percents_in_payload_order FAILED`. So the name and whatever other documentation focused on the impetus of the test is potentially made misleading because you tested something other than what you said you were testing (before testing the real thing).

gergness: Added a comment indicting which column is the MR insertion. I don't like having the whole cube as a comment because I worry about comments going stale. Hope that's a good enough compromise.

I'll follow the established conventions here, but I'm not super convinced by these arguments. My main argument against comments is that I worry about them going stale. And personally, I don't see them as cognitive overhead. To me, these intermediate tests are more like breadcrumbs that help someone in the future more quickly understand what is being tested. And problems with "hidden" or "misnamed" tests are just inherent to integration tests (what if the JSON parser broke?), so don't seem that problematic to me.

If you know of further reading on this, I'd be interested to read!

src/cr/cube/dimension.py
```diff
@@ -365,6 +365,23 @@ def apply_transforms(self, dimension_transforms):
...
+        return self.all_elements.derived_elements
+
+    @lazyproperty
+    def derived_element_ids(self):
+        """tuple of insertion-id for each derived element in this dimension.
+
+        Insertion-ids appear in the order that the derived elements are defined. Can
+        be int or string.
+        """
+        return tuple(s.element_id for s in self.derived_elements)
```
scanny, line 383: `_DerivedElements` already has `.element_ids` inherited from the `_BaseElements` collection class, so minimally the implementation should be `return self.derived_elements.element_ids` rather than reimplementing that here. At the same time, I'd want to have a story convincing myself it's not better for the caller to get those for itself if `.derived_elements` is already public.

I'm thinking either you let folks have `.derived_elements` and they use them how they want (using the standard `_Elements` interface, _OR_ you keep the collection private and publish direct access to what folks want (like `.derived_element_ids`).

I kind of suspect the latter is what's called for here (like `.derived_elements` becomes `._derived_elements`) but I'll need to get my head into the rest of it here to say.

src/cr/cube/matrix/assembler.py
```diff
@@ -1139,6 +1147,26 @@ def _subtotal_values(self):
         return measure_subtotal_rows[:, self._column_idx]
 
 
+class _SortRowsByDerivedColumnHelper(_SortRowsByBaseColumnHelper):
+    """Orders elements by the values of an opposing derived element vector.
+
+    This would be like "order rows in descending order by value of 'Coke' or 'Diet Coke'
+    array column. An opposing-element ordering is only available on a matrix, because only
+    a matrix dimension has an opposing dimension.
```
scanny, line 1155: Maybe find a place to work in that these only arise on subvariables dimensions, so far only MR but theoretically also possibly NUMA. The entire "inserted" subvariable has to be computed, like a value for every row, and that's why it needs to happen up on `zz9`.

src/cr/cube/matrix/assembler.py
scanny: @gergness are we sure that `insertion_id` matches the `element_id` in all cases? I'm not at all sure that it does.

Such an insertion can arise in the variable-definition, in which case maybe we can depend on those matching, I don't know. But they can also arise in an analysis-definition and I'm not sure how they get numbered there. However they do get numbered there can vary by the client. There's no centralized "authority" that assigns these uniformly.

I wouldn't be surprised if this passes the test by accident of normal sequential `1, 2, 3, ...` ordering or something. I'd say we need to be confident of the contracts we're counting on to identify this inserted element and to document them in the comments here for future folks who come through and need to reconstruct how it works (and possibly reason about why it's now broken :).

tests/integration/test_cubepart.py
scanny: Yeah, I suppose I'm okay with having before and after. I don't like them, and I don't think I would write one like that myself, but if you feel like they make the test easier to make sense of then I won't stand in the way.

I find the point about guaranteeing the "before" version hasn't gotten out of date pretty compelling.

I would like this example to have more different numbers though, you can't tell whether the sort is stable because you can't distinguish between the top rows that are all 0s. It's good to have two same sort values so you demonstrate what happens when values are equal, but the rest of the row needs to be distinguishable from the other row with the same sort value.

scanny (changes requested):
k, mostly looks excellent, just a couple questions to work out ...

gergness:
superseded by #339

PR #323, opened 2021-06-10
Even more consol 178491901

gergness:
A re-creation of #318


scanny:
@gergness my comments are over on #318, sorry if I missed the shift.

src/cr/cube/cubepart.py
```diff
@@ -1153,24 +1153,6 @@ def table_margin(self):
         """
         return self._assembler.table_margin
 
-    @lazyproperty
-    def table_margin_unpruned(self):
```
gergness, line 1157: @scanny - In the other PR (https://github.com/Crunch-io/crunch-cube/pull/318#discussion_r649500157) you suggested removing this. Just making sure we're on the same page, this was an external interface presented on the `_Slice`, and will require changing the exporter. Does this feel worth it?

(The `unpruned_table_margin_range` below will allow us to get the same thing that we actually used in exporter)

src/cr/cube/matrix/cubemeasure.py
```diff
@@ -87,851 +94,1251 @@ def _slice_idx_expr(cls, cube, slice_idx):
         return np.s_[slice_idx]
 
 
-# === MEANS ===
+# === COUNTS (UNWEIGHTED & WEIGHTED) ===
```
gergness, line 97: Note that Unweighted counts have not yet been migrated yet, I realized I should have been asking for more feedback sooner, so I split this PR off.

src/cr/cube/matrix/cubemeasure.py
```diff
@@ -87,851 +94,1251 @@ def _slice_idx_expr(cls, cube, slice_idx):
...
-            else _CatXMrCubeMeans
-            if dimension_types[1] == DT.MR
-            else _CatXCatCubeMeans
+    def factory(cls, weighted, cube, dimensions, slice_idx):
+        """Return _BaseCubeCounts subclass instance appropriate to `cube`
+
+        Chooses between unweighted and weighted counts based on `type`.
+        """
+        dimension_type_strings = tuple(
+            "MR"
```
gergness, line 115: Strings felt gross - didn't like the existing enums because I want to be very specific about how I'm categorizing dimension types and none of the existing ones were good enough.

gergness (commented):
@scanny - can you take another look at this?

src/cr/cube/cubepart.py
scanny: My understanding is that `.table_margin_unpruned` is only used by `exporter` to compute "table-margin-range" when it has determined that the table-margin is not a scalar, probably by seeing that the rows-dimension or columns-dimension is MR.

The table-margin-range value is a legitimate "total" and if exporter needs it (which it does) then `cr.cube` should provide it rather than leaving exporter and other clients to do that for themselves.

If we're going to fix that responsibility-leakage then exporter will have to lose some code and get that value from `_Slice.table_margin_range` or whatever instead. Once that's done the only call(s) to `.table_margin_unpruned` will be gone and it can disappear. There should be no legitimate need for an unpruned anything outside of `cr.cube`.

As far as timing is concerned, when you do that part is up to you, but it's an important part of measure consolidation in my view because that margin-range measure/total is now computed in `exporter` and needs to be consolidated into `cr.cube` so other clients (like SOA and maybe `rcrunch`) don't have to reimplement it.

src/cr/cube/matrix/cubemeasure.py
```diff
@@ -87,851 +94,1251 @@ def _slice_idx_expr(cls, cube, slice_idx):
...
 
+        The name is a mouthful, but each component is meaningful.
+        - "columns": Indicates it is a marginal in the "columns" orientation (kind of
+        like a stripe in the shape of a row).
+        - "table": Indicates that it is the base for the whole table. When the
+        `.table_base` exists (CAT X CAT), it is a repetition of that, but when
+        the rows are array (and therefore we can't sum across them), each cell has
+        its own value.
+        - "base": Indicates that it is the base, not necessarily the counts (eg the sum
+        of selected and non-selected for MR variables)
```
scanny, line 182: Indent second and later lines of "bullets" for readability, like:

```python
        * "columns": Indicates it is a marginal in the "columns" orientation (kind of
           like a stripe in the shape of a row).
        * "table": Indicates that it is the base for the whole table. When the
          `.table_base` exists (CAT X CAT), it is a repetition of that, but when
          the rows are array (and therefore we can't sum across them), each cell has
          its own value.
 ```

Btw, asterisks are the default first-level bullet in reStructuredText, which is how docstrings are interpreted when they are pulled for the documentation. I expect it works either way, but where there's an existing convention there is some reasonable argumentation to follow it.

Also, I think the intuitive interpretation here is that this marginal is only interesting when the columns-dimension is subvariables (array type), right? Otherwise all values are the same as the scalar table-base? Maybe I have that wrong, but whatever it is, it might make sense to lead with the intuitive interpretation; that's the first thing I'm always looking for when I read the docs.

File: src/cr/cube/matrix/cubemeasure.py [hunk 5]
```diff
@@ -87,851 +94,1251 @@ def _slice_idx_expr(cls, cube, slice_idx):
...
 
-class _CatXMrCubeMeans(_BaseCubeMeans):
-    """Means cube-measure for a NOT_MR_X_MR slice.
+        None if rows dimension is an ARRAY or MR because the base can vary among columns.
+        """
+        return None
```
scanny, line 186: This isn't making sense to me as stated. The whole point of this marginal is so the table-base can vary between (among) columns, right? Is what you mean to say: "`None` if rows-dimension is an ARRAY type because there is no single per-column table-base. In that case, each cell of the column can have a distinct table-base."

I suppose a quick re-word would be `s/among/within each`, but seems maybe worth it to spell things out a little more given this isn't so intuitive. Maybe also worth mentioning what this marginal is good for, like it forms the denominator in measure X when dimensions are Y or whatever it is, just to give some more possible grasping points.

src/cr/cube/matrix/cubemeasure.py
```diff
@@ -87,851 +94,1251 @@ def _slice_idx_expr(cls, cube, slice_idx):
...
 
-    Note that the rows-dimensions need not actually be CAT.
-    """
+    @lazyproperty
+    def counts(self):
+        """2D np.float64 ndarray of count for each valid matrix cell.
+
+        A valid matrix cell is one whose row and column elements are both non-missing.
```
scanny, line 192: I think the "valid" bits and definition can go now that unconditional counts are conceptually far away.

src/cr/cube/matrix/cubemeasure.py
```diff
@@ -87,851 +94,1251 @@ def _slice_idx_expr(cls, cube, slice_idx):
...
+        raise NotImplementedError(  # pragma: no cover
+            "`%s` must implement `.counts`" % type(self).__name__
+        )
 
     @lazyproperty
-    def means(self):
-        """2D np.float64 ndarray of means for each valid matrix cell."""
-        return self._means[:, :, 0]
+    def diff_nans(self):
+        return self._diff_nans
```
scanny, line 200: Definitely needs a docstring. What the heck are those darn diff-nans and why do they keep showing up so far from home?

scanny (changes requested):
Okay, I don't think I even made it half-way through, but that's all the time I have tonight. I can probably make some time tomorrow but here's a start anyway.

src/cr/cube/cubepart.py
```diff
@@ -1240,6 +1222,17 @@ def total_share_sum(self):
...
+    @lazyproperty
+    def unpruned_table_margin_range(self):
+        """[min, max] np.float64 ndarray range of the table_margin (table-weighted-base)
+
+        A CAT_X_CAT has a scalar for all table-weighted-bases, but arrays have more than
+        one table-weighted-base. This collapses all the values them to the range, and
+        it is "unpruned", meaning that it is calculated before any hiding or removing
+        of empty rows/columns.
+        """
+        return self._assembler.unpruned_table_margin_range
```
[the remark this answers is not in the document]
gergness: On the name I can change it. My worry was that it's confusing to have something called `table_margin_range` that is not the same as the range of the `table_margin`. I think this maybe comes from me not trusting the docs (or our ability to read them) enough and so am not matching the rest of the codebase's style.

On the array vs tuple, the structure I did copy from stripe. I see why a tuple might be more natural, but I think I'll keep it as an array for now to match.  (learning from the best ;-) https://github.com/Crunch-io/crunch-cube/commit/e57f164598034938cb865d3ac2a985b4b82b2c4b#diff-8605f3bf1bb4d8ebd2ec6c85d1769320435dc2745c9d490417e4398be28de8f2R139)

src/cr/cube/matrix/assembler.py
```diff
@@ -575,36 +575,18 @@ def table_margin(self):
...
 
-            * MR_X_MR - 2D ndarray with a distinct table-margin value per cell.
-            * MR_X - 1D ndarray of value per *row* when only rows dimension is MR.
-            * X_MR - 1D ndarray of value per *column* when only columns dimension is MR.
-            * CAT_X_CAT - scalar float value when slice has no MR dimension.
-
+            * CAT_X_CAT - scalar float value when slice has no ARRAY dimension.
+            * ARRAY_X - 1D ndarray of value per *row* when only rows dimension is ARRAY.
+            * X_ARRAY - 1D ndarray of value per *column* when only column is ARRAY.
+            * ARRAY_X_ARRAY - 2D ndarray with a distinct table-margin value per cell.
```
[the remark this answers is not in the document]
gergness: My goal has been to minimize external interface changes to try to reduce scope. I agree that ideally it would be changed everywhere, but this would also mean altering `_Stripe` to be truly everywhere.

Your suggestion at the end doesn't preserve the interface because `self._assembler.table_weighted_base` is always either a scalar or None, whereas `table_margin` has always been the unknown dimension thing.

src/cr/cube/matrix/cubemeasure.py
```diff
@@ -52,8 +59,8 @@ def unweighted_cube_counts(self):
     @lazyproperty
     def weighted_cube_counts(self):
         """_BaseWeightedCounts subclass object for this cube-result."""
-        return _BaseWeightedCubeCounts.factory(
-            self._cube, self._dimensions, self._slice_idx
+        return _BaseCubeCounts.factory(
+            "weighted", self._cube, self._dimensions, self._slice_idx
         )
```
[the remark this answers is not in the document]
gergness: Great, thanks, that looks better.

src/cr/cube/matrix/cubemeasure.py
gergness: Yeah, there was an unfortunate typo in there, updated

src/cr/cube/matrix/assembler.py
```diff
@@ -149,28 +149,29 @@ def columns_margin(self):
         # --- TODO: Should column_margin only be defined when it's 1D? This would
         # --- require changes to exporter to use the bases to give a
         # --- "column_margin_range"
```
scanny, line 151: I would say yes, same as `rows_margin`. That doesn't mean it needs to be done in this PR, but it's part of the scope of measure consolidation.

tests/unit/matrix/test_assembler.py
```diff
@@ -622,94 +627,114 @@ def and_it_knows_the_scalar_table_base_of_a_CAT_X_CAT_matrix(
...
-        )
+        table_weighted_base_.is_defined = True
+        table_weighted_base_.value = 4242
+        second_order_measures_.table_weighted_base = table_weighted_base_
+        _measures_prop_.return_value = second_order_measures_
         assembler = Assembler(None, None, None)
 
         assert assembler.table_margin == 4242
 
+    def it_knows_table_margin_range(
```
scanny, line 725: it knows _the_ ...

tests/unit/matrix/test_assembler.py
```diff
@@ -881,16 +906,16 @@ def _assemble_matrix_(self, request):
...
+        return instance_mock(request, _MarginUnweightedBase)
 
     @pytest.fixture
     def _BaseOrderHelper_(self, request):
         return class_mock(request, "cr.cube.matrix.assembler._BaseOrderHelper")
 
     @pytest.fixture
-    def margin_proportion_(self, request):
-        return instance_mock(request, _MarginProportion)
+    def margin_table_proportion_(self, request):
```
scanny, line 917: This one too. I realize it was wrongly positioned before you had occasion to touch it.

tests/unit/matrix/test_assembler.py
```diff
@@ -881,16 +906,16 @@ def _assemble_matrix_(self, request):
         return method_mock(request, Assembler, "_assemble_matrix")
 
     @pytest.fixture
-    def base_margin_(self, request):
-        return instance_mock(request, _UnweightedBaseMargin)
+    def margin_unweighted_base_(self, request):
+        return instance_mock(request, _MarginUnweightedBase)
```
scanny, line 910: This should go in alphabetical order, case-insensitive and ignoring any leading underscore.

scanny (commented):
Okay, a few comments, it looks pretty good to go to me after that. Were you going to ship it after this or do more work? Either way I can take another look or just focus on the last commit if you're preparing this to ship, just let me know.

src/cr/cube/matrix/measure.py
```diff
@@ -2191,52 +2418,74 @@ def _stddev_func(self):
...
+    """
 
     @lazyproperty
-    def is_defined(self):
-        """True if counts are defined."""
-        return self._counts_are_defined
+    def value(self):
+        """[min, max] np.float64 ndarray range of the table-weighted-base"""
+        bases = self._cube_measures.weighted_cube_counts.table_bases
+        return np.array([np.min(bases), np.max(bases)])
```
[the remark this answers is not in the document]
gergness: Yeah, I've thought about that. Because of what I've worked on, the division of `measures`, `margins` and `table_values` makes sense to me, and I also think it helps with how similar some of the names are (`rows_unweighted_base` and `row_unweighted_bases`). However, it doesn't help that much (a good 75% would still be in `measures`, and `table_values` would be pretty empty). 

Maybe splitting off the significance tests makes sense?

gergness:
approved in https://crunch-io.slack.com/archives/CHERY4XU0/p1623873262432000

PR #326, opened 2021-06-16
consolidation of unweighted counts

gergness:
Ready to go, requires zoom companion  branch https://github.com/Crunch-io/zoom/pull/4033


src/cr/cube/matrix/measure.py
```diff
@@ -945,23 +1053,25 @@ def blocks(self):
     @lazyproperty
```
gergness, line 1053: @malecki - change to mean significance formula here.

The big change is that I'm also using a version of the column bases that does not add over the subvariables, so results are less significant

src/cr/cube/matrix/measure.py
```diff
@@ -1040,10 +1040,23 @@ def blocks(self):
...
         """2D ndarray of float64 representing p-vals for means pairwise testing."""
-        col_bases = self._cube_measures.old_unwted_cube_counts.column_bases
-        n = col_bases[:, self._selected_column_idx] + col_bases.T
-        df = 2 * (n - 1)
-        return 2 * (1 - t.cdf(abs(self.t_stats), df=df.T))
+        return 2 * (1 - t.cdf(abs(self.t_stats), df=self._df))
+
+    @lazyproperty
+    def _df(self):
+        """TKTKTK"""
```
gergness, line 1038: and oops

tests/integration/test_pairwise_significance.py
```diff
@@ -487,35 +487,35 @@ def test_mr_x_mr_pairwise_t_tests(self):
...
-                    [0.11480579, 0.324030029, 0.49943424, 1.0],
-                    [0.31186072, np.nan, 0.3228962, np.nan],
-                    [0.03833944, 0.053756746, 0.057467762, 1.0],
-                    [0.07831369, 0.002790704, 0.01782960, np.nan],
-                    [0.32659255, 0.32563775, 0.322896241, np.nan],
-                ]
-            ),
-            nan_ok=True,
-        )
+        # transforms = {
```
gergness, line 490: oops these comments were not left in intentionally

gergness (commented):
I need to run, but found a few mistakes in a quick glance, will try to get back to them soon

src/cr/cube/matrix/assembler.py
```diff
@@ -117,14 +116,16 @@ def column_weighted_bases(self):
...
         """1D/2D np.float64 ndarray of unweighted-N for each slice column/cell."""
-        # --- an MR_X slice produces a 2D column-base (each cell has its own N) ---
-        columns_base = self._measures.columns_base
-        if self._rows_dimension.dimension_type in (DT.MR_SUBVAR, DT.NUM_ARRAY):
-            return self._assemble_matrix(columns_base.blocks)
+        # --- an MR_X slice produces a 2D columns-base (each cell has its own N) ---
+        # --- This is really just another way to call the columns_weighted_bases ---
+        # --- TODO: Should column_base only be defined when it's 1D? This would
+        # --- require changes to exporter to use the bases to give a
+        # --- "column_base_range"
```
scanny, line 123: Can't remember if I commented this on another PR, but yes, like rows-base and table-base, if columns-base is not 1D it is undefined and the caller can decide to use columns-base-range instead or grab the 2D column-unweighted-bases and figure out what it wants from there.

Whether that happens in this PR or a future one doesn't matter to me. Wherever is most convenient. It does probably make sense to do it in a PR by itself (or with closely related changes) so it's less of a big deal to coordinate with the `zoom` companion branch.

src/cr/cube/matrix/cubemeasure.py
```diff
@@ -159,6 +165,18 @@ def columns_base(self):
...
+    def columns_pruning_mask(self):
+        """1D bool np.ndarray indicating whether all cells in column are empty
+
+        The columns-pruning-mask indicates when every cell in a column has base values
+        of 0. When the column is *not* an MR, the base used is equal to the columns_base,
+        but when the column is an MR, we use the sum of selected & non-selected values from
+        the column, whereas the columns_base uses just the selected values.
+        """
+        # --- Bases are positive, so we can sum them to see if all of them are equal to 0
+        return self._columns_pruning_base == 0
```
scanny, line 178: Nice elegant solution :)

src/cr/cube/matrix/cubemeasure.py
```diff
@@ -268,6 +298,42 @@ def table_bases(self):
             "`%s` must implement `.table_bases`" % type(self).__name__
         )
 
+    @lazyproperty
+    def _columns_pruning_base(self):
+        """1D bool np.ndarray of the sum of the column-bases for cells in a column
+
+        Used to calculate he columns-pruning-mask, which indicates when every cell in a
```
scanny, line 305: s/he/the

This show up copy/pasted a bunch of times below.

Also s/overriden/overridden in the following paragraph, also repeated several times.

src/cr/cube/matrix/cubemeasure.py
```diff
@@ -268,6 +298,42 @@ def table_bases(self):
...
+        of 0. When the row is *not* an MR, the base used is equal to the rows_base,
+        but when the row is an MR, we use the sum of selected & non-selected values from
+        the row, whereas the rows_base uses just the selected values. These values
+        are not meaningful on their own, but used to calculate the rows-pruning-mask.
+
+        This method works when the row is not an MR, and must be overriden when it is MR.
+        """
+        # --- Bases are positive, so we can sum them to see if all of them are equal to 0
+        # --- To really hammer home the point that this number is invalid, we keep it
+        # --- private. But it's easier to test, so it is separate from the mask.
```
scanny, line 334: Okay, I never thought I'd say this but I think we need to work out a way to _reduce_ the documentation here. There's so much redundancy that you have to dig to find out what's different.

I think a decent approach is to be thorough in the base class (which might not even have an implementation but at least is "centralized"), and then concise in the individual subclass implementations. I'd say a concise form is from this:

```
    @lazyproperty
    def _rows_pruning_base(self):
        """1D bool np.ndarray of the sum of the row-bases for cells in a row

        Used to calculate he rows-pruning-mask, which indicates when every cell in a row
        of 0. When the row is *not* an MR, the base used is equal to the rows_base,
        but when the row is an MR, we use the sum of selected & non-selected values from
        the row, whereas the rows_base uses just the selected values. These values
        are not meaningful on their own, but used to calculate the rows-pruning-mask.

        This method works when the row is not an MR, and must be overriden when it is MR.
        """
        # --- Bases are positive, so we can sum them to see if all of them are equal to 0
        # --- To really hammer home the point that this number is invalid, we keep it
        # --- private. But it's easier to test, so it is separate from the mask.
        return np.sum(self.row_bases, axis=1)
```

to this:

```
    @lazyproperty
    def _rows_pruning_base(self):
        """1D bool np.ndarray of the sum of the row-bases for cells in a row.

        Used to compute the rows-pruning-mask; not meaningful on its own.
        """
        return np.sum(self.row_bases, axis=1)
```

I naturally consult the base-class documentation as a second source, but if we think others won't we could add "see base class for more details" or whatever.

src/cr/cube/matrix/cubemeasure.py
```diff
@@ -1245,682 +1410,3 @@ def baseline(self):
...
-        # --- (not, not) cells of the subtable for each matrix cell.
-        return np.sum(self._unweighted_counts, axis=(1, 3))
-
-    @lazyproperty
-    def unweighted_counts(self):
-        """2D np.float64 ndarray of unweighted-count for each valid matrix cell.
-
-        A valid matrix cell is one whose row and column elements are both non-missing.
-        """
-        return self._unweighted_counts[:, 0, :, 0]
```
scanny, line 1926: God I love long swaths of red ink, don't you? I feel so _clean_ afterward :)

src/cr/cube/matrix/measure.py
```diff
@@ -945,23 +996,25 @@ def blocks(self):
     @lazyproperty
     def t_stats(self):
         """2D ndarray of float64 representing t-stats for means pairwise testing.
-
```
scanny, line 948: This blank line should stay to separate the summary line from the continuation paragraph.

src/cr/cube/min_base_size_mask.py
```diff
@@ -4,10 +4,7 @@
 
 from __future__ import division
 
-import numpy as np
-
 from cr.cube.util import lazyproperty
-from cr.cube.enums import DIMENSION_TYPE as DT
 
 
 class MinBaseSizeMask(object):
```
scanny, line 10: Is there a plan for this module to go away and it's contents to be consolidated into measures and wherever?

scanny (approved):
Okay, looks pretty good to me, a few comments, I think the big thing is getting cookie-cutter docstrings to be more concise and avoid bloating the file and avoiding eye glaze-over with the same pattern showing up over and over with the only differences being obscure. Approving in advance.

src/cr/cube/matrix/cubemeasure.py
gergness: Ha, there's a Berenstain bears book where the kids try to be so polite that it'll drive the Mom crazy, but it backfires and they end up forgetting about the plan and just become "regular" polite. Hopefully that'll happen to me. (Makes sense about having the base class as the central place, as I wrote that, I wasn't sure how we felt about doing that).

src/cr/cube/min_base_size_mask.py
gergness: Sure, it can. There's a lot of boiler plate consolidation like that in my last (current) PR, so I can add it to that.

PR #329, opened 2021-06-22
[178611312]: sort-by-value stripe population consol

gergness:
- Consolidate population measure so that it can be sorted by
- Add fallback to payload order for 1D sort-by-value


scanny (approved):
Added commit to improve tests, LGTM after that :)

PR #330, opened 2021-06-22
[#178579890]: Hide MR insertions

slobodan-ilic:
* Drive by TDD
* Implement functionality by adding hidden element IDs for MR insertions

### DIFF
--- a/src/cr/cube/dimension.py
+++ b/src/cr/cube/dimension.py
@@ -628,22 +628,30 @@ class _AllElements(_BaseElements):
         """Element transform dict expressed in the dimension transforms expression."""
         transforms = self._dimension_transforms_dict.get("elements", {})
 
-        # ---include MR insertions into "hide" consideration
+        # ---include MR insertions into consideration for hidden elements
         if self._dimension_type == DT.MR:
-            hidden_mr_insertions = tuple(
-                tr
-                for tr in self._dimension_transforms_dict.get("insertions", [])
-                if tr.get("hide")
-            )
-            ids_by_names = {
-                eld["value"]["references"]["name"]: eld["id"]
-                for eld in self._element_dicts
-            }
-            for ins in hidden_mr_insertions:
-                el_id = ids_by_names[ins["name"]]
-                transforms[el_id] = {"hide": True}
+            return {**transforms, **self._hidden_mr_insertions_transforms}
 
         return transforms
+
+    @lazyproperty
+    def _hidden_mr_insertions_transforms(self):
+        """Element transform dict for MR insertions, based on insertions' hide field.
+
+        If we're operating with MR insertions, their hiddenness is generally expressed
+        via the "hide" field on the transform's insertion object. Since the MR
+        insertions are part of the actual data (that comes from ZZ9) we need to map
+        this information to the elements transforms dictionary, to perform the actual
+        hiding of those (MR insertions) elements.
+        """
+
+        mr_insertions = self._dimension_transforms_dict.get("insertions", [])
+        hidden_mr_insertions = [ins for ins in mr_insertions if ins.get("hide")]
+
+        # ---Create internal mapping between element int IDs and str IDs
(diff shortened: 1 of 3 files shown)


src/cr/cube/dimension.py
```diff
@@ -626,7 +626,24 @@ def _elements(self):
     @lazyproperty
     def _elements_transforms(self):
         """Element transform dict expressed in the dimension transforms expression."""
-        return self._dimension_transforms_dict.get("elements", {})
+        transforms = self._dimension_transforms_dict.get("elements", {})
+
+        # ---include MR insertions into "hide" consideration
+        if self._dimension_type == DT.MR:
```
gergness, line 632: Should this be `in DT.ARRAY_TYPES` for future-proofing?  I know we only support MR for now, but I thought that numeric arrays would also require these derived zz9 insertions, and may not be too far behind. (This would allow cat array subvars too, but on the subvars dim, we would need derived insertions too, right?)

gergness (commented):
One comment, I think you understand the consequences of my proposed change better than I do, so feel free to ignore it if there's something I'm not thinking of.

src/cr/cube/dimension.py
```diff
@@ -626,7 +626,32 @@ def _elements(self):
...
+        hiding of those (MR insertions) elements.
+        """
+
+        mr_insertions = self._dimension_transforms_dict.get("insertions", [])
+        hidden_mr_insertions = [ins for ins in mr_insertions if ins.get("hide")]
+
+        # ---Create internal mapping between element int IDs and str IDs
+        ids_map = {eld["value"]["id"]: eld["id"] for eld in self._element_dicts}
+
+        return {ids_map[ins["name"]]: {"hide": True} for ins in hidden_mr_insertions}
```
scanny, line 654: This together with the "call" up on line 633 require too much study to figure out what its doing, partly because part of it is being done here (computing added "hide": elements) and part is being done there (merging those into the element-transforms).

I'd be thinking the lines in `._elements_transforms` be something like:

```python
return (
    self._shimmed_element_transforms
    if self._dimension_type in DT.ARRAY_TYPES
    else self._dimension_transforms_dict.get("elements", {})
)
```

Then `._shimmed_element_transforms` returns an element-transforms dict and encapsulates the whole job in one place.

It's complicated enough that it may make sense for `._shimmed_element_transforms` to have a helper method or even be it's own helper class, but that should be driven by the test I'd say. If the property is a little complicated but the test is reasonable then I'd say just wrap it all up in one lazyproperty, especially since we hope it's going away.

Maybe something like this:

```python
    @lazyproperty
    def _shimmed_element_transforms(self):
        """Element transforms dict for array dimensions.

        To provide consistency with a poorly-defined interface for categorical
        insertions, a client can include a `"hide": true` field in a (complete) copy of
        a variable-based insertion in order to suppress that variable-based insertion.

        For the array case, these need to be translated to a "hide" transform on the
        subvariable element because such an insertion becomes a derived subvariable just
        like the other subvariables in the dimension.
        """
        # --- currently an inserted-subvariable can only be identified by name, there is
        # --- no alias for an inserted-subvariable and it does not receive a "normal"
        # --- element.id like "0001".
        hidden_insertion_names = tuple(
            insertion["name"]
            for insertion in self._dimension_transforms_dict.get("insertions", [])
            if insertion.get("hide", False)
        )

        # --- however, the hide-transform must be identified by element-id, so we need a
        # --- mapping of insertion-name to element-id
        element_id_from_name = {
            element["value"]["id"]: element["id"] for element in self._element_dicts
        }

        # --- merge hide transforms with (a copy of) the existing element transforms ---
        return {
            **self._dimension_transforms_dict.get("elements", {}),
            **{
                element_id_from_name[name]: {"hide": True}
                for name in hidden_insertion_names
            },
        }
```

scanny (changes requested):

scanny (approved):
LGTM :)

src/cr/cube/dimension.py
slobodan-ilic: refactored

PR #331, opened 2021-06-22
[178371707]: consolidate stripe counts and fix num-array base for weighted

gergness:
This consolidates weighted & unweighted counts in stripe (to match how they are in matrix) and as a consequence avoids adding over num array subvariables for the weighted base (a previous pass had fixed this for unweighted bases only).


scanny (approved):
LGTM :)

PR #336, opened 2021-07-12
isort cr.cube

slobodan-ilic:


ernestoarbitrio (approved):
Awesome 👏🏻

src/cr/cube/cube.py
```diff
@@ -5,7 +5,12 @@
 CubeSet is the main API class for manipulating Crunch.io JSON cube responses.
 """
 
-from __future__ import absolute_import, division, print_function, unicode_literals
+from __future__ import (
+    absolute_import,
+    division,
+    print_function,
+    unicode_literals,
+)
```
scanny, line 13: This line was fine as it was. `isort` needs `-l 88` or `--line-length 88` to wrap at the same place as Black.

src/cr/cube/cube.py
```diff
@@ -14,7 +19,9 @@
 
 from cr.cube.cubepart import CubePartition
 from cr.cube.dimension import AllDimensions
-from cr.cube.enums import CUBE_MEASURE, DIMENSION_TYPE as DT, NUMERIC_MEASURES
+from cr.cube.enums import CUBE_MEASURE
+from cr.cube.enums import DIMENSION_TYPE as DT
+from cr.cube.enums import NUMERIC_MEASURES
```
scanny, line 24: This one was also fine as it was. Repeating the `from x.y.module import ...` prefix obscures the boundaries between source modules in the import list. Adding `--multi-line 8` will change it to the format we already use.

scanny (changes requested):
Okay, now that I've seen all the fixes this made I feel a little better about it. The two configuration changes I've identified are must-haves, there are several unneeded changes when those are left out.

The only thing I still don't like is repeating the `from x.y.module import ...` prefix on imports that rename the imported item, like `from enums import MEASURES as M`. Maybe one of the configuration options can get rid of that.

tests/integration/test_cube.py
```diff
@@ -14,14 +14,12 @@
...
 from cr.cube.enums import DIMENSION_TYPE as DT
 
-from ..fixtures import (
-    CR,  # cube-response
-    OL,  # overlaps
-    NA,  # numeric-array
-)
+from ..fixtures import CR  # cube-response
+from ..fixtures import NA  # numeric-array
+from ..fixtures import OL  # overlaps
```
scanny, line 22: These comments should go above so these don't break onto separate redundant lines. Or maybe we just get rid of the comments altogether.

scanny (commented):
Just this one change I see, otherwise these look real good, that's a lot of fixes it caught :)

ernestoarbitrio:
@slobodan-ilic I'm gonna close this PR and open (soon) a new one with `isort` latest version.

PR #337, opened 2021-07-13
[178303395]: Smoothing consolidation

ernestoarbitrio:
Refactor `smoothing`:
- Remove `evaluate` method from `cubepart`
- Introduce `SmoothingSpec` at `dimension` level via `transforms`
- Smoothed measure dispatched directly within the `measures` modules in `stripe` and `matrix`

PVT Story: https://www.pivotaltracker.com/story/show/178303395


src/cr/cube/matrix/measure.py
```diff
@@ -38,12 +39,34 @@ def column_comparable_counts(self):
 
     @lazyproperty
     def column_index(self):
-        """_ColumnIndex measure object for this cube-result."""
+        """_ColumnIndex measure object for this cube-result.
```
scanny, line 42: There's no demand for smoothed column-index measure yet, let's not implement it until we need it. A lot could change between then and now and maybe we never need it (YAGNI).

scanny (changes requested):
Okay, here are some initial remarks to get you started ...

src/cr/cube/stripe/measure.py
```diff
@@ -7,6 +7,7 @@
 import numpy as np
 
 from cr.cube.enums import DIMENSION_TYPE as DT
+from cr.cube.noa.smoothing import SingleSidedMovingAvgSmoother
```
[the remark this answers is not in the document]
ernestoarbitrio: That's not true. We smooth mean for a stripe. And we have tests in exporter and in here that proof this

src/cr/cube/matrix/measure.py
```diff
@@ -14,6 +14,7 @@
     NanSubtotals,
     NoneSubtotals,
 )
+from cr.cube.noa.smoothing import SingleSidedMovingAvgSmoother
```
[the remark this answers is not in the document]
ernestoarbitrio: We do need smoothing for strands (like for mean measure) so I would leave this class in this subpackage

src/cr/cube/matrix/measure.py
ernestoarbitrio: IMHO, considering that we supported col index smoothed before (this was for free cause in the old implementation col index was directly computed from the smoothed col proportions), I'm inclined to leave it as well after the consolidation because even if it's not exposed in the webapp or the exporter it (maybe) is used through the API or via 3rd party apps of customers. If we remove it we will introduce a regression and that's the reason I would leave it here.

src/cr/cube/dimension.py
```diff
@@ -487,11 +487,13 @@ def shape(self):
...
-        """List of _SmoothingSpec object for transforms.smoothing dict from payload."""
-        smoothing_specs = _SmoothingSpecs(self._dimension_transforms_dict)
-        return smoothing_specs()
+    def smoothing_spec(self, measure):
+        """Optional _SmoothingSpec object for transforms.smoothing dict from payload."""
+        smoothing_dict = self._dimension_transforms_dict.get("smoothing", {})
+        measure_smoothing_dict = smoothing_dict.get(measure.value)
+        if not measure_smoothing_dict:
+            return None
+        return _SmoothingSpec(smoothing_dict.get(measure.value, {}))
```
scanny, line 496: Wouldn't this be enough?

```python
def smoothing_spec(self, measure):
    """ ... """
    smoothing_dict = self._dimension_transforms_dict.get("smoothing", {}).get(measure.value)
    return _SmoothingSpec(smoothing_dict) if smoothing_dict else None
```

src/cr/cube/matrix/measure.py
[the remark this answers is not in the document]
scanny: Yeah, okay, that makes sense now Mike has clarified :)

src/cr/cube/matrix/measure.py
scanny: Yeah, I'm okay with that.

src/cr/cube/matrix/measure.py
```diff
@@ -38,12 +44,32 @@ def column_comparable_counts(self):
...
+        if smoothing_spec:
+            return _ColumnIndexSmoothed(
+                self._dimensions, self, self._cube_measures, smoothing_spec
+            )
         return _ColumnIndex(self._dimensions, self, self._cube_measures)
 
     @lazyproperty
     def column_proportions(self):
-        """_ColumnProportions measure object for this cube-result."""
+        """_ColumnProportions measure object for this cube-result.
```
scanny, line 62: So one of the things we need to think about in the design is how we could provide _both_ smoothed and unsmoothed values. The idea is that a chart would have the smoothed values as the line and the unsmoothed values as "points", kind of bouncing around on either side of that line.

For this use case, it won't do to have smoothing be unconditionally automatic. We need to be able to get both.

We'll need to noodle how they ask for this too, because that's going to matter. Either it can be by asking for both measures, like `"measures": ["col_percent", "smoothed_col_percent"]` or it could be by making the smoothed_col_percent item an insertion, like a new row. I don't think this second one is the right answer, but I'm not 100% sure the first one is right either.

Anyway, something for us to noodle on.

In any case, I'm thinking that `"smoothed_col_percent"` kind of makes sense, in the way that `significant_columns` does, where you ask for it in one place and then specify what the parameters are in another. Like we have thresholds for significance that are separate from `significant_columns`. And maybe we can decide that all smoothing will be done the same way, regardless of measure, so there can just be an optional `"smoothing":` field in transforms and we don't need separate ones for each measure (because that spec no longer needs to specify which measures to smooth). Then we could have a default smoothing method and parameters, like (one-sided, window=3) so we don't even have to care whether the user specified or not.

The only thing left is how you specify which axis to smooth along. I'm thinking it's always the one with a CAT_DATE type, or maybe defaulting to smoothing-across-columns. Anyway, a bit more for us to noodle. Maybe that can go in the smoothing spec as well, defaulting to the CAT_DATE dimension or columns.

scanny (commented):
A few more comments. We should discuss when you have time.

src/cr/cube/dimension.py
```diff
@@ -1233,6 +1239,30 @@ def _order_dict(self):
...
+    def __init__(self, smoothing_dict):
+        self._smoothing_dict = smoothing_dict
+
+    @lazyproperty
+    def function(self):
+        """String representing the smoothing function specified in the payload.
+
+        Default smoothing function is ``one_sided_moving_avg``
+        """
+        return self._smoothing_dict.get("function", "one_sided_moving_avg")
```
scanny, line 1254: This will be more robust as:

```python
return self._smoothing_dict.get("function") or "one_sided_moving_avg"
```

This alternate expression handles the cases:

```python
"smoothing": {"function": null}
"smoothing": {"function": ""}
"smoothing": {"function": false or 0 or []}
```

Of all of these, the `null` case is most likely, but it's a welcome resiliency for any of the falsey cases.

src/cr/cube/dimension.py
```diff
@@ -1233,6 +1239,30 @@ def _order_dict(self):
...
+
+    @lazyproperty
+    def window(self):
+        """int, the window parameter specified in the smoothing expression.
+
+        The default value (if window is not specified) is 2 for the moving avarage
+        smoothing algorithms.
+        """
+        return self._smoothing_dict.get("window", 2)
```
scanny, line 1263: That same approach might be applicable here as well since while `0` is a valid integer, it is outside the range of a valid smoothing window.

```python
return self._smoothing_dict.get("window") or 2
```

src/cr/cube/dimension.py
```diff
@@ -1233,6 +1239,30 @@ def _order_dict(self):
...
+
```
scanny, line 1264: The more I think about this the more I think that a smoothing spec should not exist like this. This works fine for the current case because there is only one type so far. But how would it work if there were two or three or four smoothing types that had different fields?

I think what makes more sense is for `Dimension` to provide the raw smoothing dict from the transforms (keeping in mind the smoothing-dict isn't actually a transform at all, it's just hitching a ride there as an option). This dict is retrieved into a `_BaseSmoother` object over in measures or whatever which feeds it into its factory to produce the right smoother subtype. Then the smoother subtype interprets the dict fields directly. All smoother types have a `"function":` keyword (which serves as its `"type":` field), but all other fields (function arguments) depend on the particular function.

Let me know if you want to discuss to understand this better, but this is the intuition I had initially about this `SmoothingSpec`, it just took me seeing it this third time to actually crystalize the clear rationale for it being elsewhere.

src/cr/cube/enums.py
```diff
@@ -143,17 +147,28 @@ class CUBE_MEASURE(enum.Enum):
...
+        CUBE_MEASURE.SUM,
         CUBE_MEASURE.STDDEV,
-        CUBE_MEASURE.VALID_COUNT_UNWEIGHTED,
-        CUBE_MEASURE.VALID_COUNT_WEIGHTED,
+        CUBE_MEASURE.UNWEIGHTED_VALID_COUNT,
+        CUBE_MEASURE.WEIGHTED_VALID_COUNT,
+    )
+)
+
+SECOND_ORDER_NUMERIC_MEASURES = frozenset(
```
scanny, line 164: I think this could just be `NUMERIC_MEASURES` for conciseness. It would be unlikely to be confused with `NUMERIC_CUBE_MEASURES` I expect and we don't generally make a big deal about specifying "second-order" elsewhere. When we say "measure" we always mean one of these and when we mean "cube-measure" we always spell it out.

Plus, if this is imported by a client like exporter, they have no need for the concept of "second-order" measure since cube-measures are a `cr.cube` internal concept.

scanny (changes requested):

scanny:
@ernestoarbitrio happy to discuss this latest structural change if you like, just ping when you're available.

src/cr/cube/dimension.py
ernestoarbitrio: I got your point, and it's totally legit. But I think that `_BaseSmoother` should be in the `smoothing` because we smooth measure in both stripe and matrix so if we put this object in the `measure.py` we should duplicate it wheter it's needed in a stripe measure or a matrix measure. I'll try to move the spec somewhere i think it feets. Then we can discuss on this new implementation

src/cr/cube/matrix/measure.py
```diff
@@ -613,18 +611,6 @@ def _intersections(self):
...
-
-        Raises `NotImplementedError` if the function in the smooting spec is different
-        from `one_sided_moving_avg` the only function available today as smoothing
-        algorithm.
-        """
-        function = smoothing_spec.function
-        if function != "one_sided_moving_avg":
-            raise NotImplementedError("Function {} is not available.".format(function))
-        return SingleSidedMovingAvgSmoother(smoothing_spec.window, self._dimensions[-1])
-
```
ernestoarbitrio: This goes away cause the actual smoother is returned directly in the `SmoothingSpec` factory

src/cr/cube/smoothing.py
```diff
@@ -7,12 +7,35 @@
...
+        that at the moment is the only one implemented.
+        """
+        smoothing_dict = dimension.smoothing_dict
+        function = smoothing_dict.get("function") or "one_sided_moving_avg"
+        if function != "one_sided_moving_avg":
+            raise NotImplementedError("Function {} is not available.".format(function))
+        return _SingleSidedMovingAvgSmoother(
+            window=smoothing_dict.get("window") or 2,
+            dimension_type=dimension.dimension_type,
+        )
```
ernestoarbitrio, line 30: Returns directly the smoother and not the spec cause the smoother is the actual object we need to apply the algorithm to the second order measures in the measure.py module

src/cr/cube/matrix/measure.py
```diff
@@ -427,9 +427,9 @@ def smoothed_column_proportions(self):
         object will be returned, otherwise fallback to unsmoothed measure object.
         """
         column_dimension = self._dimensions[-1]
-        smoothing_spec = column_dimension.smoothing_spec
+        smoother = BaseSmoothingSpec.smoother(column_dimension)
```
ernestoarbitrio, line 430: we get the smoother directly from the `BaseSmoothingSpec` and pass it to the measure class

scanny (changes requested):
Okay, a couple further steps but we're getting close :)

scanny (approved):
LGTM :)

PR #338, opened 2021-07-15
Allows sorting by label

gergness:
Implements sort-by-label in cube. 

I think it's pretty much ready, but I have one question in comments about how to make something more idiomatic - we special case `nan`s when sorting, but `np.isnan` gives error on string types (before all of the sortable things were numeric, but we're now adding a string type in label).

### DIFF
--- a/src/cr/cube/collator.py
+++ b/src/cr/cube/collator.py
@@ -389,17 +389,17 @@ class SortByValueCollator(_BaseCollator):
         These values appear in sorted order. The sequence is determined by the
         `._target_values` property defined in the subclass and the "top" and "bottom"
         anchored elements specified in the `"order": {}` dict.
+
+        If values are nan, they are placed at the end in their original payload order.
         """
         fixed_idxs = frozenset(self._top_fixed_idxs + self._bottom_fixed_idxs)
-        sorted_value_idx_pairs = sorted(
-            (
-                (value, idx)
-                for idx, value in enumerate(self._element_values)
-                if idx not in fixed_idxs
-            ),
-            reverse=self._descending,
-        )
-        return tuple(idx for _, idx in sorted_value_idx_pairs)
+        keys, nans = [], []
+        for i, val in enumerate(self._element_values):
+            if i not in fixed_idxs:
+                group = nans if self._is_nan(val) else keys
+                group.append((val, i))
+
+        return tuple(idx for _, idx in (sorted(keys, reverse=self._descending) + nans))
 
     @lazyproperty
     def _bottom_fixed_idxs(self):
@@ -431,6 +431,19 @@ class SortByValueCollator(_BaseCollator):
         """
         return self._order_spec.descending
 
+    @staticmethod
+    def _is_nan(value):
+        """True if `value` is NaN, False otherwise.
+
+        We can't use `np.isnan` directly because we want to treat any string value as
+        non-NaN (for eg sorting by label), but numpy barfs on types without NaN
+        defined.
+        """
+        try:
+            return np.isnan(value)
+        except TypeError:
+            return False
+
     def _iter_fixed_idxs(self, fixed_element_ids):
         """Generate the element-idx of each element represented by `fixed_element_ids`.
 
@@ -451,6 +464,9 @@ class SortByValueCollator(_BaseCollator):
         These values appear in sorted order. The values are determined by the
         `._subtotal_values` property defined in the subclass and the sort direction
         is derived from the `"order": {}` dict.
+
+        If subtotal values are nan, they are placed at the end in their original
+        payload order.
         """
         subtotal_values = self._subtotal_values
         n_values = len(subtotal_values)
@@ -459,12 +475,7 @@ class SortByValueCollator(_BaseCollator):
         keys, nans = [], []
         for i, val in enumerate(subtotal_values):
             neg_idx = i - n_values
-            # --- TODO: this try except does not seem idiomatic, but np.isnan fails
-            # --- on strings, and we now sort by labels which are strings
-            try:
-                group = nans if np.isnan(val) else keys
-            except TypeError:
-                group = keys
+            group = nans if self._is_nan(val) else keys
             group.append((val, neg_idx))
 
         return tuple(idx for _, idx in (sorted(keys, reverse=self._descending) + nans))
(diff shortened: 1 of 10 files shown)


src/cr/cube/collator.py
```diff
@@ -459,7 +459,12 @@ def _subtotal_idxs(self):
         keys, nans = [], []
         for i, val in enumerate(subtotal_values):
             neg_idx = i - n_values
-            group = nans if np.isnan(val) else keys
+            # --- TODO: this try except does not seem idiomatic, but np.isnan fails
```
gergness, line 462: I don't like this code, but wasn't sure the best way to accomplish it.

src/cr/cube/dimension.py
```diff
@@ -479,6 +487,14 @@ def shape(self):
         """int count of *all* elements in this dimension, both valid and missing."""
         return len(self.all_elements)
 
+    @lazyproperty
+    def subtotal_labels(self):
+        """tuple of string element-labels for each valid subtotal in this dimension.
+
+        Element-labels appear in the order defined in the cube-result.
+        """
+        return tuple(e.label for e in self.subtotals)
```
scanny, line 496: This one I wasn't so sure about but I think I've come around.

`Subtotals` is a collection that is already used by callers, and I bet somewhere in the code we already have an expression like `s.label for s in dimension.subtotals` (might be worth looking for that and changing it)[Ah, I see you caught that one :)]. My initial thought would be to find these at `dimension.subtotals.labels`, but I think you're right:

1. `dimension.subtotal_labels` is parallel to `dimension.element_labels` and therefore less surprising than `dimension.subtotals.labels`.
2. The likelihood of needing any other `dimension.subtotal_{x}s` seems low, except maybe in future `.subtotal_element_ids`, but that would be justified by rationale 1 above (actually looks like we have `.insertion_ids` already, now that I look a second time).

Okay, cool, just took me a minute to talk myself into it.

scanny, line 496: I think `s` instead of `e`, wouldn't you? like `s.label for s in self.subtotals`?

src/cr/cube/matrix/assembler.py
```diff
@@ -1161,6 +1159,43 @@ def _subtotal_values(self):
...
+            return SortByValueCollator.display_order(
+                self._rows_dimension,
+                self._element_values,
+                self._subtotal_values,
+                self._empty_row_idxs,
+            )
+        except ValueError:
+            return PayloadOrderCollator.display_order(
+                self._rows_dimension, self._empty_row_idxs
+            )
```
scanny, line 1187: It's interesting that this exact implementation of `._order` is duplicated in all 5 sort-by-value Helper classes but can't be extracted to the base class `_RowOrderHelper` because that already has one (this implementation is specific to sort-by-value).

The current object hierarchy looks like this:

* `_BaseOrderHelper`
  * `_ColumnOrderHelper`
  * `_RowOrderHelper`
    * `_SortRowsByBaseColumnHelper`
    * `_SortRowsByInsertedColumnHelper`
    * `_SortRowsByLabelsHelper`
    * `_SortRowsByMarginalHelper`

The object hierarchy is already a bit deep, but I'm thinking we should probably extract a `_BaseSortRowsByValueHelper` to be the superclass to the five sort-by-value classes and move this `._order` implementation there.

src/cr/cube/stripe/assembler.py
```diff
@@ -121,8 +121,7 @@ def row_labels(self):
         row.
         """
         return np.array(
-            [e.label for e in self._rows_dimension.valid_elements]
-            + [s.label for s in self._rows_dimension.subtotals]
+            self._rows_dimension.element_labels + self._rows_dimension.subtotal_labels
```
scanny, line 124: Nice.

src/cr/cube/stripe/assembler.py
```diff
@@ -367,6 +368,49 @@ def _display_order(self):
...
+            return SortByValueCollator.display_order(
+                self._rows_dimension,
+                self._element_values,
+                self._subtotal_values,
+                self._empty_row_idxs,
+            )
+        except ValueError:
+            return PayloadOrderCollator.display_order(
+                self._rows_dimension, self._empty_row_idxs
+            )
```
scanny, line 393: Same for this. The implementation is identical so I'm thinking we should extract a sort-by-value base class.

scanny (changes requested):
Okay, a couple changes, mostly technical debt that had accumulated before this effort.

src/cr/cube/collator.py
gergness: Hm, now that I think about it, I don't see why only the subtotals behave this way, body values can also be nan. For example, the scale mean marginal would be nan if a row only has counts in columns that do not have a numeric value assigned.

src/cr/cube/matrix/assembler.py
gergness: Oh, didn't see you proposed a name. I chose `_SortRowsBaseHelper` to try to cheat and keep things in alphabetical order and also valid.

src/cr/cube/matrix/assembler.py
```diff
@@ -1161,7 +1175,29 @@ def _subtotal_values(self):
         return intersections[:, self._insertion_idx]
 
 
-class _SortRowsByMarginalHelper(_RowOrderHelper):
+class _SortRowsByLabelHelper(_SortRowsBaseHelper):
```
scanny, line 1178: I think this looks much cleaner now. Basically every different type needs to compute element-values and subtotal-values, and everything else is handled but its superclasses. This feels even cleaner than I thought. Makes me want to invent a new sort-by-value type just to take advantage of that :)

src/cr/cube/stripe/assembler.py
```diff
@@ -367,8 +368,15 @@ def _display_order(self):
         return CollatorCls.display_order(self._rows_dimension, self._empty_row_idxs)
 
 
-class _SortByMeasureHelper(_BaseOrderHelper):
-    """Orders rows by the value of a specified measure."""
+class _SortBaseHelper(_BaseOrderHelper):
```
scanny, line 371: I'd say `_BaseSortByValueHelper`.

scanny (approved):
k, couple small proposed changes, good to go after those, approving in advance.

src/cr/cube/matrix/assembler.py
gergness: The customers are clearly asking for a `_SortRowsByRandomValueHelper` 😉

PR #339, opened 2021-08-02
Shim element ids in dimension dictionaries and allow sort by MR insertion

gergness:


src/cr/cube/dimension.py
```diff
@@ -525,6 +559,135 @@ def valid_elements(self):
...
+        # --- Translate explicit order element ids if present
+        element_ids = self._dimension_transforms_dict.get("order", {}).get(
+            "element_ids"
+        )
+        if element_ids:
+            shim["order"]["element_ids"] = [
+                self.translate_element_id(_id) for _id in element_ids
+            ]
+
+        # --- sort-by-value on the opposing dimension also refers to element ids, but
```
gergness, line 683: Not sure if this is the right approach, but having to look up opposing dimension here seemed difficult

tests/fixtures/mr_insertions/cat-x-mr.json
```diff
@@ -306,78 +306,78 @@
         "measures": {
             "count": {
                 "data": [
-                    1.5,
```
gergness, line 309: These changes are because on the last PR you mentioned that the test was a little hard to interpret because there were so many rows of all 0's. Rather than add a new fixture, I changed this one to have a wider variety of numbers.

tests/fixtures/mr_insertions/cat-x-mr.json
```diff
@@ -166,7 +166,7 @@
                             "value": {
                                 "derived": true,
                                 "references": {
-                                    "alias": "A&B",
+                                    "alias": "A_B",
```
gergness, line 169: Changed so that alias is not the same as name

scanny:
@gergness This looks like the right idea on first glance, I wasn't able to give it quality time today but will look at it more closely tomorrow afternoon after I get back from my appointment.

src/cr/cube/dimension.py
```diff
@@ -725,6 +734,214 @@ def _elements(self):
...
+    - "element_id": Stored in the cube result as the object name in
+        `dimensions[i].type.elements[j].id`. For subvariables, zz9 currently puts
+        the index integer here. Long term zz9 may change this to the the alias.
+    - "subvariable_id": Subvariables have an id stored in
+        `dimensions[i].type.elements[j].value.id`, generally this is a 4 digit,
+        0-padded index of the subvariable when it was first created (eg "0001",
+        "0002", ...), though it is not required to be. For derived insertions,
+        currently the name is used here.
+    - "alias": Subvariables also have an alias that identifies them. It
+    """
```
scanny, line 755: Couple formatting changes.

* We put a period after the summary-line of the docstring. It's not always a whole sentence, but it's just what we've come to do.
* There should be a blank like before the first bullet. The text in a docstring is restructuredText interpreted by Sphinx to generate the documentation and it won't know what to do without a space before the bulleted list.
* The default first-level bullet character for restructuredText is an asterisk ("*"), so probably doesn't hurt to stick with that for consistency.
* Second and later lines of a bullet-point should line up under the first character of the first line of the bullet point, like:

  ```rst
  - "element_id": Stored in the cube result as the object name in
    `dimensions[i].type.elements[j].id`. For subvariables, zz9 currently puts
    the index integer here. Long term zz9 may change this to the the alias.
  ```

  I'm not sure how it gets formatted if it doesn't align, but I'm pretty sure it's not what we would expect or want.

src/cr/cube/dimension.py
```diff
@@ -725,6 +734,214 @@ def _elements(self):
...
+        self._dimension_dict = dimension_dict
+        self._dimension_transforms_dict = dimension_transforms_dict
+
+    @lazyproperty
+    def shimmed_dimension_dict(self):
+        """Copy of dimension dictionary with shimmed `element_id`s
+
+        We want to move to a world where elements on a subvariables dimension are
+        identified by their alias, but right now the "element_id" from zz9 is
+        an index for subvariables.
```
scanny, line 768: I keep catching on "is an index". I guess it's not a term that commonly has meaning to me when appearing on it's own. Maybe `(unstable) element-index` or `is simply the subvariable's (unstable) cardinal position in the subvariable sequence`.

src/cr/cube/dimension.py
```diff
@@ -725,6 +734,214 @@ def _elements(self):
...
+        """Replace the list list of element ids with a list of aliases
+
+        The explicit order transform includes a list of ids that can be specified in
+        many different ways, this translate them to the subvariable aliases.
+        """
+        return [self.translate_element_id(_id) for _id in element_ids]
+
+    @lazyproperty
+    def _subvar_aliases(self):
+        """Tuple of string aliases for each element of a subvariable dimension
```
scanny, line 926: Just a note on this style of summary string. It is `{sequence-type} of {item-type} {singular-item-name} for each {one-to-one-corresponding-thing}`.

So this would be `tuple of str alias for each element of a subvariable dimension.`

* `tuple` is a type and so is spelled as its type, even though it's the first word in the "sentence".
* `str` is a type, so we are as explicit as we can be, `str` here instead of `string`.
* `alias` is singular, rather than aliases.

You could think of it as: "tuple consisting of the str alias for each element of ...". We just leave out the consisting of and preposition "the" for brevity.

It's a fine point, but that's how it got the way you see it around, even though it takes a second sometimes to read it as intended.

src/cr/cube/dimension.py
```diff
@@ -725,6 +734,214 @@ def _elements(self):
...
+        fixture, but I don't think it can happen in the real world anymore.)
+        """
+        return tuple(
+            element.get("value", {}).get("references", {}).get("alias", element["id"])
+            for element in self._dimension_dict["type"]["elements"]
+        )
+
+    @lazyproperty
+    def _subvar_ids(self):
+        """Tuple of string subvariable ids for each element of a subvariable dimension"""
```
scanny, line 938: same on this one.

Also we should note in the docstring that referring to this property is only valid on a subvars dimension. This will raise `KeyError` on a CAT dimension I'm pretty sure.

scanny (approved):
This looks real good @gergness. A few nits, but good to go after that, approving in advance.

Nice job :)

src/cr/cube/dimension.py
```diff
@@ -507,14 +506,28 @@ def subtotals(self):
...
+        0) If dimension is not a subvariables dimension, return the _id.
+        1) If id matches an alias, then just use it.
+        2) If id matches a subvariable id, translate to corresponding alias.
+        3) If id matches an element id, translate to corresponding alias.
+        4) If id can be parsed to int and matches an element id, translate to alias.
+        5) If id is int (or can be parsed to int) and can be used as index (eg in range
+           0-# of elements), use _id'th alias.
+        6) If all of these fail, return None.
+        """
+        return self._element_id_shim.translate_element_id(_id)
```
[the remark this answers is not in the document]
gergness: I'm not super happy about it, but it's for the opposing element sort by value transform. This shimmed dimension transform dictionary does not have a translated element id here, because to translate it would require the dimension to understand it's opposing dimension at creation time, which seemed hard to manage.

Added the following to the docstring.

>        This is needed for the opposing dimension's sort by opposing element, because
>        when creating a dimension, we don't have access to the other dimension's 
>        ids to transform it. Therefore, the id for opposing element sort by value
>        transforms is not translated at creation time.

PR #342, opened 2021-08-18
Drop python2 [#179262966]

ernestoarbitrio:


src/cr/cube/util.py
```diff
@@ -54,65 +42,4 @@ def fget(self):
...
-        is, there would be nothing to stop assignment to the cached value,
-        which would overwrite the result of `fget()` and break both the
-        immutability and idempotence guarantees of this decorator.
-
-        The performance with this __set__() method in place was roughly 0.4
-        usec per access when measured on a 2.8GHz development machine; so
-        quite snappy and probably not a rich target for optimization efforts.
-        """
-        raise AttributeError("can't set attribute")
+    return property(functools.lru_cache()(f))
```
scanny, line 45: I'm open to arguments on how this can be improved, but I don't think this is it. Happy to talk it over with you but in the meantime I think we should change this back to the way it was.

src/cr/cube/collator.py
```diff
@@ -10,15 +10,16 @@
 base vectors and inserted vectors, respectively.
 """
 
-import collections
 import sys
 
 import numpy as np
 
+from collections import OrderedDict
```
scanny, line 17: I'm sure this is purely subjective, but I prefer to always import Python library modules whole and then refer to members with qualification, like `collections.OrderedDict`, so this one was made this way on purpose.

scanny (changes requested):
Overall I like the main bits like getting rid of `__future__` imports and `(object)` on classes where it's not strictly necessary. I think I'll get used to the f-strings, we'll see.

The lazyproperty change is not an improvement as far as I can tell, although I'm open to hearing why you thought it should be changed. I think we're much safer for now leaving that the way it is and discussing any changes prior to implementing them.

Otherwise, looks good to me.

slobodan-ilic (approved):

PR #344, opened 2021-09-01
Mr insertion explicit order 179379022

gergness:
This ended up being slightly bigger than expected because MR insertions can be anchored before or after elements (cat subtotals can only be anchored after).

The other bit of weirdness is the way that we use negative indexes to indicate categorical subtotals. Derived MR subtotals do not have negative indexes like the categorical ones, so the index has to be created from both base and derived elements.


scanny (approved):
One typo, otherwise LGTM

PR #346, opened 2021-09-06
Add mypi to Crunch Cube CI

slobodan-ilic:


src/cr/cube/cube.py
```diff
@@ -267,24 +281,24 @@ def covariance(self):
...
 
     @lazyproperty
-    def cube_index(self):
+    def cube_index(self) -> int:
         """Offset of this cube within its CubeSet."""
         return 0 if self._cube_idx_arg is None else self._cube_idx_arg
 
     @lazyproperty
-    def description(self):
+    def description(self) -> "Optional[str]":
```
ernestoarbitrio, line 289: Why this is between `"` quotes?

src/cr/cube/cube.py
```diff
@@ -7,15 +7,18 @@
...
 
 import numpy as np
 
 from cr.cube.cubepart import CubePartition
-from cr.cube.dimension import AllDimensions
-from cr.cube.enums import CUBE_MEASURE, DIMENSION_TYPE as DT, NUMERIC_MEASURES
+from cr.cube.dimension import AllDimensions, _ApparentDimensions
+from cr.cube.enums import CUBE_MEASURE
+from cr.cube.enums import DIMENSION_TYPE as DT
+from cr.cube.enums import NUMERIC_MEASURES
```
ernestoarbitrio, line 18: ?? isort? I do prefer the old way though

src/cr/cube/cube.py
```diff
@@ -428,7 +442,7 @@ def title(self):
         return self._cube_dict["result"].get("title", "Untitled")
 
     @lazyproperty
-    def unweighted_counts(self):
+    def unweighted_counts(self) -> "np.ndarray[np.int64]":
```
ernestoarbitrio, line 445: also here why type is in quotes?

ernestoarbitrio (changes requested):
Some of the return types seems weird to me ... I just left some comments

src/cr/cube/cube.py
slobodan-ilic: fixed

src/cr/cube/cube.py
slobodan-ilic: because `np.ndarray[np.int64]` fails with the error about the `type` can't be initialized or something...

src/cr/cube/cube.py
slobodan-ilic: fixed

src/cr/cube/cube.py
```diff
@@ -1028,7 +1044,7 @@ class _UnweightedCountMeasure(_BaseMeasure):
     """Unweighted counts for cube."""
 
     @lazyproperty
-    def _flat_values(self):
+    def _flat_values(self) -> "np.ndarray":
```
ernestoarbitrio, line 1047: why `"` quotes here?

ernestoarbitrio (changes requested):
some small comments

src/cr/cube/cube.py
slobodan-ilic: fixed

tests/unit/test_dimension.py
```diff
@@ -442,10 +442,12 @@ def it_resolves_a_categorical_type_to_help(
         is_logical_type,
         expected_value,
         _is_array_cat_prop_,
+        _is_cat_date_prop_,
         _has_selected_category_prop_,
         _is_logical_type_prop_,
     ):
         _is_array_cat_prop_.return_value = is_array_cat
+        _is_cat_date_prop_.return_value = False
```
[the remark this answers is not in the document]
slobodan-ilic: Because I don't want it to enter the function code, because then it'll crash. The crash is because the iterable is initialized to `None` from the unit test. But we don't need to cover that case in the real world (iterable being set to `None`), so I've taken that part of the code out. And now it's setting the unit test in a more realistic way. Here's the specific change that this allows:

https://github.com/Crunch-io/crunch-cube/pull/346/files#diff-84a1fae6b8f21e7c557dfbf8ea91f3306a10a3cb9e704b279f29d3ab6d915964L254

src/cr/cube/collator.py
```diff
@@ -12,9 +12,11 @@
 
 import collections
 import sys
+from typing import Dict, FrozenSet, Iterator, Tuple, Union, List
```
ernestoarbitrio, line 15: isort should order import in A-Z isn't it?

src/cr/cube/cube.py
```diff
@@ -7,15 +7,16 @@
 
 import copy
 import json
+from typing import FrozenSet, Iterator, List, Optional, Tuple, Union, Dict
```
ernestoarbitrio, line 10: same here

src/cr/cube/cube.py
```diff
@@ -131,20 +138,20 @@ def population_fraction(self):
...
         """The valid count summary values from first cube in this set."""
         return self._cubes[0].valid_counts_summary
 
     @lazyproperty
-    def _cubes(self):
+    def _cubes(self) -> Tuple["Cube", ...]:
         """Sequence of Cube objects containing data for this analysis."""
 
-        def iter_cubes():
+        def iter_cubes() -> Iterator[Cube]:
```
ernestoarbitrio, line 154: I'm wondering why here you're using `Cube` and 3 lines above `"Cube"`

src/cr/cube/dimension.py
```diff
@@ -132,19 +133,19 @@ def __init__(self, dimension_dicts):
         self._dimension_dicts = dimension_dicts
 
     @classmethod
-    def iter_dimensions(cls, dimension_dicts):
+    def iter_dimensions(cls, dimension_dicts) -> Iterator["Dimension"]:
```
ernestoarbitrio, line 136: Why this is `Iterator` and not `Generator`?

src/cr/cube/dimension.py
```diff
@@ -132,19 +133,19 @@ def __init__(self, dimension_dicts):
...
         """Generate Dimension object for each of *dimension_dicts*."""
         return cls(dimension_dicts)._iter_dimensions()
 
-    def _iter_dimensions(self):
+    def _iter_dimensions(self) -> Iterator["Dimension"]:
```
ernestoarbitrio, line 140: same

src/cr/cube/dimension.py
```diff
@@ -578,15 +574,15 @@ def __len__(self):
         return len(self._elements)
 
     @lazyproperty
-    def element_ids(self):
+    def element_ids(self) -> Tuple[int, ...]:
```
ernestoarbitrio, line 577: is element_id always an `int`?

ernestoarbitrio (changes requested):
I think after this it should be good to go 😉

src/cr/cube/cube.py
slobodan-ilic: Because this one is defined inside of the function, which means each time the parent function is called, so it's aware of the definition of the `Cube` class. The parent function, on the other hand, is defined when the file is loaded for the first time, and is not aware of the `Cube` class at the time of the definition.

src/cr/cube/collator.py
slobodan-ilic: fixed

src/cr/cube/cube.py
slobodan-ilic: fixed

src/cr/cube/dimension.py
slobodan-ilic: Yes, because:
```    def element_id(self) -> int:
```

ernestoarbitrio (approved):

PR #347, opened 2021-09-14
[179448418]: add metadata public property needed for csv export

ernestoarbitrio:


src/cr/cube/cubepart.py
```diff
@@ -268,6 +268,16 @@ def __repr__(self):
 
     # ---interface ---------------------------------------------------
 
+    @lazyproperty
```
slobodan-ilic, line 271: Can you do the type annotation for mypy while you're at it? Only for your new work I mean...

src/cr/cube/dimension.py
```diff
@@ -1019,58 +1040,62 @@ def label(self) -> str:
...
+        """
+        return bool(self._element_dict.get("missing"))
+
+    @lazyproperty
+    def numeric_value(self) -> Union[int, float]:
+        """Numeric value assigned to element by user, np.nan if absent."""
+        numeric_value = self._element_dict.get("numeric_value")
+        return np.nan if numeric_value is None else numeric_value
+
+    def _retreive_value(self, key: str) -> str:
```
slobodan-ilic, line 1061: * retrieve

src/cr/cube/cubepart.py
ernestoarbitrio: sure

src/cr/cube/dimension.py
slobodan-ilic: `_retrieve_value` is not very informative. I'd prefer a different name. Like, what are we actually trying to do here? It used to be a `label`, but it was too complicated for just the label. So it's more like "textual representation of an element". So maybe something like `_as_string()`, or something similar. Or `stringify`, or something like that. I know these are not perfect, but it tells us that the intended value is `str`. Or maybe, if you wanted to be as explicit as possible, then you'd do something like `get_str_representation_for(key="name")`...

slobodan-ilic (approved):

src/cr/cube/dimension.py
ernestoarbitrio: yeah seems a very good point

PR #348, opened 2021-09-16
Indexes of differences

slobodan-ilic:
* Drive with TDD
* implement functionality similarly to other elements idxs

### DIFF
--- a/src/cr/cube/matrix/assembler.py
+++ b/src/cr/cube/matrix/assembler.py
@@ -248,6 +248,16 @@ class Assembler:
         return self._derived_element_idxs(self._rows_dimension, self._row_order)
 
     @lazyproperty
+    def diff_row_idxs(self):
+        """tuple(int) of difference row elements' indexes, can be empty."""
+        return self._diff_element_idxs(self._rows_dimension, self._row_order)
+
+    @lazyproperty
+    def diff_column_idxs(self):
+        """tuple(int) of difference column elements' indexes, can be empty."""
+        return self._diff_element_idxs(self._columns_dimension, self._column_order)
+
+    @lazyproperty
     def inserted_column_idxs(self):
         """tuple of int index of each subtotal column in slice."""
         # --- insertions have a negative idx in their order sequence ---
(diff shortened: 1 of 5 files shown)


src/cr/cube/matrix/assembler.py
```diff
@@ -821,6 +831,20 @@ def _derived_element_idxs(self, dimension, order):
...
+            element_index
+            for element_index, is_difference in enumerate(
+                np.array(
+                    # ---Valid elements are not differences
+                    [False for _ in dimension.valid_elements]
+                    + [element.is_difference for element in dimension.subtotals]
+                )[order]
+            )
+            if is_difference
+        )
```
ernestoarbitrio, line 846: IMHO i think that this logic should be divided into who belongs in the dimension and who in the assembler. I would propose that this property becomes:
```python
    def _diff_element_idxs(self, dimension, order):
        """Return tuple(int) of difference elements' indices for a dimension."""
        return tuple(np.where(dimension.subdiff_map[order])[0])
```

Where `subdiff_map` is this property within the `Dimension` class in the `dimension.py` module

```python
    @lazyproperty
    def subdiff_map(self) -> np.ndarray:
        return np.array(
            # ---Valid elements are not differences
            [False for _ in self.valid_elements]
            + [subtotal.is_difference for subtotal in self.subtotals]
        )
```
What do you think?

src/cr/cube/stripe/assembler.py
```diff
@@ -63,6 +63,21 @@ def derived_row_idxs(self):
...
+            for element_idx, is_difference in enumerate(
+                np.array(
+                    # ---Valid elements are not differences
+                    [False for _ in self._rows_dimension.valid_elements]
+                    + [e.is_difference for e in self._rows_dimension.subtotals]
+                )[self._row_order]
+            )
+            if is_difference
+        )
+
```
ernestoarbitrio, line 80: same here

ernestoarbitrio (changes requested):
I think that `_diff_element_idxs` can be refactored a bit moving the logic within the dimension. Take a look at the comment in the sections :D

src/cr/cube/matrix/assembler.py
slobodan-ilic: Well, nobody is going to use the `subdiff_map`, other than this property. So I'd say that it's unearned abstraction. I do get your point, but I'd still keep it this way. Maybe I can simplify the expression a little bit.

ernestoarbitrio: ok maybe you can use the `np.where` instead of the external iteration

ernestoarbitrio (approved):

PR #350, opened 2021-09-24
Subtotal fills

slobodan-ilic:
* Drive by TDD
* Add `fill` prop to `_Subtotal`
* Add interleaved subtotal fills to row dimension fills in assemblers

### DIFF
--- a/tests/unit/stripe/test_assembler.py
+++ b/tests/unit/stripe/test_assembler.py
@@ -6,38 +6,33 @@ import numpy as np
 import pytest
 
 from cr.cube.cube import Cube
-from cr.cube.dimension import Dimension, _Element, _OrderSpec
+from cr.cube.dimension import Dimension, _Element, _OrderSpec, _Subtotal
 from cr.cube.enums import COLLATION_METHOD as CM
 from cr.cube.stripe.assembler import (
+    StripeAssembler,
     _BaseOrderHelper,
     _BaseSortByValueHelper,
     _OrderHelper,
     _SortByLabelHelper,
     _SortByMeasureHelper,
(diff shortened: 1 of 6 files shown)


tests/unit/stripe/test_assembler.py
```diff
@@ -108,15 +103,18 @@ def it_knows_the_rows_dimension_fills(
...
         assembler = StripeAssembler(None, rows_dimension_, None, None)
 
         print(assembler.rows_dimension_fills)
         assert assembler.rows_dimension_fills == (
             "012345",
-            None,
+            "STF_2",
             "6789ab",
-            None,
+            "STF_1",
```
ernestoarbitrio, line 117: I'm not sure this is correct. if the subtotals fills are ["STF_1", "STF_2"] and the idx is -2 why the second element of the expectation is `_2` and not `_1`, maybe I'm wrong but this seems weird to me.

ernestoarbitrio (commented):
I do need to solve my concern :D

tests/unit/stripe/test_assembler.py
slobodan-ilic: yeah, I think it's actually correct... because the indexes are assigned starting from the negative value that's closest to zero, and then decreasing... so the payload order would be -1, -2... and then here, since they're flipped explicitly, with reordering transform, they come the other way round...

ernestoarbitrio: maybe I'm doing confusion ... but does `-1` mean the last item? like the python indexing starting from the bottom?

ernestoarbitrio (approved):

PR #354, opened 2021-10-06
add upload dist action

ernestoarbitrio:


.github/workflows/release.yml
```diff
@@ -0,0 +1,54 @@
...
+        run: twine upload --repository testpypi dist/*
+      - name: Publish on PYPI
+        env:
+          TWINE_USERNAME: __token__
+          TWINE_PASSWORD: ${{ secrets.PYPI_APIKEY }}
+        run: twine upload dist/*
+      - name: Build wheels and upload to Nexus
+        run: |
+          VERSION=$(echo $GITHUB_REF | sed 's#.*/v##')
+          echo -e "numpy==1.16.2\ncr.crube==${VERSION}" > requirements-whl.txt
```
arlind87, line 46: should this be copied over to requirements.txt ?

arlind87: I do not think we use `requirements-whl.txt` in here

ernestoarbitrio: Considering that we do not use a `requirements.txt` file I'm gonna get rid of that line and use an alternative that work fine as well. Tested on my local env, it produces the same whl versions. Let me know what do you think

PR #358, opened 2022-08-15
[182979152]: fix bug for hiding nonexistant mr insertion

gergness:
If an MR insertion doesn't exist by name, don't error, just don't hide it. Can happen in fused scorecards.


ernestoarbitrio (approved):
LGTM, but I think you have to re-format a file cause black is complaining about something. The thing is that I updated the pre-commit and the CI job for that, so please be sure you're using pre-commit locally and re-commit the changes to make the CI passes :D

PR #359, opened 2022-09-01
Expose new properties needed for color palette fix in pptx exporter 183094432

ernestoarbitrio:


gergness (approved):
Looks good to me, only suggestion is to elaborate the docstrings in `cubepart`?

I think it needs to at least mention something like:

```
Positive integers indicate the 1-indexed position in payload of regular elements, while negative integers are the subtotal insertions.
```

And maybe also explain why we need them?

PR #362, opened 2022-11-22
[#183769192]: Add `repr` to various sequences

slobodan-ilic:
Add `repr` to various sequence classes, to ease the work in console. For example, if we're dealing with a dimension class, it'd be nice to be able to "just" see its type and elements, without having to resource to its fields/properties. This is especially useful for sequence classes that use generators and don't print out their contents by default.

### DIFF
--- a/src/cr/cube/dimension.py
+++ b/src/cr/cube/dimension.py
@@ -33,6 +33,11 @@ class _BaseDimensions(Sequence):
         """Implements len(elements)."""
         return len(self._dimensions)
 
+    def __repr__(self) -> str:
+        "str representation of this dimension sequence." ""
+        dim_lines = "\n".join(f"{d}" for d in self._dimensions)
+        return f"Dimensions:\n{dim_lines}"
+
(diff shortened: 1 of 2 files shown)


src/cr/cube/dimension.py
```diff
@@ -330,6 +335,10 @@ def __init__(self, dimension_dict, dimension_type, dimension_transforms=None):
         self._dimension_type = dimension_type
         self._unshimmed_dimension_transforms_dict = dimension_transforms or {}
 
+    def __repr__(self) -> str:
+        """str representation of this dimension."""
+        return f"{self.name} ({self.dimension_type}): {self.valid_elements}"
```
ernestoarbitrio, line 340: I would not put the `valid_elements` here in the `__repre__` for 1 reason:
Real cubes have tons of valid elements especially MR and this repr could be not readable.

IMHO this should be only `name` and `dim_type`

ernestoarbitrio (approved):
only 1 comment

PR #363, opened 2022-12-08
[183976784]: allow scorecard xforms

gergness:
The shim layer of element ids gets confused by transforms on a scorecard's fused variables dimension

### DIFF
--- a/tests/fixtures/scorecard.json
+++ b/tests/fixtures/scorecard.json
@@ -1,0 +1,333 @@
+{
+  "result": {
+    "n": 210,
+    "counts": [
+      40,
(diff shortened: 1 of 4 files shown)


tests/fixtures/scorecard.json
```diff
@@ -0,0 +1,333 @@
+{
```
gergness, line 1: This file isn't strictly necessary, as there is already a scorecard fixture, but it has insertions and the interaction of transforms on a scorecard is even less well defined than without them.

ernestoarbitrio (approved):

PR #365, opened 2022-12-20
add subtotals in payload order property in Dimension

ernestoarbitrio:


src/cr/cube/dimension.py
```diff
@@ -530,8 +530,36 @@ def subtotals(self) -> "_Subtotals":
...
+            original paylod anchor.
+            """
+            name = insertion.get("name")
+            occurence = [el for el in view_insertions if el["name"] == name]
+            if occurence:
+                insertion["anchor"] = occurence[0]["anchor"]
+            return insertion
+
+        if self.dimension_type in (DT.MR, DT.CA_SUBVAR):
+            insertion_dicts: List[Optional[Dict]] = []
```
gergness, line 556: For my curiosity, why do you need this type annotation? `: List[Optional[Dict]]`

tests/unit/test_dimension.py
```diff
@@ -724,6 +724,76 @@ def and_it_overrides_with_subtotal_transforms_when_present(
...
+                    {"anchor": "bottom", "name": "b", "hide": True},
+                ],
+            ),
+            (
+                DT.CAT,
+                [{"anchor": "top", "name": "a"}],
+                {},
+                [{"anchor": "top", "name": "a"}],
+            ),
+        ),
```
gergness, line 761: I think your code fails in the case of two insertions anchored to the same thing but with a flipped order between variable view & analysis. This is important because it their order breaks the tie when they are anchored to the same thing.

Here's another test case, can you add?
```python
(
  DT.CAT,
  [{"anchor": 1, "name": "a"}, {"anchor": 1, "name": "b"}],
  [{"anchor": 1, "name": "b"}, {"anchor": 1, "name": "a"}],
  [{"anchor": 1, "name": "a"}, {"anchor": 1, "name": "b"}],
)
```

gergness (changes requested):
I think I see a problem with insertions anchored to the same thing.

src/cr/cube/dimension.py
ernestoarbitrio: mypy complains about it ... dunno why honestly. Probably latest version of mypy

src/cr/cube/dimension.py
```diff
@@ -530,8 +530,36 @@ def subtotals(self) -> "_Subtotals":
...
+        its addends and anchor.
+        """
+
+        def apply_order(insertion, view_insertions):
+            """
+            Change the anchor of the dimension transforms insertions with the
+            original paylod anchor.
+            """
+            name = insertion.get("name")
+            occurence = [el for el in view_insertions if el["name"] == name]
```
gergness, line 550: Oh shoot, also this will have problems if they rename the subtotal. The frontend uses id, and if there's no id they use the index starting at 1.

ernestoarbitrio: so on what we should rely on?

gergness: Use id if present, and the 1 based index if not. I believe it’s always present on the analysis insertions, so you don’t have to worry about reordering affecting the ids

src/cr/cube/dimension.py
```diff
@@ -614,7 +614,12 @@ def _dimension_transforms_dict(self) -> Dict:
     def _view_insertion_dicts(self) -> List[Optional[Dict]]:
         """List of insertion dicts included in the dimension view."""
         view = self._dimension_dict.get("references", {}).get("view") or {}
-        return view.get("transform", {}).get("insertions", [])
+        insertions = view.get("transform", {}).get("insertions", [])
+        view_insertions = [
```
gergness, line 618: Can you add a comment like `If there's no id defined on the insertion defined on the variable view, use the 1-based positional index as the id to match frontend behavior`

gergness (approved):
LGTM (but think we should add a comment)

PR #369, opened 2023-02-16
Dashboard datetime xform 184468428

gergness:
Makes datetimes shimmed like subvariables are so that cr.cube can understand both deck and dashboard formats of transforms for datetimes. (this way we don't have to shim in the "dashboard shim")

### DIFF
--- a/src/cr/cube/dimension.py
+++ b/src/cr/cube/dimension.py
@@ -931,10 +931,7 @@ class _ElementIdShim:
             return _id
 
         if self._dimension_type == DT.DATETIME:
-            try:
-                int_id = int(_id)
-            except ValueError:
-                int_id = _id
+            int_id = int(_id) if isinstance(_id, str) and _id.isnumeric() else _id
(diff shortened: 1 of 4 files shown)


src/cr/cube/dimension.py
```diff
@@ -922,9 +927,17 @@ def translate_element_id(self, _id) -> Optional[str]:
...
         """
-        if self._dimension_type not in DT.ARRAY_TYPES:
+        if self._dimension_type not in DT.SHIMMED_TYPES:
             return _id
 
+        if self._dimension_type == DT.DATETIME:
+            try:
+                int_id = int(_id)
+            except ValueError:
+                int_id = _id
```
ernestoarbitrio, line 937: is this try block used to check is `_id` is a str representation for a numeric value? If yes please use `str.isnumeric()` instead of using that try block

ernestoarbitrio (approved):
LGTM but please check my comment

PR #374, opened 2023-03-23
Datetime label format 184749447

gergness:
Allows exporter to reflect the date format requested on the variable's view instead of using the date format from zz9 on dimensions.

### DIFF
--- a/src/cr/cube/dimension.py
+++ b/src/cr/cube/dimension.py
@@ -4,7 +4,8 @@
 
 import copy
 from collections.abc import Sequence
-from typing import Dict, Iterator, List, Optional, Tuple, Union
+from datetime import datetime
+from typing import Any, Dict, Iterator, List, Optional, Tuple, Union
 
 import numpy as np
 
@@ -354,6 +355,7 @@ class Dimension:
             self._dimension_dict["type"],
             self._dimension_transforms_dict,
             self._dimension_type,
+            self._format_element_label,
         )
 
     def apply_transforms(self, dimension_transforms) -> "Dimension":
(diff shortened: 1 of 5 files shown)


src/cr/cube/dimension.py
```diff
@@ -610,6 +612,50 @@ def _dimension_transforms_dict(self) -> Dict:
...
+            "3M": "%Y-%m",
+            "M": "%Y-%m",
+            "W": "%Y-%m-%d",
+            "D": "%Y-%m-%d",
+            "h": "%Y-%m-%dT%H",
+            "m": "%Y-%m-%dT%H:%M",
+            "s": "%Y-%m-%dT%H:%M:%S",
+            "ms": "%Y-%m-%dT%H:%M:%S.%f",
+            "us": "%Y-%m-%dT%H:%M:%S.%f",
+        }.get(resolution)
```
ernestoarbitrio, line 657: This supposes that resolution always exist in case of datetime. Is that true or do we need a fallback option in case of None value?

src/cr/cube/dimension.py
```diff
@@ -610,6 +612,50 @@ def _dimension_transforms_dict(self) -> Dict:
...
+        if self.dimension_type == DT.DATETIME:
+            return self._format_datetime_element_label(x)
+        else:  # --- TODO: Should round numeric values like the frontend does
+            return str(x)
+
+    @lazyproperty
+    def _incoming_element_datetime_format(self) -> Optional[str]:
+        """optional str of date formatting requested if datetime and is found"""
+        if self.dimension_type != DT.DATETIME:
+            return None
```
ernestoarbitrio, line 642: is there any reason why all this methods/properies belong to the `_Dimension` class? I'm for having this in the `_AllElements` maybe or in the `_Element` itself. Probably passing the `_element_data_format` to the `_AllElements` is all you need to move all this logic there. Is there any impediment that I don't see? What do you think?

src/cr/cube/dimension.py
gergness: `{...}.get(None)` is None, right? Do you want some way to make it more clear?

src/cr/cube/dimension.py
gergness: Yeah, originally wanted it in `_Element`, but then I'd have to pass `dim_type`, `incoming_format` and `desired_format` into each element. Maybe in `_AllElements` is better because the only thing new would be `incoming_format`

src/cr/cube/dimension.py
ernestoarbitrio: my mistake ... that's fine

ernestoarbitrio (approved):

PR #378, opened 2023-04-05
[#184749279]: Bug fix hidden subvars in MR with insertions

ernestoarbitrio:


src/cr/cube/dimension.py
```diff
@@ -982,6 +982,35 @@ def translate_element_id(self, _id) -> Optional[str]:
...
+        #                       'enjoy_mr_sweet__4')
+        # --- Relying on the subvar_ids this will get the self._subvar_ids.index(_id)
+        # --- that corresponds to 3 and so subvar_aliases[3] == enjoy_mr_spicy__4.
+        # --- This means that we're hiding the col Spicy rather than Savory.
+        # --- Using the _raw_element_ids instead will return the correct hidden alias.
+        # --- NOTE: this goes away when we will use aliases rather than id.
+        if self._has_mr_insertion:
+            elements = self._dimension_dict["type"]["elements"]
+            insertions_mask = [
+                1 if el["value"].get("references", {}).get("anchor") else 0
```
slobodan-ilic, line 1004: I'd rather use the construction `"anchor" in obj` than `obj.get("anchor")`. The first one is more explicit in that you don't care about the contents of the anchor, but just that it's present (because it defines the insertion).

tests/integration/test_multiple_response.py
```diff
@@ -1745,3 +1745,71 @@ def test_mr_x_mr_min_base_size_mask():
...
+                },
+            ],
+        },
+        "rows_dimension": {
+            "order": {"element_ids": [1, 2, 3, 4], "type": "explicit"},
+        },
+    }
+    for idx in hidden_insertions:
+        transforms["columns_dimension"]["insertions"][idx].update({"hide": True})
+    slice_ = Cube(MRI.CAT_X_MR_HS, transforms=transforms).partitions[0]
```
slobodan-ilic, line 1813: Maybe, just maybe, consider the name of this fixture file. The MR insertions are not really H&S. But we'll probably get right away what it is in the context of MR as well. I'm thinking of our future selves reading these tests. Maybe, `CAT_X_MR_W_INS` would work better? But I'm not 100% sure, if we already have MR_HS cases in the code then it's completely fine :)

slobodan-ilic (approved):

src/cr/cube/dimension.py
ernestoarbitrio: right thx

PR #381, opened 2023-04-12
fix subtotals in payload order to consider different ids when specifi…

ernestoarbitrio:
…ed in transforms

### DIFF
--- a/src/cr/cube/dimension.py
+++ b/src/cr/cube/dimension.py
@@ -5,7 +5,7 @@
 import copy
 from collections.abc import Sequence
 from datetime import datetime
-from typing import Any, Dict, Iterator, List, Optional, Tuple, Union
+from typing import Any, Dict, Iterator, List, Optional, Tuple, Union, cast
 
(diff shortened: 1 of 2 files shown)


src/cr/cube/dimension.py
```diff
@@ -546,9 +536,23 @@ def subtotals_in_payload_order(self) -> "_Subtotals":
...
+            insertion_dicts = copy.deepcopy(self._view_insertion_dicts)
+            for ins in insertion_dicts:
+                ins_dict: dict = cast(dict, ins)
+                original_id = ins_dict["id"]
+                # adapt view insertions with ids of the final transform insertions
+                insertions_map = {
+                    insertion["name"]: insertion["id"]
+                    for insertion in self._insertion_dicts
+                }
+                if ins_dict["name"] in insertions_map:
```
gergness, line 548: Does this work if the subtotal name is changed in the transforms?

gergness (changes requested):

PR #382, opened 2023-04-12
Another attempt at fixing color when insertions are in weird order

gergness:
This has a lot of failing unit tests, and I haven't checked that it actually fixes the issue (but it basically undoes what I changed in the shim, so I think it should). Wanted to see what you thought before cleaning it up.

### DIFF
--- a/src/cr/cube/dimension.py
+++ b/src/cr/cube/dimension.py
@@ -630,13 +630,7 @@ class Dimension:
         """List of insertion dicts included in the dimension view."""
         view = self._dimension_dict.get("references", {}).get("view") or {}
         insertions = view.get("transform", {}).get("insertions", [])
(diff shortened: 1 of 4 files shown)


src/cr/cube/dimension.py
```diff
@@ -1618,7 +1678,7 @@ def fill(self) -> str:
     @lazyproperty
     def insertion_id(self) -> int:
         """int unique identifier of this subtotal within this dimension's insertions."""
-        return self._subtotal_dict.get("id", self._fallback_insertion_id)
+        return self._subtotal_dict["id"]
```
ernestoarbitrio, line 1681: is there any remote change that a subtotal dict doesn't have an `id`?

ernestoarbitrio: I guess we're safe. A took a look at the code and it should be ok

PR #385, opened 2023-07-25
Simplify

aminusfu:
crunch-cube is more complicated than it needs to be. For example, dimension.py has 24 cooperating classes (!). This makes it hard to read, debug, and improve. In addition, the practice of putting each statement in its own function makes it hard to improve or adapt to changes in the cube output, like the categorical "order" field that prompted this effort, or the future schema/metadata revamp of variable definitions.


src/cr/cube/cubepart.py
```diff
@@ -561,7 +609,9 @@ def columns_scale_median_margin(self):
     @lazyproperty
     def counts(self):
         """2D np.float64 ndarray of weighted cube counts."""
-        return self._assembler.weighted_counts
+        return self._assemble_matrix(self._measures.weighted_counts.blocks)
+
+    weighted_counts = counts
```
gergness, line 614: I don't know python evaluation well enough to know: Does this remove the "lazyness"? so that counts are always calculated because you assing weighted_counts = counts?

If so do we care enough?

tests/unit/matrix/test_assembler.py
```diff
@@ -1,12 +1,14 @@
...
-"""Unit test suite for `cr.cube.matrix.assembler` module."""
+"""Unit test suite for `cr.cube.cubepart assembly` module."""
 
+import mock
 import numpy as np
 import pytest
 
 from cr.cube.cube import Cube
-from cr.cube.dimension import Dimension, _Element, _OrderSpec, _Subtotal, _Subtotals
+from cr.cube.cubepart import _Slice
```
gergness, line 10: Should we move these tests that depend on `_Slice` to `tests/unit/test_cubepart.py`?

gergness (approved):
Looks good to me, bummer that the only file that grew was already one of the longest. I think I agree the lower total line count is worth it.

I think it was Slobodan who had strongest desire to do code review, but I realize we should probably check if Ernesto wants to as well.

Also because I usually forget, you've run zoom exporter tests against these changes, right?

src/cr/cube/cubepart.py
aminusfu: it does not remove the lazyness; the weighted_counts attribute is just a new name for the same property getter

tests/unit/matrix/test_assembler.py
aminusfu: I kept these in test_assembler just for some modularity; although they are now testing methods on the _Slice/_Strand object, they all still call `assemble_*` at some point and it seemed good to keep that separate.

src/cr/cube/cubepart.py
```diff
@@ -19,12 +19,21 @@
...
 
-from cr.cube.enums import CUBE_MEASURE as CM, DIMENSION_TYPE as DT, ORDER_FORMAT
+from cr.cube.collator import PayloadOrderCollator
+from cr.cube.enums import (
+    CUBE_MEASURE as CM,
+    DIMENSION_TYPE as DT,
+    ORDER_FORMAT,
+    MARGINAL_ORIENTATION as MO,
+)
+from cr.cube.matrix.assembler import _BaseOrderHelper
```
ernestoarbitrio, line 29: Could you get rid of the `_` in the `_BaseOrderHelper` name?

src/cr/cube/cubepart.py
```diff
@@ -19,12 +19,21 @@
...
+from cr.cube.matrix.measure import SecondOrderMeasures
+from cr.cube.matrix.subtotals import SumSubtotals
 from cr.cube.min_base_size_mask import MinBaseSizeMask
-from cr.cube.matrix import Assembler
 from cr.cube.measures.pairwise_significance import PairwiseSignificance
 from cr.cube.scalar import MeansScalar
-from cr.cube.stripe.assembler import StripeAssembler
+from cr.cube.stripe.assembler import _BaseOrderHelper as stripe_BaseOrderHelper
```
ernestoarbitrio, line 35: I'd go with:
```
from cr.cube.stripe.assembler import BaseOrderHelper as StripeBaseOrderHelper
```

tests/integration/test_stripe.py
```diff
@@ -17,63 +16,59 @@ class DescribeStripeAssembler:
 
     def it_provides_values_for_univariate_cat(self):
```
ernestoarbitrio, line 17: I guess this test should be moved in test_cubepart, or maybe this is not needed anymore cause in the test_cubepart (integration) we do have something similar (to be checked)

ernestoarbitrio (commented):
LGTM! a few minor comments, please see if you can address them :D

src/cr/cube/collator.py
```diff
@@ -16,7 +16,7 @@
 
 import numpy as np
 
-from cr.cube.dimension import Dimension, _Element, _OrderSpec, _Subtotals
+from cr.cube.dimension import Dimension, Element, _OrderSpec, _Subtotals
```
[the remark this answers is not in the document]
aminusfu: I think they should be once they are refactored and are worth being public. I plan to do that in a subsequent PR.

PR #391, opened 2023-11-30
[#186590795]: Implement squared weights in `cr.cube`

slobodan-ilic:


malecki (approved):
Thanks for the tabular comparison, this checks out

PR #393, opened 2024-01-27
[#186925787]: Fix pairwise comparison

slobodan-ilic:
* Pairwise comparison t-stats to use weighted bases


src/cr/cube/matrix/measure.py
```diff
@@ -1545,21 +1545,27 @@ def _column_bases(self):
             calculated column bases (either 'effective' or unweighted counts) for
             the analysis.
         """
-        unweighted_blocks = self._second_order_measures.column_unweighted_bases.blocks
+        # unweighted_blocks = self._second_order_measures.column_unweighted_bases.blocks
```
ernestoarbitrio, line 1548: get rid of commented code plz :D

src/cr/cube/matrix/measure.py
```diff
@@ -1545,21 +1545,27 @@ def _column_bases(self):
...
+        weighted_blocks = self._second_order_measures.column_weighted_bases.blocks
         if self._second_order_measures.columns_squared_base.is_defined:
             squared_blocks = self._second_order_measures.column_squared_bases.blocks
             effective_blocks = [
                 [
-                    unweighted_blocks[0][0] ** 2 / squared_blocks[0][0],
-                    unweighted_blocks[0][1] ** 2 / squared_blocks[0][1],
+                    # unweighted_blocks[0][0] ** 2 / squared_blocks[0][0],
+                    # unweighted_blocks[0][1] ** 2 / squared_blocks[0][1],
```
ernestoarbitrio, line 1555: same

src/cr/cube/matrix/assembler.py
```diff
@@ -64,15 +64,24 @@ def row_display_order(cls, dimensions, second_order_measures, format):
...
-            else _SortRowsByDerivedColumnHelper
-            if collation_method == CM.OPPOSING_INSERTION and dim_type in DT.ARRAY_TYPES
-            else _SortRowsByInsertedColumnHelper
-            if collation_method == CM.OPPOSING_INSERTION
-            else _SortRowsByLabelHelper
-            if collation_method == CM.LABEL
-            else _SortRowsByMarginalHelper
-            if collation_method == CM.MARGINAL
-            else _RowOrderHelper
+            else (
```
ernestoarbitrio, line 67: IMHO this is quite difficult to read. Could you think about something more readable?

ernestoarbitrio (approved):
just some minor comments

src/cr/cube/matrix/assembler.py
slobodan-ilic: yeah this is crap... I dunno man... will try to figure out something better... maybe the switch?

slobodan-ilic: Is this better:

```
    @classmethod
    def factory(cls, cube, dimensions, slice_idx):
        """Return _BaseCubeMeans subclass instance appropriate to `cube`.

        Raises `ValueError` if the cube-result does not include a cube-means measure.
        """
        if cube.means is None:
            raise ValueError("cube-result does not contain cube-means measure")

        def get_cube_means_cls(dim_types):
            if dim_types == (DT.MR, DT.MR):
                return _MrXMrCubeMeans
            if dim_types[0] == DT.MR:
                return _MrXCatCubeMeans
            if dim_types[1] == DT.MR:
                return _CatXMrCubeMeans
            return _CatXCatCubeMeans

        dimension_types = cube.dimension_types[-2:]
        CubeMeansCls = get_cube_means_cls(dimension_types)

        return CubeMeansCls(
            dimensions, cube.means[cls._slice_idx_expr(cube, slice_idx)]
        )
```

ernestoarbitrio: yes I like it more

src/cr/cube/matrix/measure.py
```diff
@@ -1545,21 +1545,27 @@ def _column_bases(self):
...
+                    # unweighted_blocks[1][0] ** 2 / squared_blocks[1][0],
+                    # unweighted_blocks[1][1] ** 2 / squared_blocks[1][1],
+                    weighted_blocks[1][0] ** 2 / squared_blocks[1][0],
+                    weighted_blocks[1][1] ** 2 / squared_blocks[1][1],
                 ],
             ]
             return effective_blocks
-        return unweighted_blocks
+        # return unweighted_blocks
+        return weighted_blocks
```
malecki, line 1568: I think the outer return here is what happens in the absence of weighting, or absence of w^2. In the first case (no weights) weighted == unweighted, which would be fine. In the second instance, where you might have weights but fail to supply w^2, unweighted would still be correct here. The bug was only in the effective calculation above.

malecki (changes requested):

src/cr/cube/matrix/measure.py
slobodan-ilic: @malecki Can you re-check this? I think I figured it out finally.

malecki (approved):
Yes, much more reasonable set of changed expectations. Thank you!

PR #394, opened 2024-02-01
remove deepcopy from cube response

ernestoarbitrio:


src/cr/cube/cube.py
```diff
@@ -602,7 +597,7 @@ def _measures(self) -> "_Measures":
         Provides access to count based measures and numeric measures (e.g. mean, sum)
         when available.
         """
-        return _Measures(self._cube_dict, self._all_dimensions, self._cube_idx_arg)
+        return _Measures(self._cube_response, self._all_dimensions, self._cube_idx_arg)
```
gaetano-guerriero, line 600: Here we build `_Measures` without the "dimension inflation". Do we still need it ?

gaetano-guerriero (commented):
Great simplification in removing the `deepcopy`, I have a small question for `_Measures`

src/cr/cube/cube.py
ernestoarbitrio: actually self._all_dimensions has the inflation

gaetano-guerriero (approved):

gaetano-guerriero (approved):

PR #395, opened 2024-02-23
Remove deepcopy from the codebase

ernestoarbitrio:
https://www.pivotaltracker.com/story/show/187090869

deepcopy(s) causing OOM on task system

### DIFF
--- a/src/cr/cube/dimension.py
+++ b/src/cr/cube/dimension.py
@@ -933,7 +933,7 @@ class Element:
         return self._element_dict.get("value", {}).get("references", {}).get("anchor")
 
     @lazyproperty
-    def element_id(self) -> int:
+    def element_id(self) -> Union[int, str]:
         """int identifier for this category or subvariable."""
         return _build_element_id(self._element_dict, self._dim_type)
 
--- a/src/cr/cube/dimension.py
+++ b/src/cr/cube/dimension.py
@@ -649,7 +649,9 @@ class _ElementIdShim:
 
         # --- array variables are identified by thier aliases
         if self.dimension_type in DT.ARRAY_TYPES:
-            # --- Replace element ids with the alias
+            # --- Add new field to the shim for the alias to be associated to the
+            # --- existing ID. This augmentation of the dimension dict it a temporary
+            # --- hack for subvariable ref that should change in future
             for idx, alias in enumerate(self._subvar_aliases):
                 shim["type"]["elements"][idx]["subvar_alias"] = alias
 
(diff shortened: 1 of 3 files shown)


src/cr/cube/dimension.py
```diff
@@ -896,11 +906,14 @@ class Element:
     This object resolves the transform cascade for element-level transforms.
     """
 
-    def __init__(self, element_dict, index, element_transforms, label_formatter):
+    def __init__(
+        self, element_dict, index, element_transforms, label_formatter, dim_type
```
rikirenz, line 910: Maybe `dim_type` can be an optional value with default value `None`.

src/cr/cube/dimension.py
```diff
@@ -675,7 +685,7 @@ def shimmed_dimension_transforms_dict(self) -> Dict:
         - "alias": Subvariables also have an alias that identifies them. It is stored
           in `dimensions[i].type.elements[j].value.references.alias`.
         """
-        shim = copy.deepcopy(self._dimension_transforms_dict)
+        shim = self._dimension_transforms_dict
```
rikirenz, line 688: Makes sense.
Any changes made to shim will directly affect `self._dimension_transforms_dict` and vice versa. 

This will definitely solve the problem.
But this could potentially lead to unintended side effects if modifications to shim were meant to be isolated but I guess this is not the case.

Am I right?

src/cr/cube/dimension.py
```diff
@@ -629,33 +639,33 @@ def __init__(self, dimension_type, dimension_dict, dimension_transforms_dict):
...
     def shimmed_dimension_dict(self) -> Dict:
-        """Copy of dimension dictionary with shimmed `element_id`s
+        """Augmented dimension dictionary with shimmed `element_id`s
 
         We want to move to a world where elements on a subvariables dimension are
         identified by their alias, but right now the "element_id" from zz9 is
         an index for subvariables.
         """
-        shim = copy.deepcopy(self._dimension_dict)
+        shim = self._dimension_dict
```
rikirenz, line 648: Please check the other comment. My main concern is the same.

rikirenz (approved):
LGTM. I believe this will solve the spikes in memory.

I have just 1 small question. Please let me know.

src/cr/cube/dimension.py
ernestoarbitrio: well this should be mandatory, so I do prefer to change the unit tests with `None` in order to prevent any future regression

src/cr/cube/dimension.py
ernestoarbitrio: Well the shim has been modified for sure but with an `append` in this case, all the original part are not touched, so I think we're safe, but of course for this aspect I'd like to have @gergness opinion

src/cr/cube/dimension.py
ernestoarbitrio: same answer

PR #398, opened 2024-03-18
augment single col filter cubes

ernestoarbitrio:


gergness (approved):
Looks good, only thing I don't get is what `_is_single_filter_col_cube` means? Is it just an indicator that a set of cubes is from a tabook? Or does it mean something specific about what kind of variables are being used?

ernestoarbitrio:
> Looks good, only thing I don't get is what `_is_single_filter_col_cube` means? Is it just an indicator that a set of cubes is from a tabook? Or does it mean something specific about what kind of variables are being used?

It's a key that the querybuilder adds in the cube response to indicate that that respose is a cube made from a filter cause this needs to be used somewhere in exporter to check some conditions.

PR #401, opened 2024-07-24
Enum refactoring 187998654

ernestoarbitrio:


gergness (approved):
I don't have strong opinions on this, but not sure why moving the `NUMERIC_CUBE_MEASURES` out of enums is helpful? To me it feels like the natural place to define collections of measures is in `cr.cube.enums`, even if some of them are only used in one place.

I lose track of which `enum` conventions support what, but could it be a property of `CUBE_MEASURES` or does that only work when we don't subclass `enum.Enum`?

Again, I don't have strong opinions, so feel free to overrule.

ernestoarbitrio:
> I don't have strong opinions on this, but not sure why moving the `NUMERIC_CUBE_MEASURES` out of enums is helpful? To me it feels like the natural place to define collections of measures is in `cr.cube.enums`, even if some of them are only used in one place.
> 
> I lose track of which `enum` conventions support what, but could it be a property of `CUBE_MEASURES` or does that only work when we don't subclass `enum.Enum`?
> 
> Again, I don't have strong opinions, so feel free to overrule.

agreed done! Thanks

PR #402, opened 2024-11-11
[188552693]: Rework subtotal diffs for cat dates

ernestoarbitrio:


gergness:
We'll talk tomorrow, but here's what I was thinking this morning. I haven't made sure it will work, but my suspicion is that it will be best to change the behavior in `matrix.measures._ColumnProportions` here:
https://github.com/Crunch-io/crunch-cube/blob/adb1c24da3ec22717edb529f5db511248895d9be/src/cr/cube/matrix/measure.py#L784-L812

(and for `_RowProportions` and I guess `_TableProportions`?)

PR #403, opened 2024-11-18
[#188552693]: Rework subtotals

ernestoarbitrio:


src/cr/cube/matrix/subtotals.py
```diff
@@ -370,6 +371,166 @@ def _subtotal_row(self, subtotal):
...
+    This class handles a special case for wave differences when a CAT_DATE variable is
+    involved in the calculation.
+
+    A wave difference for a CAT_DATE variable is calculate subtracting at the
+    percentages level: (count1/base1) - (count2/base2).
+    """
+
+    def __init__(
+        self,
+        base_values,
```
gergness, line 386: I'm not sure how much this matters, but this is confusingly named...

In the `_BaseSubtotals` class, `base_values` means the non-subtotal rows & columns (eg `self.blocks[0][0]`, the base from which the subtotals are calculated).

But this new class we're passing in the `base_values` of the "weighted_base" measure and the "counts" measure. (but calling the "weighted_base" the `base_values).

It doesn't matter because we're not using the `.blocks` property (just the `subtotal_rows` and `subtotal_columns`, but this means that the first element in the `.blocks` property is wrong.

I kind of find the `blocks` thing overwrought, so maybe the easiest way to make this clear is to not have `WaveDiffSubtotal` inherit from `_BaseSubtotal` and only have the `subtotal_rows` & `subtotal_columns` properties? Then we won't have to contend with the 2 different meanings of "base".

Another option that would make more sense to me would be to pass in
`self._count_blocks[0][0] / self._weighted_base_blocks[0][0]` as the `base_values` and then pass in the subtotal rows/columns counts and weighted bases separately. When we're calculating the wave difference the calculation should subtract the base values from each other, but when we're in a regular proportion it should be `count/base`.

src/cr/cube/matrix/subtotals.py
```diff
@@ -370,6 +371,166 @@ def _subtotal_row(self, subtotal):
...
+        counts,
+        default_insertions,
+        dimensions,
+        diff_cols_nan=False,
```
gergness, line 390: These aren't actually used, right? Can you remove them?

src/cr/cube/matrix/subtotals.py
```diff
@@ -370,6 +371,166 @@ def _subtotal_row(self, subtotal):
...
+
+        return default
+
+    def _subtotal_row(self, subtotal, default):
+        """Return (n_cols,) ndarray of values for `subtotal` row."""
+
+        if (
+            self._dimensions[0].dimension_type == DT.CAT_DATE
+            and len(subtotal.subtrahend_idxs) > 0
+            and len(subtotal.addend_idxs) > 0
```
gergness, line 502: Maybe this should just be cat date with any subtrahends? Seems weird that having 2 subtrahends and 0 addend would work the old way, but 2 subtrahends and 1 addend would be NaN?

gergness (changes requested):

gergness (approved):
LGTM

PR #407, opened 2024-11-26
bug fix len vs any

ernestoarbitrio:


src/cr/cube/matrix/subtotals.py
```diff
@@ -459,8 +459,9 @@ def _subtotal_columns(self):
 
     def _subtotal_column(self, subtotal, default):
         """Return (n_rows,) ndarray of values for `subtotal` column."""
-        if self._dimensions[1].dimension_type == DT.CAT_DATE and any(
```
ernestoarbitrio, line 462: `any([0])` returns False and this is a bug cause the `[0]` means we have a subtrahend in the first position.

gergness (approved):
LGTM. !dm truthiness

PR #409, opened 2024-12-10
Sort by value column labels

ernestoarbitrio:


tests/integration/test_cubepart.py
```diff
@@ -1016,9 +1058,28 @@ def test_it_can_sort_by_marginal_with_nan_in_body(self):
...
+                    "Quadrefiore",
+                    "Orecchiette",
+                    "Fileja",
+                    "Bucatini",
+                    "Boccoli",
+                ],
+            ),
+        ),
+    )
+    def test_it_can_fix_order_of_subvars_identified_by_bogus_id(
```
gergness, line 1078: I kind of think this test should be split out rather than parametrized - the test name is about MR bogus ids (the integers rather than MR alias), but since it is MRxCAT, your new column based test is actually using the real category ids.

Also, by "luck" there's no difference between "fixed category 3 to the bottom" because apparently category 3 is "Boccoli" which would have been last anyways.

gergness (approved):
Minor comments but looks good

PR #411, opened 2025-03-18
Sort col by opposing row

ernestoarbitrio:


gergness (approved):
Looks good.

Only thing I see is that I don't think the pairwise letters work properly, but that's still true of any order transform, right? Surprised we haven't been asked to fix that yet.

PR #412, opened 2025-05-01
[ZC-617]: add audience ratio measure

ernestoarbitrio:


.github/workflows/ci.yml
```diff
@@ -10,14 +10,14 @@ on:
 
 jobs:
   build:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
```
ernestoarbitrio, line 13: gh removed runner on 20.04

.github/workflows/ci.yml
```diff
@@ -10,14 +10,14 @@ on:
...
     env:
       GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       COVERALLS_REPO_TOKEN: ${{ secrets.COVERALLS_REPO_TOKEN }}
     strategy:
       max-parallel: 5
       matrix:
-        python-version: ["3.6", "3.7", "3.8", "3.9", "3.10", "3.11", "3.12"]
+        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
```
ernestoarbitrio, line 20: ubuntu 22.04 min python support is 3.8

tox.ini
```diff
@@ -13,22 +11,20 @@ python =
...
+    py3: python3.12
 
 commands =
     pip install cr.cube[testing]
     py.test --cov --cov-report=xml {posargs:}
 
 [testenv:lint]
 skip_install = True
-basepython = python3.8
+basepython = python3.10
```
ernestoarbitrio, line 27: default python version in ubuntu 22.04 is 3.10

src/cr/cube/matrix/assembler.py
```diff
@@ -50,19 +50,17 @@ def column_display_order(cls, dimensions, second_order_measures, format):
         # --- form consistent with `.row_display_order()` and we'll elaborate this when
         # --- we add sort-by-value to columns.
         collation_method = dimensions[1].order_spec.collation_method
+        # fmt: off
```
gergness, line 53: !dm black 😢 

For my curiosity, there's a native switch statement in newer versions of python right? How far off are we from being able to use it?

src/cr/cube/matrix/measure.py
```diff
@@ -830,6 +835,80 @@ def _weighted_base_blocks(self):
         return self._second_order_measures.column_weighted_bases.blocks
 
 
+class _AudienceRatio(_ColumnProportions):
+    """Provides the audience-ratio measure for a matrix.
+
+    This measure (also known as "Index") is a 2D np.float64 ndarray of the ratio of the
+    proportions of the first column to the proportions of the second column.
+
+    Audience ratio (Index) is a convenient way of showing the ratio of the Target % and
```
gergness, line 844: This along with smoothing and the reference column statistical significance really makes me wish we had time to design a way to pass arguments into measures better. (so the audience measure would specify which column was the control column instead of being fixed to the second one)

src/cr/cube/matrix/measure.py
```diff
@@ -830,6 +835,80 @@ def _weighted_base_blocks(self):
...
+    def _intersections(self):
+        """(n_row_subtotals, n_col_subtotals) ndarray of intersection values.
+
+        An intersection value arises where a row-subtotal crosses a column-subtotal.
+
+        Always nan because the audience ratio is not defined for the intersection
+        """
+        # --- do not propagate divide-by-zero warnings to stderr ---
+        with np.errstate(divide="ignore", invalid="ignore"):
+            intersections = self._count_blocks[1][1] / self._weighted_base_blocks[1][1]
```
gergness, line 881: If you're just using it for the shape, do you need to divide anything?

src/cr/cube/matrix/measure.py
```diff
@@ -830,6 +835,80 @@ def _weighted_base_blocks(self):
...
+        """2D np.float64 ndarray of audience ratio values.
+
+        This is the second "block" and has the shape (n_rows, n_col_subtotals).
+
+        Always empty because the audience ratio has no subtotal columns
+        """
+        # --- do not propagate divide-by-zero warnings to stderr ---
+        with np.errstate(divide="ignore", invalid="ignore"):
+            default_value = self._count_blocks[0][1] / self._weighted_base_blocks[0][1]
+            return np.empty(default_value.shape)
```
gergness, line 895: Same, why not just use `self._count_blocks[0][1]` for the shape?

.github/workflows/ci.yml
gergness: Zoom is past 3.8 right?

gergness (approved):
Looks good, a minor suggestion

src/cr/cube/matrix/assembler.py
ernestoarbitrio: yes there's the `match` statement that is introduced in 3.10, we could use that but we need to drop 3.8 and 3.9 support, that is feasible to me considering that exporter works on 3.11 now, but dunno if cube is used even directly in other lib like pycrunch or scrunch, so I'd live it with this horrible thing for now :D ... and yes !dm black

src/cr/cube/matrix/measure.py
ernestoarbitrio: yes I do agree, and I had a PR ready for pw signif 2 years ago, but it has been frozen since then because the FE needed to change to much stuff, and since then no PRD has been approved to take care of this issue :(

src/cr/cube/matrix/measure.py
ernestoarbitrio: u right, done

src/cr/cube/matrix/measure.py
ernestoarbitrio: u right

.github/workflows/ci.yml
ernestoarbitrio: yes

PR #418, opened 2026-04-15
Provide more information about missing values

gergness:
This is to support the proposed displays of missingness for the explore view. 
Details in notion here:  https://www.notion.so/Tile-aggregation-V2-30c8529f379e80aba348f6c89fb19517?source=copy_link#33d8529f379e80c69430d7a9ba30e230

Essentially we need to add:
- Ability to break out counts by types of missing (system missing & user defined ones like, "Refused", "Don't Know", etc.) on the display of categorical & categorical array variables
- Ability to show a count of the missing values along an MR (in the same way we show the bases)

Some things I'm not sure I've quite got right yet:
- The types of `*_disaggregated_missing_unweighted_counts` are weird - For `_Slice`, I have 1D arrays of tuples. This allows me to treat them as marginals when arranging the slice, but feels weird.
- I've introduced `_Strand.table_base` to support `_Strand.table_missing`, which does the same weird non-deterministic return dimensions as the `_Slice.rows/columns_base`. I wanted the symmetry, but I don't think we generally want unstable return types like this.
- Even though we are currently only showing displays of 1 variable, we have to add support in both `_Strand` & `_Slice` because: 1) categorical arrays have 2 dimension and 2) we want the aggregate API to work with arbitrary tiles, not just the ones on the explore view. I've tried not to over engineer features we don't use yet, but not sure I've got this balance right. (For example, I don't support subtotals in the disaggregated missing values because we don't have CAT x CAT yet, but I do have the new `rows_/columns_/tables_missing` values work on CAT even though they won't be used.

### DIFF
--- a/src/cr/cube/cubepart.py
+++ b/src/cr/cube/cubepart.py
@@ -1345,7 +1345,7 @@ class _Slice(CubePartition):
         return self._measures.rows_disaggregated_missing_unweighted_counts.labels
 
     @lazyproperty
-    def rows_disaggregated_missing_counts(self):
+    def rows_disaggregated_missing_unweighted_counts(self):
         """Optional 1D ndarray of tuples of missing values by type
 
         The disaggregated missings are only available on strands with CAT dimensions.
--- a/src/cr/cube/cubepart.py
+++ b/src/cr/cube/cubepart.py
@@ -463,6 +463,19 @@ class _Slice(CubePartition):
         return self._dimensions[1].name
 
     @lazyproperty
+    def columns_disaggregated_missing_element_ids(self):
+        """Optional tuple of element ids of disaggregated missing elements
+
+        The disaggregated missings are only available on slices with CAT dimensions.
+        The tuple's shape is not related to the shape of actual categories, it depends
+        on how many missing categories are defined (generally there's a system missing
+        and possibly a handful of missing categories like "Refused").
+        """
+        return (
+            self._measures.columns_disaggregated_missing_unweighted_counts.element_ids
+        )
+
+    @lazyproperty
     def columns_disaggregated_missing_labels(self):
         """Optional tuple of labels of disaggregated missing elements
 
@@ -1339,6 +1352,17 @@ class _Slice(CubePartition):
         return self._rows_dimension.dimension_type
 
     @lazyproperty
+    def rows_disaggregated_missing_element_ids(self):
+        """Optional tuple of element ids of disaggregated missing elements
+
+        The disaggregated missings are only available on slices with CAT dimensions.
+        The tuple's shape is not related to the shape of actual categories, it depends
+        on how many missing categories are defined (generally there's a system missing
+        and possibly a handful of missing categories like "Refused").
+        """
+        return self._measures.rows_disaggregated_missing_unweighted_counts.element_ids
+
+    @lazyproperty
     def rows_disaggregated_missing_labels(self):
         """Optional tuple of labels of disaggregated missing elements
 
@@ -2118,6 +2142,17 @@ class _Strand(CubePartition):
         n_valids = len(rows_dim.valid_elements)
         diffs = [False] * n_valids + [e.is_difference for e in rows_dim.subtotals]
         return tuple(np.where(np.array(diffs)[self._row_order_signed_indexes])[0])
+
+    @lazyproperty
+    def disaggregated_missing_element_ids(self):
+        """Optional tuple of element_ids for missing categories
+
+        The disaggregated missings are only available on strands with CAT dimensions.
+        The tuple's shape is not related to the shape of actual categories, it depends
+        on how many missing categories are defined (generally there's a system missing
+        and possibly a handful of missing categories like "Refused").
+        """
+        return self._measures.disaggregated_missings.element_ids
 
     @lazyproperty
     def disaggregated_missing_labels(self):
--- a/src/cr/cube/cubepart.py
+++ b/src/cr/cube/cubepart.py
@@ -2533,7 +2533,7 @@ class _Strand(CubePartition):
 
     @lazyproperty
     def table_base(self):
-        """scalar/1D np.float64 ndarray of unweighted-N for the table/each cell of strand.
+        """scalar/1D np.float64 ndarray of unweighted-N for table/each cell of strand.
 
         This array is 1D (a distinct base for each cell) when the rows dimension is MR,
         because each MR-subvariable has its own unweighted N. This is because not every
@@ -2569,7 +2569,7 @@ class _Strand(CubePartition):
 
     @lazyproperty
     def table_missing(self):
-        """scalar/1D np.float64 ndarray of unweighted-N for the table/each cell of strand.
+        """scalar/1D np.float64 ndarray of unweighted-N for table/each cell of strand.
 
         This is the complement to `.table_base`
         """
--- a/src/cr/cube/cubepart.py
+++ b/src/cr/cube/cubepart.py
@@ -2056,6 +2056,28 @@ class _Strand(CubePartition):
         return tuple(np.where(np.array(diffs)[self._row_order_signed_indexes])[0])
 
     @lazyproperty
+    def disaggregated_missing_labels(self):
+        """Optional tuple of labels for missing categories
+
+        The disaggregated missings are only available on strands with CAT dimensions.
+        The tuple's shape is not related to the shape of actual categories, it depends 
+        on how many missing categories are defined (generally there's a system missing 
+        and possibly a handful of missing categories like "Refused").
+        """
+        return self._measures.disaggregated_missings.labels
+
+    @lazyproperty
+    def disaggregated_missing_unweighted_counts(self):
+        """Optional tuple of unweighted counts for missing categories
+
+        The disaggregated missings are only available on strands with CAT dimensions.
+        The tuple's shape is not related to the shape of actual categories, it depends 
+        on how many missing categories are defined (generally there's a system missing 
+        and possibly a handful of missing categories like "Refused").
+        """
+        return self._measures.disaggregated_missings.unweighted_counts
+
+    @lazyproperty
     def inserted_row_idxs(self):
         """tuple of int index of each inserted row in this strand.
 
--- a/src/cr/cube/cubepart.py
+++ b/src/cr/cube/cubepart.py
@@ -2569,9 +2569,9 @@ class _Strand(CubePartition):
 
     @lazyproperty
     def table_missing(self):
-        """scalar/1D np.float64 ndarray of unweighted-N for table/each cell of strand.
+        """scalar/1D np.float64 ndarray of unwtd-N of missings for table/each cell.
 
-        This is the complement to `.table_base`
+        This is the complement to `.table_base`.
         """
         return self._cube.n_responses - self.table_base
 
--- a/src/cr/cube/cubepart.py
+++ b/src/cr/cube/cubepart.py
@@ -466,7 +466,7 @@ class _Slice(CubePartition):
     def columns_disaggregated_missing_labels(self):
         """Optional tuple of labels of disaggregated missing elements
 
-        The disaggregated missings are only available on strands with CAT dimensions.
+        The disaggregated missings are only available on slices with CAT dimensions.
         The tuple's shape is not related to the shape of actual categories, it depends
         on how many missing categories are defined (generally there's a system missing
         and possibly a handful of missing categories like "Refused").
@@ -475,14 +475,17 @@ class _Slice(CubePartition):
 
     @lazyproperty
     def columns_missing(self):
-        """TKTKTKTK"""
+        """1D/2D np.float64 ndarray of unweighted counts of missing for each cell/column
+
+        This is the complement to `.columns_base`.
+        """
         return self._cube.n_responses - self.columns_base
 
     @lazyproperty
     def columns_disaggregated_missing_unweighted_counts(self):
         """Optional 1D ndarray of tuples of missing values by type
 
-        The disaggregated missings are only available on strands with CAT dimensions.
+        The disaggregated missings are only available on slices with CAT dimensions.
         The tuple's shape is not related to the shape of actual categories, it depends
         on how many missing categories are defined (generally there's a system missing
         and possibly a handful of missing categories like "Refused").
@@ -1339,7 +1342,7 @@ class _Slice(CubePartition):
     def rows_disaggregated_missing_labels(self):
         """Optional tuple of labels of disaggregated missing elements
 
-        The disaggregated missings are only available on strands with CAT dimensions.
+        The disaggregated missings are only available on slices with CAT dimensions.
         The tuple's shape is not related to the shape of actual categories, it depends
         on how many missing categories are defined (generally there's a system missing
         and possibly a handful of missing categories like "Refused").
@@ -1350,7 +1353,7 @@ class _Slice(CubePartition):
     def rows_disaggregated_missing_unweighted_counts(self):
         """Optional 1D ndarray of tuples of missing values by type
 
-        The disaggregated missings are only available on strands with CAT dimensions.
+        The disaggregated missings are only available on slices with CAT dimensions.
         The tuple's shape is not related to the shape of actual categories, it depends
         on how many missing categories are defined (generally there's a system missing
         and possibly a handful of missing categories like "Refused").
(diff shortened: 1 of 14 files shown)


src/cr/cube/stripe/measure.py
```diff
@@ -688,6 +734,12 @@ def table_base_range(self):
         bases = self._unweighted_cube_counts.bases
         return np.array([np.min(bases), np.max(bases)])
 
+    @lazyproperty
+    def table_base_scalar(self):
+        """"TKTKTK"""
+        if self._rows_dimension.dimension_type in DT.ARRAY_TYPES:
```
ernestoarbitrio: please fix this docstring :D

src/cr/cube/cubepart.py
```diff
@@ -462,6 +427,33 @@ def columns_dimension_name(self):
...
+        The disaggregated missings are only available on strands with CAT dimensions.
+        The tuple's shape is not related to the shape of actual categories, it depends 
+        on how many missing categories are defined (generally there's a system missing 
+        and possibly a handful of missing categories like "Refused").
+        """
+        return self._measures.columns_disaggregated_missing_unweighted_counts.labels
+
+    @lazyproperty
+    def columns_missing(self):
+        """TKTKTKTK"""
```
ernestoarbitrio: docstring :D

src/cr/cube/cubepart.py
```diff
@@ -1351,6 +1350,11 @@ def rows_margin_proportion(self):
         # --- otherwise rows-margin is a vector ---
         return self._assemble_marginal(self._measures.rows_table_proportion)
 
+    @lazyproperty
+    def rows_missing(self):
+        """TKTKTKTK"""
```
ernestoarbitrio: docstring

src/cr/cube/cubepart.py
```diff
@@ -2465,6 +2477,11 @@ def table_margin_range(self):
         """
         return self._measures.weighted_bases.table_margin_range
 
+    @lazyproperty
+    def table_missing(self):
+        """TKTKTKTK"""
```
ernestoarbitrio: docstring

src/cr/cube/cubepart.py
```diff
@@ -2445,6 +2441,22 @@ def tab_alias(self):
...
+    def table_base(self):
+        """scalar/1D np.float64 ndarray of unweighted-N for the table/each cell of strand.
+
+        This array is 1D (a distinct base for each cell) when the rows dimension is MR,
+        because each MR-subvariable has its own unweighted N. This is because not every
+        possible response is necessarily offered to every respondent.
+
+        In all other cases, it is a scalar, containing one value for the whole table.
+        """
+        if self._measures.unweighted_bases.table_base_scalar:
```
ernestoarbitrio: This uses truthiness, so if `table_base_scalar` returns 0 (or 0.0), it would be falsy and fall through to self.rows_base. Should it be:                                                                         
```                                                                                                                                                                                                                
if self._measures.unweighted_bases.table_base_scalar is not None:
```
instead?

src/cr/cube/matrix/measure.py
```diff
@@ -2698,6 +2651,68 @@ def _opposing_numeric_values(self):
...
+            raise ValueError(
+                f"{self.orientation.value}-disaggreagted-values is defined "
+                "only across categorical dimensions."
+            )
+        missing_mask = [
+            idx for idx, el 
+            in enumerate(self._opposing_dimension.all_elements) 
+            if el.missing
+        ]
+        dtype = ",".join(["f"] * len(missing_mask))
```
ernestoarbitrio: This creates a structured numpy array with float32 fields (f). If I remember the rest of the codebase uses float64 (np.float64). Consider using "d" (float64) instead of "f" to match precision conventions.

src/cr/cube/matrix/measure.py
```diff
@@ -2679,11 +2618,25 @@ def _counts_are_defined(self):
             return self._second_order_measures.column_comparable_counts.is_defined
         return self._second_order_measures.row_comparable_counts.is_defined
 
+    @lazyproperty
+    def _opposing_dimension(self):
+        """opposing dimension for this marginal's orientation"""
+        if self._orientation == MO.ROWS:
```
ernestoarbitrio: for consistency I'd use `.orientation` or `._orientation` in both this property and `_subtotal_shape`

src/cr/cube/cubepart.py
```diff
@@ -2055,6 +2037,28 @@ def diff_row_idxs(self):
...
+    @lazyproperty
+    def disaggregated_missing_unweighted_counts(self):
+        """Optional tuple of unweighted counts for missing categories
+
+        The disaggregated missings are only available on strands with CAT dimensions.
+        The tuple's shape is not related to the shape of actual categories, it depends 
+        on how many missing categories are defined (generally there's a system missing 
+        and possibly a handful of missing categories like "Refused").
+        """
+        return self._measures.disaggregated_missings.unweighted_counts
```
ernestoarbitrio, line 2177: this `_Strand.disaggregated_missing_unweighted_counts` returns a tuple, while `_Slice.rows_disaggregated_missing_unweighted_counts` returns an ndarray of       
  structured tuples (via _assemble_marginal → np.hstack). 
This can create API inconsistency for consumers. I'm inclined to think that we should unify the return type (or de we have this kind of mixing return types somewhere else too? Not remembering right now :D )

src/cr/cube/cubepart.py
```diff
@@ -462,6 +427,33 @@ def columns_dimension_name(self):
         """
         return self._dimensions[1].name
 
+    @lazyproperty
+    def columns_disaggregated_missing_labels(self):
+        """Optional tuple of labels of disaggregated missing elements
+
+        The disaggregated missings are only available on strands with CAT dimensions.
```
ernestoarbitrio: in this docstring you are mentioning strand but this property is on the _Slice

src/cr/cube/matrix/measure.py
```diff
@@ -2698,6 +2651,68 @@ def _opposing_numeric_values(self):
         )
 
 
+class _DisaggregatedMissingValues(_BaseMarginal):
+    """Provides the dissaggregated missing values for a slice
```
ernestoarbitrio: typo: dissaggregated :D

src/cr/cube/stripe/measure.py
```diff
@@ -216,6 +221,47 @@ def _smoother(self):
...
+    """Provides the dissaggregated missing values for a stripe
+
+    Disaggregated missing values are the missing values from a categorical
+    dimension that show the counts per missing category (whether it be
+    system missing or user defined ones like "Refused"). 
+
+    The dimensionality of this is weird. It can be thought of as a marginal,
+    which on a stripe would generally mean that it's 0D. However, because there
+    can be more than one type of missing, we gain back a dimension. The shape
+    is therefore not necesarilly the same as the shape of measures. For this
```
ernestoarbitrio: typo: necesarilly :D

src/cr/cube/matrix/measure.py
```diff
@@ -2698,6 +2651,68 @@ def _opposing_numeric_values(self):
...
+
+    Disaggregated missing values are the missing values from a categorical
+    dimension that show the counts per missing category (whether it be
+    system missing or user defined ones like "Refused"). 
+
+    The dimensionality of this is weird. It can be thought of as a marginal,
+    which on a slice would generally mean that it's 1D. However, because there
+    can be more than one type of missing, we gain back a dimension. The shape
+    is therefore not necesarilly the same as the shape of measures. For this
```
ernestoarbitrio: typo: necesarilly ?

src/cr/cube/matrix/measure.py
```diff
@@ -2698,6 +2651,68 @@ def _opposing_numeric_values(self):
         )
 
 
+class _DisaggregatedMissingValues(_BaseMarginal):
```
ernestoarbitrio: This class `_DisaggregatedMissingValues` class has non-trivial logic (structured dtypes, transpose, masking) that would benefit from targeted unit tests, especially edge cases like zero missing categories or a single missing category.
I've only seen integration tests but maybe we can benefit from unit tests for this class

src/cr/cube/cubepart.py
```diff
@@ -462,6 +427,33 @@ def columns_dimension_name(self):
...
+        The tuple's shape is not related to the shape of actual categories, it depends 
+        on how many missing categories are defined (generally there's a system missing 
+        and possibly a handful of missing categories like "Refused").
+        """
+        return self._measures.columns_disaggregated_missing_unweighted_counts.labels
```
ernestoarbitrio, line 487: this goes directly to `_DisaggregatedMissingValues.labels` bypassing the `assemble_marginal`, instead `columns_disaggregated_missing_unweighted_counts` goes through the marginals and means that in case of a ordering the counts tuples will be re-ordered accordingly, can it be that this labels then do not respect the order at some point?

src/cr/cube/matrix/measure.py
```diff
@@ -2698,6 +2651,68 @@ def _opposing_numeric_values(self):
...
+        ]
+        dtype = ",".join(["f"] * len(missing_mask))
+
+        # --- Since we're collapsing to 1D anyways, it's easiest to just transpose the counts on columns
+        # --- orientation so the code is more similar between orientations
+        counts = self._cube_measures.unweighted_unconditional_cube_counts.counts
+        if self._orientation == MO.COLUMNS:
+            counts = counts.transpose()
+        counts = counts[:, missing_mask]
+        base_values = np.array(list(tuple(row) for row in counts), dtype=dtype)
```
ernestoarbitrio:
```suggestion
        base_values = np.array([tuple(row) for row in counts], dtype=dtype)
```

src/cr/cube/matrix/measure.py
```diff
@@ -2698,6 +2651,68 @@ def _opposing_numeric_values(self):
...
+
+        # --- For now, don't bother reimplementing the subtotal logic for 
+        # --- disaggregated misings. We aren't currently using them in a place
+        # --- where subtotals are possible, so wait until there is a product need
+        # --- for them
+        # --- Instead, just send NaNs if encountered
+        subtotal_values = np.repeat(np.array((np.nan, ) * len(missing_mask), dtype=dtype), self._subtotal_shape)
```
ernestoarbitrio:
```suggestion
      nan_tuple = (np.nan,) * len(missing_mask)                                                                                                                                                                 
      subtotal_values = np.array([nan_tuple] * self._subtotal_shape, dtype=dtype) 
```

ernestoarbitrio (changes requested):
overall looks ok, left some comments though

src/cr/cube/cubepart.py
gergness: IMO this is analogous to how `_Strand.scale_mean` returns a float but `_Slice.rows_scale_mean` returns an array of floats. The marginal on `_Strand` is a single item of the of the related marginal's array from `_Slice`. It looks weird because the tuple also has dimensionality, but I think it's similar enough.

Does that make sense?

src/cr/cube/cubepart.py
gergness: The labels don't have dimensionality from the slice and so don't need to be reordered. They are related to the dimensionality of the tuple. (eg if `columns_disaggregated_missing_unweighted_counts` is an array of length N tuples, this is always a single length N tuple.

src/cr/cube/cubepart.py
ernestoarbitrio: ok make sense

src/cr/cube/cubepart.py
ernestoarbitrio: ok gotcha

gergness (commented):
A few things:

- ~I tried to have a representation of the "disaggregated missing" when there were no actual missing, but I think it's causing test failures for python 3.8. Maybe we should just say that the disaggregated missing marginal isn't defined when there are no missing values?~ ~Shoot, this is actually happening on the length 1 arrays, which  I do think we need for consistency. Do you know why 3.8 isn't understanding them?~ This is now fixed, by specifying the dtypes in a more explicit way that works in all python/numpy versions
- I'm having trouble with the formatters. [This commit](https://github.com/Crunch-io/crunch-cube/pull/418/changes/901255cfe5b7abd414398813d5c581db9d12ca9d) I had to do with `--no-verify` because `ruff` in my git hooks doesn't match the `black` in the `tox` run during the CI. Thoughts on what I should do?
- Also, I put my python env in `.env` in the folder, but then when I to run `tox -e lint`, it lints the files in there too. Is there a way to avoid this? Should I move my env?

gergness:
~Also, I'd believe that I missed a line in my code coverage, but codecov seems confused about which line it is (this is a blank line): https://app.codecov.io/gh/Crunch-io/crunch-cube/pull/418?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=checks&utm_campaign=pr+comments&utm_term=Crunch-io~

I bet the rebasing I just did will help

ernestoarbitrio:
> A few things:
> 
> * ~I tried to have a representation of the "disaggregated missing" when there were no actual missing, but I think it's causing test failures for python 3.8. Maybe we should just say that the disaggregated missing marginal isn't defined when there are no missing values?~ Shoot, this is actually happening on the length 1 arrays, which  I do think we need for consistency. Do you know why 3.8 isn't understanding them?
> * I'm having trouble with the formatters. [This commit](https://github.com/Crunch-io/crunch-cube/pull/418/changes/901255cfe5b7abd414398813d5c581db9d12ca9d) I had to do with `--no-verify` because `ruff` in my git hooks doesn't match the `black` in the `tox` run during the CI. Thoughts on what I should do?
> * Also, I put my python env in `.env` in the folder, but then when I to run `tox -e lint`, it lints the files in there too. Is there a way to avoid this? Should I move my env?

how did you setup the env?

gergness:
I think with `python -m venv env` (with python 3.9 active apparently).

Was able to push through, so I this PR is ready for another round of review

src/cr/cube/matrix/measure.py
```diff
@@ -2679,11 +2693,25 @@ def _counts_are_defined(self):
...
+            return self._dimensions[1]
+        return self._dimensions[0]
+
     @lazyproperty
     def _squared_weights_are_defined(self):
         """Bool indicating whether squared weights are defined."""
         return self._second_order_measures.column_squared_bases.is_defined
 
+    @lazyproperty
+    def _subtotal_shape(self):
```
ernestoarbitrio: why the name of this property is `_subtotal_shape` if we're returning the len of them?

src/cr/cube/matrix/measure.py
```diff
@@ -2698,6 +2726,102 @@ def _opposing_numeric_values(self):
...
+        return len(self._missing_mask)
+
+    @lazyproperty
+    def _missing_mask(self):
+        """list of logicals indicating which rows/columns are missing values"""
+        return [
+            idx
+            for idx, el in enumerate(self._opposing_dimension.all_elements)
+            if el.missing
+        ]
```
ernestoarbitrio: `_missing_mask` returns indices, not booleans: The name "mask" typically implies a boolean array in numpy convention. It's actually a list of integer indices. Consider        
  _missing_indices or _missing_idxs

src/cr/cube/cubepart.py
```diff
@@ -2465,6 +2567,14 @@ def table_margin_range(self):
         """
         return self._measures.weighted_bases.table_margin_range
 
+    @lazyproperty
+    def table_missing(self):
+        """scalar/1D np.float64 ndarray of unweighted-N for table/each cell of strand.
```
ernestoarbitrio: this line is the same of table_base docstring, I see that at line 2574 you mention that is a complement of the table_base, but maybe it's better to mention here that these are the N for missings?

src/cr/cube/cubepart.py
```diff
@@ -2646,6 +2756,11 @@ def table_base(self):
         """Int scalar of the unweighted N of the table."""
         return self._scalar.table_base
 
+    @lazyproperty
+    def table_missing(self):
```
ernestoarbitrio, line 2795: Maybe this is a stupid question :D , but this `table_missing` is defined for Strand and Nub, not for Slice, I can see that maybe there's no reason to implement a table_missing from a 2D object, so just asking if this is intentional :D

ernestoarbitrio (approved):
LGTM, just left some stylistic comments + 1 question, see if you can address those
Thx

src/cr/cube/cubepart.py
gergness: I guess there could be, but we don't need it yet. It would be even more type unstable (a scalar on CAT x CAT, a 1D array of the length of rows (MR X CAT), a 1D array the length of the columns (CAT X MR) or 2D (MR X MR).

PR #419, opened 2026-04-27
Fixes for missing values

gergness:
Fixes two sentry errors I found: 
https://crunchio.sentry.io/issues/7445203302/?project=155823&referrer=slack
https://crunchio.sentry.io/issues/7436331489/?project=155203&referrer=slack

### DIFF
--- a/src/cr/cube/cube.py
+++ b/src/cr/cube/cube.py
@@ -789,7 +789,7 @@ class Cube:
         all_idxs = np.ix_(
             *tuple(d.all_elements.element_idxs for d in self._all_dimensions)
         )
-        # The dimension dimension order can change in case of numeric array variable on
+        # The dimension order can change in case of numeric array variable on
         # the row, and so valid indices needs to be returned in an ordered way.
         return tuple(all_idxs[i] for i in self._all_dimensions.dimension_order)
 
--- a/src/cr/cube/cube.py
+++ b/src/cr/cube/cube.py
@@ -549,7 +549,11 @@ class Cube:
         In case of presence of valid counts in the cube response the counts are replaced
         with the valid counts measure.
         """
-        return self.unweighted_counts_with_missings[self._valid_idxs]
+        return (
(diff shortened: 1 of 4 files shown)


src/cr/cube/cube.py
```diff
@@ -777,6 +783,16 @@ def _valid_idxs(self) -> Tuple[np.ndarray, ...]:
         # the row, and so valid indices needs to be returned in an ordered way.
         return tuple(valid_idxs[i] for i in self._all_dimensions.dimension_order)
 
+    @lazyproperty
+    def _all_idxs(self) -> Tuple[np.ndarray, ...]:
+        """Tuple of int64 ndarrays of all elements idx for each dimension."""
```
gergness, line 788: This feels a little silly, because we're not actually filtering any items, but I wasn't sure the best way to get the `np.ix_` objects so that they could be reordered into `dimension_order`

src/cr/cube/cube.py
```diff
@@ -777,6 +783,16 @@ def _valid_idxs(self) -> Tuple[np.ndarray, ...]:
...
+        all_idxs = np.ix_(
+            *tuple(d.all_elements.element_idxs for d in self._all_dimensions)
+        )
+        # The dimension dimension order can change in case of numeric array variable on
```
ernestoarbitrio:
```suggestion
        # The dimension order can change in case of numeric array variable on
```

src/cr/cube/stripe/measure.py
ernestoarbitrio, line 748: at line 756 there's this guard:
```
if self._unweighted_cube_counts.bases.shape == (0,):
```
so this means that if this happen here as well in `table_base_range` we can have the same issue. Because if bases = np.array([]) so an array with (0,) shape the min an max raise an error I suppose.

ernestoarbitrio (changes requested):

ernestoarbitrio (approved):

PR #420, opened 2026-04-28
maintenance: remove python 3.8 ci

gergness:
As discussed in https://crunch-io.slack.com/archives/C054H4GV389/p1776458190858319, we no longer need to support python 3.8

Looks like we're already 2 versions behind the latest, did we want to add 3.13 and 3.14?

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -49,7 +49,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v5
         with:
-          python-version: 3.8
+          python-version: 3.9
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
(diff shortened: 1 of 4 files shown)


.github/workflows/release.yml
```diff
@@ -49,7 +49,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v5
         with:
-          python-version: 3.8
+          python-version: 3.9
```
gergness, line 52: Wasn't sure why this was 3.8 but pypi was on 3.11? Should this be the lowest version we support?

ernestoarbitrio: it doesn't matter the 3.8 was to one that was available on github runner, but I believe 3.9 is also there.

ernestoarbitrio (approved):

PR #422, opened 2026-05-12
Native namespace package

gaetano-guerriero:


setup.py
```diff
@@ -49,12 +49,9 @@ def ascii_bytes_from(path, *paths):
     author_email="dev@crunch.io",
     license="MIT License",
     install_requires=install_requires,
-    tests_require=test_requires,
     extras_require={"testing": test_requires},
-    packages=find_packages("src", exclude=["tests"]),
```
gaetano-guerriero, line 54: find_namespaced_packages should be used, but also automatic discovery works

ernestoarbitrio (approved):

src/cr/cube/__init__.py
```diff
@@ -2,4 +2,4 @@
 
 """Initialization module for crunch-cube package."""
 
-__version__ = "3.3.7"
+__version__ = "3.3.8"
```
ernestoarbitrio, line 5: Did you try installing the package locally after moving this init ?

gaetano-guerriero: Yes, I made the similar change to crunch-ai, crunch-core, zoom (lib,server,exporter) and installed everything in a single environment.

So the three changes must be released together, I'm working on it

ernestoarbitrio: i see thanks

PR #424, opened 2026-05-21
Allow label transform functions

jjdelc:


gergness (approved):
Looks good, consider adding an integration test though? 

In particular, it wasn't clear to me whether it will consider "No Data" when looking for the common prefix in `trim_common`.  (I think it does the right thing because we access the labels via `Dimension.labels` which calls `self.all_elements.valid_elements`

Also note there are a few failing unit tests because of the internal changes.

ernestoarbitrio (approved):
LGTM

